PR #10, opened 2018-04-08
New deployment

rstormsf:
Problem: In order to deploy contracts, you have to fully sync a blockchain 
Solution: Use private keys and sendRawTx requests


cwhinfrey:
👍 Looks good

PR #20, opened 2018-04-24
Audit feedback: fix bug when requiredSignatures has changed

rstormsf:
Set 
```solidity
setNumDepositsSigned(hashMsg, 2**256-1);
```
to max value so the next tx will fail

### DIFF
--- a/contracts/upgradeable_contracts/U_ForeignBridge.sol
+++ b/contracts/upgradeable_contracts/U_ForeignBridge.sol
@@ -117,8 +117,6 @@ contract ForeignBridge is ERC677Receiver, Validatable {
     }
 
     function deposit(address recipient, uint256 value, bytes32 transactionHash) public onlyValidator {
-        require(address(erc677token()) != address(0x0));
-
         // Protection from misbehaing authority
         bytes32 hashMsg = keccak256(recipient, value, transactionHash);
         bytes32 hashSender = keccak256(msg.sender, hashMsg);
@@ -134,6 +132,7 @@ contract ForeignBridge is ERC677Receiver, Validatable {
         if (signed == validatorContract().requiredSignatures()) {
             // If the bridge contract does not own enough tokens to transfer
             // it will couse funds lock on the home side of the bridge
+            setNumDepositsSigned(hashMsg, 2**256-1);
             erc677token().mint(recipient, value);
             emit Deposit(recipient, value, transactionHash);
         }
@@ -155,7 +154,7 @@ contract ForeignBridge is ERC677Receiver, Validatable {
         bytes32 hash = keccak256(message);
         bytes32 hashSender = keccak256(msg.sender, hash);
 
-        uint256 signed = numMessagesSigned(hash) + 1;
+        uint256 signed = numMessagesSigned(hash).add(1);
 
         if (signed > 1) {
             // Duplicated signatures
@@ -174,6 +173,7 @@ contract ForeignBridge is ERC677Receiver, Validatable {
 
         emit SignedForWithdraw(msg.sender, hash);
         if (signed == validatorContract().requiredSignatures()) {
+            setNumMessagesSigned(hash, 2**256-1);
             emit CollectedSignatures(msg.sender, hash);
         }
     }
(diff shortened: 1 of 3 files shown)


Eenae:
> Set
> `setNumDepositsSigned(hashMsg, 2**256-1);`
> to max value so the next tx will fail

I think this solution is a bit dirty: set value to max `uint` so that next increment will hit safety break (`assert` inside function `add`). Safety breaks should't be hit during normal operation - control should reach `assert` only in case of unexpected bug. The simplest way to fix this is to add a check for `2**256-1` before increment, making `2**256-1` a special value, and even better - to introduce it as a constant, like size_t(-1) in C.

Other option is to store flag of processed deposit as a bit in the uint and pack and unpack (bool processed, uint signatureNumber) using bit operations.

And the cleanest but most expensive solution, of course, is an extra mapping.

akolotov:
Right. Probably the approach to store a flag in the uint is a good option since it will help control number of signed deposit if it is necessary in the future.  

```
    function deposit(address recipient, uint256 value, bytes32 transactionHash) public onlyValidator {
        // Protection from misbehaing authority
        bytes32 hashMsg = keccak256(recipient, value, transactionHash);
        bytes32 hashSender = keccak256(msg.sender, hashMsg);
        // Duplicated deposits
        require(!depositsSigned(hashSender));
        uint256 signed = numDepositsSigned(hashMsg);
        // Do not handle deposits which were already handled
        require(signed & 2**255 != 2**255);
        setDepositsSigned(hashSender, true);

        signed = signed + 1;
        setNumDepositsSigned(hashMsg, signed);

        emit SignedForDeposit(msg.sender, transactionHash);

        if (signed == validatorContract().requiredSignatures()) {
            // If the bridge contract does not own enough tokens to transfer
            // it will couse funds lock on the home side of the bridge
            setNumDepositsSigned(hashMsg, signed | 2**255);
            erc677token().mint(recipient, value);
            emit Deposit(recipient, value, transactionHash);
        }
    }
```

contracts/upgradeable_contracts/U_ForeignBridge.sol
```diff
@@ -150,34 +152,37 @@ contract ForeignBridge is ERC677Receiver, Validatable {
...
+        emit SignedForWithdraw(msg.sender, hashMsg);
         if (signed == validatorContract().requiredSignatures()) {
-            emit CollectedSignatures(msg.sender, hash);
+            uint256 protectionNumber = signed | 2**255;
+            setNumMessagesSigned(hashMsg, protectionNumber);
+            emit CollectedSignatures(msg.sender, hashMsg);
         }
     }
 
+    function isDepositProcessed(uint256 _number) public view returns(bool) {
```
akolotov: since the method is called for two different scenarios: depost and withdraw, let's give a common name to it like `isAlreadyProcessed`

contracts/upgradeable_contracts/U_ForeignBridge.sol
```diff
@@ -150,34 +152,37 @@ contract ForeignBridge is ERC677Receiver, Validatable {
...
         setSignatures(signIdx, signature);
 
-        setNumMessagesSigned(hash, signed);
+        setNumMessagesSigned(hashMsg, signed);
 
-        emit SignedForWithdraw(msg.sender, hash);
+        emit SignedForWithdraw(msg.sender, hashMsg);
         if (signed == validatorContract().requiredSignatures()) {
-            emit CollectedSignatures(msg.sender, hash);
+            uint256 protectionNumber = signed | 2**255;
```
akolotov: in order to hide 'magic' bitwise operation, is it worth to introduce a method 'markAsProcessed'?

```
function markAsProcessed(uint256 _v) internal pure returns(uint256) {
    return _v | 2 ** 255;
}
```

contracts/libraries/Helpers.sol
```diff
@@ -58,6 +58,7 @@ library Helpers {
 
 library MessageSigning {
     function recoverAddressFromSignedMessage(bytes signature, bytes message) internal pure returns (address) {
+        require(message.length == 116);
```
akolotov: for me, it is not logical to put the message length check here since, logically it is assumed that we can recover an address for the message of any length.

PR #21, opened 2018-04-26
Audit team2

rstormsf:
This PR depends on previous PR #20 

Things completed:
- Change some functions from `public` to `external`
- Fix `transferAndCall`. Use `requeire`
- move ` require(_message.length == 116);` to Library
- revert() on `hasEnoughValidSignatures`
- use 
```
 recipient := and(mload(add(message, 20)), 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)
```
in getRecepient
- use `uint256` vs `string` in Proxy contracts for versioning
- add ` mstore(0x40, add(ptr, size))` to Proxy.sol
- Fix deployment scripts


rstormsf:
I'm sorry @Eenae I think you are right, I should remove all flats and git ignore them.

akolotov:
@rstormsf we need flat files to put them to etherscan later for the contract verification. It was the intent why we introduced them.

contracts/upgradeable_contracts/U_ForeignBridge.sol
```diff
@@ -110,30 +108,31 @@ contract ForeignBridge is ERC677Receiver, Validatable {
...
         bytes32 hashMsg = keccak256(recipient, value, transactionHash);
         bytes32 hashSender = keccak256(msg.sender, hashMsg);
         // Duplicated deposits
         require(!depositsSigned(hashSender));
         setDepositsSigned(hashSender, true);
 
-        uint256 signed = numDepositsSigned(hashMsg).add(1);
+        uint256 signed = numDepositsSigned(hashMsg);
+        require(!isAlreadyProcessed(signed));
+        signed = signed.add(1);
```
akolotov: please change by the same as you did in `submitSignature()`

Eenae:
> @rstormsf we need flat files to put them to etherscan later for the contract verification. It was the intent why we introduced them.

Sure, but in our practice we are treating them as build artifacts, storing them under build/flattened/ and not committing them. This files could always be re-generated from sources by flatten.sh.

contracts/upgradeable_contracts/U_ForeignBridge.sol
```diff
@@ -110,30 +108,32 @@ contract ForeignBridge is ERC677Receiver, Validatable {
...
+        uint256 signed = numDepositsSigned(hashMsg);
+        require(!isAlreadyProcessed(signed));
+        // the check above assumes that the case when the value could be overflew will not happen in the addition operation below
+        signed = signed + 1;
+
         setNumDepositsSigned(hashMsg, signed);
 
         emit SignedForDeposit(msg.sender, transactionHash);
 
         if (signed == validatorContract().requiredSignatures()) {
```
Eenae: There is another issue that concerns me. What if `requiredSignatures` is decreased below the current `signed` value for some reason? Current deposit would never be processed.
Looks like `>=` should be used instead of `==`.

contracts/upgradeable_contracts/U_ForeignBridge.sol
```diff
@@ -147,37 +147,41 @@ contract ForeignBridge is ERC677Receiver, Validatable {
...
-        bytes32 signIdx = keccak256(hash, (signed-1));
+        bytes32 signIdx = keccak256(hashMsg, (signed-1));
         setSignatures(signIdx, signature);
 
-        setNumMessagesSigned(hash, signed);
+        setNumMessagesSigned(hashMsg, signed);
 
-        emit SignedForWithdraw(msg.sender, hash);
+        emit SignedForWithdraw(msg.sender, hashMsg);
         if (signed == validatorContract().requiredSignatures()) {
```
Eenae: Issue in case of a sudden `requiredSignatures` decrease (see above).

rstormsf:
@Eenae yes,  I have removed them from repo.

Eenae:
Now everything is ok!

Issue #27, opened 2018-05-05
setRequiredSignatures must raise an event to upgrade values on the bridge instances


akolotov:
Currently if `setRequiredSignatures()` is invoked, bridge instances do not have ability to know that number of required signatures was changed.

rstormsf:
#28

Issue #26, opened 2018-05-04
Add an ability to calculate actual gas usage prior to send deployment transactions


akolotov:
Currently `gas limit` is hardcoded:

https://github.com/poanetwork/poa-bridge-contracts/blob/8cf94e4b8b9651d163f15a135ea7f8841b5d2d46/deploy/src/deploymentUtils.js#L54

For example, it causes fail on Ropsten.

rstormsf:
fixed #27

PR #28, opened 2018-05-06
Add GasPrice, requiredConfirmations, optimize hashMessage, deployment script update

### LINKED ISSUE #25: Add and ability to see JSON response in case of error
Currently if a deployment transaction fails there is no way to see the reason:
https://github.com/poanetwork/poa-bridge-contracts/blob/8cf94e4b8b9651d163f15a135ea7f8841b5d2d46/deploy/src/deploymentUtils.js#L79-L80

rstormsf:
1. Add GasPrice to contracts in order to submit when constructing a tx on poa-bridge rust side
2. Add requiredConfirmations to be read from poa-bridge side
3. Optimize hashMessage by  removing uintToString method
4. Deployment script update to show error message

fixes
#25 
#26 
#27


contracts/libraries/Message.sol
```diff
@@ -81,7 +81,9 @@ library Message {
 
     function hashMessage(bytes message) internal pure returns (bytes32) {
         bytes memory prefix = "\x19Ethereum Signed Message:\n";
```
akolotov: check that the message length is really 116.

contracts/upgradeable_contracts/U_Validatable.sol
```diff
@@ -19,4 +20,24 @@ contract Validatable is EternalStorage {
         _;
     }
 
+    function setGasPrice(uint256 _gasPrice) public onlyOwner {
```
akolotov: Why do you put these calls to `Validatable` contract? It is not look logical for me. I would introduce a contract something like `BasicBridge` and move there this logic, management of tx and daily limits. Since it is common for both bridges.

deploy/README.md
```diff
@@ -7,28 +7,33 @@ poa-parity
...
-FOREIGN_MAX_AMOUNT_PER_TX=100000000000000000
-FOREIGN_MIN_AMOUNT_PER_TX=10000000000000000
+HOME_OWNER_MULTISIG=0x
+HOME_UPGRADEABLE_ADMIN_VALIDATORS=0x
+HOME_UPGRADEABLE_ADMIN_BRIDGE=0x
+HOME_DAILY_LIMIT=30000000000000000000000000
+HOME_MAX_AMOUNT_PER_TX=1500000000000000000000000
+HOME_MIN_AMOUNT_PER_TX=500000000000000000
+HOME_REQUIRED_BLOCK_CONFIRMATIONS=1
+HOME_GAS_PRICE=1
```
akolotov: Do we want to configure `gasLimitWithdrawRelay` in `.env` as well?

deploy/README.md
```diff
@@ -7,28 +7,33 @@ poa-parity
...
+
+FOREIGN_RPC_URL=https://sokol.poa.network
+FOREIGN_OWNER_MULTISIG=0x
+FOREIGN_UPGRADEABLE_ADMIN_VALIDATORS=0x
+FOREIGN_UPGRADEABLE_ADMIN_BRIDGE=0x
+FOREIGN_DAILY_LIMIT=15000000000000000000000000
+FOREIGN_MAX_AMOUNT_PER_TX=750000000000000000000000
+FOREIGN_MIN_AMOUNT_PER_TX=500000000000000000
+FOREIGN_REQUIRED_BLOCK_CONFIRMATIONS=8
+FOREIGN_GAS_PRICE=10
```
akolotov: Do we want to configure `gasLimitDepositRelay` and `gasLimitWithdrawConfirm` in `.env` as well?

contracts/upgradeable_contracts/U_ForeignBridge.sol
```diff
@@ -40,6 +42,8 @@ contract ForeignBridge is ERC677Receiver, Validatable {
         uintStorage[keccak256("deployedAtBlock")] = block.number;
         uintStorage[keccak256("maxPerTx")] = _maxPerTx;
         uintStorage[keccak256("minPerTx")] = _minPerTx;
+        uintStorage[keccak256("gasPrice")] = _homeGasPrice;
```
akolotov: I think there is a mistake here.
Initially foreign contract operated with a gas price to send it as part of `Withdraw()` event to notify home side how much a user would like to compensate validators for relaying his/her transaction. It was assumed that gas price was specified by the user as part of the withdraw request. We do not use the compensation but didn't have resources to change Rust code to not operate with gas prices received in `Withdraw()` event, so we hardcoded this value to 1 gwei. Moreover this value does not need to be returned through getter since there is no scenario when someone would like to know which gas price is used for this _"legacy"_ code.

Now we are introducing new entity - a gas price to be requested by the bridge instance to use it in its transaction. This is completely different use case. 

Here is my suggestion:
1. to have two entities: `gasPriceForCompensationAtHomeSide` and `gasPrice`
2. `gasPriceForCompensationAtHomeSide` will be always 1gwei. We will not allow to change it. It is used as a stub in `Withdraw()` event.
3. `gasPrice` will be configured at initialization stage and could be changed by `setGasPrice()`. It is used by the bridge to set gas price for transactions if the oracle is not accessible.

contracts/upgradeable_contracts/U_HomeBridge.sol
```diff
@@ -16,18 +16,24 @@ contract HomeBridge is EternalStorage, Validatable {
         address _validatorContract,
         uint256 _homeDailyLimit,
         uint256 _maxPerTx,
-        uint256 _minPerTx
+        uint256 _minPerTx,
+        uint256 _foreignGasPrice,
```
akolotov: this should not be called `_foreignGasPrice` since it actually reflects gas price which should be used for transactions sent to Home Bridge.

contracts/libraries/Message.sol
rstormsf: it's already been done in all places before that call is made.

deploy/.env.example
```diff
@@ -1,24 +1,28 @@
 DEPLOYMENT_ACCOUNT_ADDRESS=0xb8988b690910913c97a090c3a6f80fad8b3a4683
 DEPLOYMENT_ACCOUNT_PRIVATE_KEY=67..14
+GAS_LIMIT=4000000
```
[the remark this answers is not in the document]
rstormsf: agree

deploy/.env.example
```diff
@@ -1,24 +1,28 @@
...
+GAS_PRICE=10
```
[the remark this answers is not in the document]
rstormsf: Nah...POA is always cheap

contracts/libraries/Message.sol
akolotov: thanks!

contracts/upgradeable_contracts/U_Validatable.sol
rstormsf: ok, lets rename validatable into BasicBridge?

deploy/README.md
rstormsf: well, we are not passing it into `initialize` method

deploy/README.md
rstormsf: we are not passing and calling those methods yet since rust is not reading them

deploy/.env.example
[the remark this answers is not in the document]
akolotov: you mean it is hardcoded in deployment scripts and always is 1 gwei, right?

contracts/upgradeable_contracts/U_ForeignBridge.sol
rstormsf: ok

contracts/upgradeable_contracts/U_HomeBridge.sol
rstormsf: ok

contracts/upgradeable_contracts/U_Validatable.sol
akolotov: I would say that `Validatable` like `Ownable` for tokens.
At the same time tokens have some common part which is called `BasicToken`. So, I would use the same logic for bridge: differentiate logic related to validators and logic related to another bridge configuration.
But it is really `my liking` so up to you.

deploy/README.md
akolotov: but if we passing gas price there should we pass gas limits too since the logic is the same for both: they should be used to build transactions by bridge.

deploy/README.md
akolotov: the same as for 'gas price' and 'required block confirmations'.
just trying to be consistent.

deploy/.env.example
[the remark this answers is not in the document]
rstormsf: no I mean, it's ok to spend 10 gwei on POA chain

PR #50, opened 2018-07-06
add feature on token transfer invoke contractFallback silently

rstormsf:
1. Change `transfer` method to invoke `onTokenTransfer` if recipient is contract

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -26,10 +26,23 @@ contract ERC677BridgeToken is
     function transferAndCall(address _to, uint _value, bytes _data)
         external validRecipient(_to) returns (bool)
     {
-        require(transfer(_to, _value));
-        emit Transfer(msg.sender, _to, _value, _data);
+        require(_value <= balances[msg.sender]);
+
+        balances[msg.sender] = balances[msg.sender].sub(_value);
+        balances[_to] = balances[_to].add(_value);
         if (isContract(_to)) {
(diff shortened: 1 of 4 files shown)


contracts/ERC677BridgeToken.sol
```diff
@@ -26,20 +26,32 @@ contract ERC677BridgeToken is
     function transferAndCall(address _to, uint _value, bytes _data)
         external validRecipient(_to) returns (bool)
     {
-        require(transfer(_to, _value));
-        emit Transfer(msg.sender, _to, _value, _data);
+        require(_value <= balances[msg.sender]);
```
akolotov: Personally I don't like code duplication. Do you think it is OK to simplify the changes a bit? Will it work?

```
function transferAndCall(address _to, uint _value, bytes _data)
    external validRecipient(_to) returns (bool)
{
    require(__transfer(_to, _value));
    emit Transfer(msg.sender, _to, _value, _data);

    if (isContract(_to)) {
        require(contractFallback(_to, _value, _data));
    }
    return true;
}

function __transfer(address _to, uint256 _value) internal returns(bool)
{
    return super.transfer(_to, _value);
}

function transfer(address _to, uint256 _value) public returns (bool)
{
    require(__transfer(_to, _value));
    if (isContract(_to)) {
        contractFallback(_to, _value, new bytes(0));
    }
    return true;
}
```

rstormsf:
Good idea, please check last commit

PR #51, opened 2018-07-10
Prevent an attack to send cross chain transfer using contract without payable fallback

rstormsf:
this PR depends on #50 

Scenario:

1. Deploy a contract with `revert` in `fallback` on both chains: Home and Foreign
2. Use a bridge to initiate a transfer
3. Make sure the bridge still is able to send funds to recipient using self-destructed contract.

### DIFF
--- a/contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
+++ b/contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
@@ -5,6 +5,12 @@ import "../BasicBridge.sol";
 import "../../upgradeability/EternalStorage.sol";
 import "../BasicHomeBridge.sol";
 
+contract Sacrifice {
+    constructor(address _recipient) public payable {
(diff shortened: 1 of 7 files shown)


contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
```diff
@@ -42,7 +48,9 @@ contract HomeBridgeNativeToErc is EternalStorage, BasicBridge, BasicHomeBridge {
     }
 
     function onExecuteAffirmation(address _recipient, uint256 _value) internal returns(bool) {
-        _recipient.transfer(_value);
+        if (!_recipient.send(_value)) {
+            (new Sacrifice).value(_value)(_recipient);
```
akolotov: Do you think that comments to clarify the code will decrease level of security of the code?

rstormsf: ? what do you want?

PR #52, opened 2018-07-11
Msg checking

rstormsf:
this PR depends on #51 #50 

Record a tx at the time of reception.
Run `onMitmPrevention` check to make sure it wasn't spoofed

### DIFF
--- a/contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
+++ b/contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
@@ -5,6 +5,12 @@ import "../BasicBridge.sol";
 import "../../upgradeability/EternalStorage.sol";
 import "../BasicHomeBridge.sol";
 
+contract Sacrifice {
+    constructor(address _recipient) public payable {
+        selfdestruct(_recipient);
+    }
+}
+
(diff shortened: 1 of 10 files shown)


contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
```diff
@@ -44,6 +44,11 @@ contract HomeBridgeNativeToErc is EternalStorage, BasicBridge, BasicHomeBridge {
         require(msg.data.length == 0);
         require(withinLimit(msg.value));
         setTotalSpentPerDay(getCurrentDay(), totalSpentPerDay(getCurrentDay()).add(msg.value));
+        incrementNonce(msg.sender);
```
akolotov: I don't like an idea that we are increasing gas consumption significantly (40 000) whereas the issue for Foreign->Home transactions still exists.
I would suggest to not address these changes at all since in order to implement actual attack:
1. connection must be spoofed for every bridge instance
AND
2. HTTP needs to be configured for RPC channel explicitly instead of HTTPS

rstormsf: we agreed to make those changes and now what? going back again  to your decision? what the hell is going on?

rstormsf:
Decision came to not implement this feature

PR #56, opened 2018-07-24
add deployedAtBlock for Validators contract

rstormsf:
#54 same change for  refactor_v1 branch


akolotov:
The same changes were commited under #58. That's why I am closing this PR without actual commit.

PR #57, opened 2018-07-24
 Double spending fix

rstormsf:
Due to lack of difference between signatures for NativeToERC and ERC20ToERC20 (home to foreign) transfers, those can be used for double spending.

Recomendations: Implement different message for NativeToERC and ERC20ToERC20 transfers. Consider adding tokenAddress to ERC20ToERC20 transfers message to avoid double spending and changing token provider contract.

# Do NOT MERGE until the proper fix is added to nodejs-bridge repo


contracts/upgradeable_contracts/BasicForeignBridge.sol
```diff
@@ -9,14 +9,14 @@ import "../libraries/Message.sol";
...
-
     event RelayedMessage(address recipient, uint value, bytes32 transactionHash);
     function executeSignatures(uint8[] vs, bytes32[] rs, bytes32[] ss, bytes message) external {
         Message.hasEnoughValidSignatures(message, vs, rs, ss, validatorContract());
         address recipient;
         uint256 amount;
         bytes32 txHash;
-        (recipient, amount, txHash) = Message.parseMessage(message);
+        address contractAddress;
+        (recipient, amount, txHash, contractAddress) = Message.parseMessage(message);
```
akolotov: `contractAddress` is extracted from the message but not used to make sure that the message was not taken from another type of the bridge.
Consider to add:
```require(contractAddress == address(this));```

rstormsf: please check in latest commit

PR #61, opened 2018-08-07
Avoid missing directory error on flatten.sh

ArvsIndrarys:
Added directory presence check to avoid error on script call. If the directory doesn't exist, create it.
(can be propagated to all branches)

### DIFF
--- a/flatten.sh
+++ b/flatten.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+if [ ! -d "flats" ]; then
+	mkdir flats
+fi
+
 rm -rf flats/*
 ./node_modules/.bin/truffle-flattener contracts/upgradeable_contracts/U_ForeignBridge.sol > flats/ForeignBridge_flat.sol
 ./node_modules/.bin/truffle-flattener contracts/upgradeable_contracts/U_BridgeValidators.sol > flats/BridgeValidators_flat.sol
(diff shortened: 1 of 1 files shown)


flatten.sh
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+if [ ! -d "flats" ]; then
```
akolotov: Does it make sense to extend your change to remove the directory content only if the directory exists:

```
if [ ! -d "flats" ]; then
    mkdir flats
else
    rm -rf flats/*
fi
```

ArvsIndrarys: It is what this script does : 
If the folder doesn't exist,the script doesn't fail and creates the directory. That is my change.

If the folder already exists, removes its content and flatten the .sol.

It is useful in case you have a newer version of the SC but the directory already exists.

ArvsIndrarys:
Change should be done

akolotov:
Currently we accept PRs only for `refactor_v1` branch since it is considered so far as the main development branch. Could you rebase your branch to `refactor_v1` otherwise I am getting a merge error.

ArvsIndrarys:
Should be okay now.

akolotov:
image
I am getting a merge conflict error

rstormsf:
this PR should not have 
`contracts/upgradeable_contracts/U_HomeBridge.sol`

ArvsIndrarys:
I do not know how to do it from the Github GUI. I'll do another pull request with the said changes.

PR #68, opened 2018-08-16
Add requiredMessageLength method to BasicHomeBridge contract

fvictorio:
This is necessary for https://github.com/poanetwork/bridge-nodejs/issues/56.

I also modified the library method to `internal` so that it's inlined in the bridge contract method. This is necessary because the deploy script is not deploying and linking the `Message` library, so using non-internal methods causes the deploy to fail.


rstormsf:
we should probably add a test to it.

fvictorio:
Added some tests.

PR #69, opened 2018-08-17
Allow using the deploy script from another directory

fvictorio:
When working on the bridge, I often do this:

```
cd submodules/poa-bridge-contracts/deploy
node deploy.js
cd -
```

when I could do:

```
node submodules/poa-bridge-contracts/deploy/deploy.js
```

Even better: we could have a npm script that deploys the contracts. Right now this isn't possible because the deploy script uses paths relative to the cwd of the process. This PR makes these paths independent.

### DIFF
--- a/deploy/deploy.js
+++ b/deploy/deploy.js
@@ -1,5 +1,8 @@
 const fs = require('fs');
-require('dotenv').config();
+const path = require('path')
+require('dotenv').config({
+  path: path.join(__dirname, '.env')
+});
 
 async function deployNativeToErc(){
   const deployHome = require('./src/native_to_erc/home');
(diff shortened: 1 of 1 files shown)


deploy/deploy.js
```diff
@@ -11,7 +14,8 @@ async function deployNativeToErc(){
   console.log(`[   Home  ] HomeBridge: ${homeBridge.address} at block ${homeBridge.deployedBlockNumber}`)
   console.log(`[ Foreign ] ForeignBridge: ${foreignBridge.address} at block ${foreignBridge.deployedBlockNumber}`)
   console.log(`[ Foreign ] POA20: ${erc677.address}`)
-  fs.writeFileSync('./bridgeDeploymentResults.json', JSON.stringify({
+  const deployResultsPath = path.join(__dirname, './bridgeDeploymentResults.json')
```
akolotov: Should you do the same for `deployErcToErc()`?

deploy/deploy.js
```diff
@@ -1,5 +1,8 @@
 const fs = require('fs');
-require('dotenv').config();
+const path = require('path')
+require('dotenv').config({
+  path: path.join(__dirname, '.env')
+});
 
 async function deployNativeToErc(){
   const deployHome = require('./src/native_to_erc/home');
```
akolotov: If there is no usage of `__dirname` in `require('./src/native_to_erc/home');` does it mean that `require` works with paths in different manner?

deploy/deploy.js
fvictorio: Oh, I missed that, thanks!

deploy/deploy.js
fvictorio: Yes, `require` is always relative to the file, so you don't need to do the `path.join(__dirname, ...)` trick.

Issue #70, opened 2018-08-17
Deployment with existing ERC20 token on Home side 


akolotov:
Current version of the bridge contracts deployment in `erc20-to-erc20` mode assumes that a new ERC20/ERC677 token contract is deployed in `Home` network together with other bridge contracts.

It make sense to have an address of existing ERC20/ERC677 token on `Home` side configurable in `.env` file. 

For example, we introduce `HOME_ERC677_TOKEN_ADDRESS` parameter as so, if it exists in `.env` file, the new token will not be deployed and the configured token address is used in `Home` bridge contract, otherwise (if the parameter is not specified) `Home` bridge contract will use a new token deployed as part of the bridge contracts initialization.

rstormsf:
use case?

akolotov:
Use case is the following:
An application would like to use it own version of token on Home (probably a modification of our token). Most probably they will use their own procedure to deploy and initialize token as so the token contract will be available in the chain before the bridge contracts will be deployed.
At the current moment there is the ability to change the token address in the bridge contract after the bridge contracts deployment: some dummy token will be deployed as part of run `deploy.js` and then its address replaced by another token. It does not seems rational - much more logical to have the ability to provide the token address on the deployment stage in order to reduce 'garbage' appeared in the chain.

akolotov:
As it was mentioned in https://github.com/poanetwork/poa-bridge-contracts/pull/71#discussion_r211768506 it makes sense to consider the following enhancement:

we add the following configuration parameters in `.env`-file:
```
HOME_ERC677_TOKEN_ADDRESS 
HOME_ERC677_TOKEN_NAME
HOME_ERC677_TOKEN_SYMBOL
HOME_ERC677_TOKEN_DECIMALS
```
and
```
FOREIGN_ERC20_TOKEN_ADDRESS 
FOREIGN_ERC677_TOKEN_NAME
FOREIGN_ERC677_TOKEN_SYMBOL
FOREIGN_ERC677_TOKEN_DECIMALS
```
where FOREIGN_ERC20_TOKEN_ADDRESS will be used instead of ERC20_TOKEN_ADDRESS.

They will be used by deployment script:
In `native-to-erc` mode we expect:
  * either `FOREIGN_ERC20_TOKEN_ADDRESS` which will link the bridge contract to an existing ERC677 token;
[...]

PR #71, opened 2018-08-18
Fix an  issue when  receipt is returning blockNumber null; add name, symbol, decimals to erc677

rstormsf:
## Issue 1:
I noticed when  deploying to Sokol network that getReceipt returned a response with  blockNumber being null
While it's very strange behavior, we can fix the issue in deployment.

The  only reason why this could happen I could think  of is when blockchain has reorgs.

## Issue 2:

Provide a way to set NAME, SYMBOL, DECIMALS for erc20-to-erc20 bridgeable token contract

## Issue 3:
requireBlockConfirmations wasn't passed to ForeingBridgeErcToErc (https://github.com/poanetwork/poa-bridge-contracts/issues/74)

## Issue 4:
pass as env var erc677 name, symbol, decimals


rstormsf:
I found this  issue:
https://github.com/paritytech/parity-ethereum/issues/3569#issuecomment-262299428

still don't know understand why receipt call would  return blockNumber null

deploy/.env.example
```diff
@@ -26,6 +26,9 @@ FOREIGN_REQUIRED_BLOCK_CONFIRMATIONS=8
 FOREIGN_GAS_PRICE=10
 #for bridge erc to erc mode
 ERC20_TOKEN_ADDRESS=
+HOME_BRIDGEBLE_TOKEN_NAME=
```
akolotov: - [ ] Move these three lines to that part of the file which is related to `Home` network.
- [ ] Update README.md to reflect the changes (https://github.com/poanetwork/poa-bridge-contracts/blob/refactor_v1/deploy/README.md)

rstormsf: I moved into this section under comment `#for bridge erc to erc mode`

rstormsf: will do

deploy/src/erc_to_erc/foreign.js
```diff
@@ -71,8 +73,6 @@ async function deployForeign() {
     url: FOREIGN_RPC_URL
   });
   assert.equal(txInitializeForeign.status, '0x1', 'Transaction Failed');
-  const validatorOwner = await bridgeValidatorsForeign.methods.owner().call();
```
akolotov: Just to write it somewhere: as we discussed on the call this issue could appear in situation when a block which includes the transaction to change the state of the contract become an uncle. In this case the state introduced by the transaction will not be changed until the transaction is included in a new block. 

We remove the call since
a) there is no need get owner of the contract since it is used just to display the owner address, but it is known any way;
b) there is no simple way without significant delay (to wait some number of blocks to finalize state) to fix this issue.

Since the issue common for all chains and to have user interface consistency please consider to remove similar code from:
1. `erc_to_erc/home.js`
2. `native_to_erc/foreign.js`
3. `native_to_erc/home.js`

deploy/src/native_to_erc/foreign.js
```diff
@@ -37,7 +41,7 @@ async function deployForeign() {
   console.log('========================================\n')
 
   console.log('\n[Foreign] deploying POA20 token')
```
akolotov: Replace `POA20` in the log by the corresponding value of `BRIDGEABLE_TOKEN_SYMBOL` variable.
Please do this for all places where it is applicable.

deploy/.env.example
```diff
@@ -5,6 +5,10 @@ DEPLOYMENT_GAS_LIMIT=4000000
 DEPLOYMENT_GAS_PRICE=10
 GET_RECEIPT_INTERVAL_IN_MILLISECONDS=3000
 
+BRIDGEABLE_TOKEN_NAME="Your New Bridged Token"
```
akolotov: As per our discussion on the call please consider to
1. make a configurable option to use already deployed token on Home side (`erc20-to-erc20` mode, https://github.com/poanetwork/poa-bridge-contracts/issues/70)
2. have ability to deploy an erc20/677 token on Foreign side (`erc20-to-erc20` mode)

In this case we will need the following configuration parameters:
```
HOME_ERC677_TOKEN_ADDRESS 
HOME_ERC677_TOKEN_NAME
HOME_ERC677_TOKEN_SYMBOL
HOME_ERC677_TOKEN_DECIMALS
```
and 
```
FOREIGN_ERC20_TOKEN_ADDRESS 
FOREIGN_ERC677_TOKEN_NAME
FOREIGN_ERC677_TOKEN_SYMBOL
FOREIGN_ERC677_TOKEN_DECIMALS
```

where `FOREIGN_ERC20_TOKEN_ADDRESS` will be used instead of `ERC20_TOKEN_ADDRESS`.

deploy/src/native_to_erc/foreign.js
rstormsf: will do

deploy/src/erc_to_erc/foreign.js
rstormsf: will do

rstormsf:
let me know if everything looks good now

PR #72, opened 2018-08-20
Avoid missing directory error on flatten

ArvsIndrarys:
Added a directory check before any updates in the flats folder. This avoid errors on the first time the script is used.

### DIFF
--- a/flatten.sh
+++ b/flatten.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-rm -rf flats/*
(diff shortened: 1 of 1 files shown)


ArvsIndrarys:
not needed anymore.

akolotov:
Could you share why you marked this as "not needed anymore"?

ArvsIndrarys:
The fix would work on the master branch. As the `flats/` directory already exists on this branch (refactor_v1), no more errors arise.

Issue #73, opened 2018-08-20
Arbitrary message bridging


akolotov:
_Based on feature description available here: https://hackmd.io/61fw0kf8T2C9bvDTOpzyVQ_

POA network projects provides a set of bridges working in EVM-based 
networks which allows to transfer assets from one blockchain to another.
At this moment the POA bridge supports two modes:
  * `native-to-erc20` (used in production)
  * `erc20-to-erc20` (in development stage)

Intorduction of every new bridge mode requires changes as on both the contract side and server side. It increase efforts needed for development and testing. That's why it is good idea to think of generic bridging approach which would allow transfer arbitrary data between two blockchains as so any depending on receiver the data could be interpreted as transfer of digital assets or an arbitrary contract method invocation.

## Basic bridge functionality

Let's consider that such kind of bridge has several components:
1. A contract in `Home` network which sends a message to another blockchain - `HomeSender`;
2. A contract in `Foreign` to receive the message from `Home` blockchain - `ForeignReceiver`;
3. A bridge contract in `Home` network which is responsible for transferring the message from `Home`- `HomeBridge`;
4. A bridge contract in `Foreign` network which is responsible for delivering the message to `ForeignReceiver`.
5. Bridge process (server) connected to both `Home` and `Foreign` blockchains which is responsible for listening events on `Home` side and creating transactions in `Foreign` side.

Having this we could describe the process of transferring the message as the set of the following steps:

1. A user performs an action to interact with `HomeSender` by invoking directly or indirectly its method as so `HomeSender` needs to call the method `Action1` from `ForeignReceiver` contract with some integer parameter `P`.

```
function Action1(uint256 P);
```

2. `HomeSender` invokes the method `requireToPassMessage` from `HomeBridge` and pass the following parameters:
    - `contractAddress` - address of `ForeignReceiver`;
    - `data` - set of bytes where the selector of `Action1` and value of parameter `P` encoded;
    - `gas` - estimated value of _gas_ which is expected to be spent for `Action1` execution.

```
function requireToPassMessage(address contractAddress, bytes data,
                           uint256 gas);
```

_Example of way to encode `data`:_
```
bytes4 methodSelector = ForeignReceiver(0).Action1.selector;
bytes memory data = abi.encodeWithSelector(methodSelector, P);
```
which equals to
```
bytes4 methodSelector = keccak256("Action1(uint256)");
bytes memory data = abi.encodeWithSelector(methodSelector, P);
```

[...]

PR #77, opened 2018-09-04
Add arbitrary message bridging contracts

patitonar:
Relates to #73 

This PR adds the AMB implementation for the subsidized mode only.

Draft implementation of defrayal mode can be found on branch: `amb-defrayal-mode`


contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -0,0 +1,30 @@
...
+import "../Validatable.sol";
+import "../BasicBridge.sol";
+
+
+contract BasicAMB is BasicBridge {
+    bytes constant internal SUBSIDIZED_MODE = bytes(abi.encodePacked("AMB-subsidized-mode"));
+    bytes constant internal DEFRAYAL_MODE = bytes(abi.encodePacked("AMB-defrayal-mode"));
+
+    function setSubsidizedModeForForeign() public onlyOwner {
+        bytesStorage[keccak256(abi.encodePacked("foreignBridgeMode"))] = SUBSIDIZED_MODE;
```
akolotov: Since it is `BasicAMB` which means to be generic, it should not contain a setter and a getter for `foreignBridgeMode`

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
```diff
@@ -0,0 +1,115 @@
...
+            if (isWithdrawFromDepositSelector(_data)) {
+                accountForAction = _sender;
+            }
+        }
+
+        require(_contract.call.gas(_gas)(_data));
+    }
+
+    function _defrayAndPassMessage(address _sender, address _contract, bytes _data, uint256 _gas) internal {
+        uint256 fee = (PASS_MESSAGE_GAS + _gas) * tx.gasprice;
```
akolotov: please use SafeMath here and in other places where arithmetic operations with variables are used.

contracts/upgradeable_contracts/arbitrary_message/ForeignAMB.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity 0.4.24;
+
+import "./BasicHomeAMB.sol";
+import "./BasicForeignAMB.sol";
+
+
+contract ForeignAMB is BasicForeignAMB {
+    function initialize() public returns(bool) {
+        require(!isInitialized(), "already initialized");
```
akolotov: Since string revert reasons consume lots of gas during deployment, please avoid to use them. If revert reason is really needed consider to have a uint256 instead.

patitonar: I'll remove them for now, we can add revert reasons as uint256 later if we really need them.

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
patitonar: Thanks! Fixed

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -10,6 +10,31 @@ contract BasicAMB is BasicBridge {
...
+        require(!isInitialized());
+        require(_validatorContract != address(0));
+        require(_gasPrice > 0);
+        require(_requiredBlockConfirmations > 0);
+        require(_minPerTx >= 0 && _maxPerTx > _minPerTx && _dailyLimit > _maxPerTx);
+        addressStorage[keccak256(abi.encodePacked("validatorContract"))] = _validatorContract;
+        uintStorage[keccak256(abi.encodePacked("deployedAtBlock"))] = block.number;
+        uintStorage[keccak256(abi.encodePacked("dailyLimit"))] = _dailyLimit;
+        uintStorage[keccak256(abi.encodePacked("maxPerTx"))] = _maxPerTx;
+        uintStorage[keccak256(abi.encodePacked("minPerTx"))] = _minPerTx;
```
akolotov: we definitely don't need `_dailyLimit` abd `minPerTx` parameters
Probably we need to have something similar to `_maxPerTx` to limit gas consumption by one request.

patitonar: In this case `dailyLimit`, `minPerTx` and `maxPerTx` are used to limit the gas consumption. I reused the names of how they are stored to avoid defining new getter and setters.

https://github.com/poanetwork/poa-bridge-contracts/blob/ee1502762dcee3d2a6374103f99ed37aa93728f2/contracts/upgradeable_contracts/arbitrary_message/HomeAMB.sol#L10-L14

Probably `minPerTx` in this case doesn't apply so I should remove it. Regarding `dailyLimit` it was mentioned on the issue that it could be used, but if there is no use case maybe I should remove it too.

So on `requireToPassMessage` only a check for maximum gas will be performed. Sound good?

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
patitonar: The getters and setters are for the bridge mode which will be needed on both contracts to know how to pass the messages. I updated the name from `foreignBridgeMode` to `homeToForeignMode` so it's easier to understand.

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
patitonar: Done

contracts/upgradeable_contracts/arbitrary_message/HomeAMB.sol
```diff
@@ -9,15 +9,15 @@ contract HomeAMB is BasicHomeAMB {
 
     function requireToPassMessage(address _contract, bytes _data, uint256 _gas) public {
         require(keccak256(homeToForeignMode()) == keccak256(SUBSIDIZED_MODE));
-        checkAndUpdateGasLimits(_gas);
+        require(_gas <= maxPerTx());
```
akolotov: I would recommend to extend this check in the way described here: https://github.com/poanetwork/poa-bridge-contracts/issues/73#issuecomment-419936405

patitonar:
@akolotov what should we consider in order to define a proper value for `PASS_MESSAGE_GAS` ?

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
```diff
@@ -0,0 +1,79 @@
...
+    function _defrayAndPassMessage(address _sender, address _contract, bytes _data, uint256 _gas) internal {
+        uint256 fee = PASS_MESSAGE_GAS.add(_gas).mul(tx.gasprice);
+        require(balanceOf(_sender) >= fee);
+        require(address(this).balance >= fee);
+
+        setBalanceOf(_sender, balanceOf(_sender).sub(fee));
+
+        _passMessage(_sender, _contract, _data, _gas);
+
+        msg.sender.transfer(fee);
```
patitonar: @akolotov does it makes sense to emit some kind of `transfer` event after sending fee to validator?

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -0,0 +1,52 @@
...
+contract MessageDelivery {
+    using SafeMath for uint256;
+
+    function requireToPassMessage(address _contract, bytes _data, uint256 _gas) public {
+        require(isMessageDeliverySubsidizedMode());
+        require(_gas >= getMinimumGasUsage(_data) && _gas <= getMaxGasPerTx());
+        emitEventOnMessageRequest(abi.encodePacked(msg.sender, _contract, _gas, uint8(0x00), _data));
+    }
+
+    function requireToPassMessage(address _contract, bytes _data, uint256 _gas, uint256 _gasPrice) public {
```
patitonar: @akolotov Right now there is no check for `gasPrice` specified by the user. There is a potential issue where gas price could be very low and so when the transaction is send to the other network by the validator, it may not be processed and also will affect the next transactions sent by the validator because of the nonce.

One approach to prevent this issue, could be to introduce a new parameter `minGasPrice` which can be the same value as `gasPrice` of the other network.
This parameter could be set on `initialize` method and we can add the following check on `requireToPassMessage`:
```
require(_gasPrice >= minGasPrice());
```

Does it makes sense? Do you have any other ideas?

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -0,0 +1,81 @@
...
+import "../../upgradeability/EternalStorage.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "openzeppelin-solidity/contracts/AddressUtils.sol";
+import "../Validatable.sol";
+import "../BasicBridge.sol";
+
+contract BasicAMB is BasicBridge {
+    bytes internal constant SUBSIDIZED_MODE = bytes(abi.encodePacked("AMB-subsidized-mode"));
+    bytes internal constant DEFRAYAL_MODE = bytes(abi.encodePacked("AMB-defrayal-mode"));
+    bytes32 internal constant HOME_TO_FOREIGN_MODE = keccak256(abi.encodePacked("homeToForeignMode"));
```
akolotov: Is it really needed that the corresponding contract knows its own mode? My understanding that it is more important to know the mode of the opposite site to manage proper set of arguments to relay.

So, my suggestion is to have move functionality related to the mode of Home side to `BasicForeignAMB` directly and the mode of the Foreign side to `BasicHomeAMB`. The method `initialize` of `BasicAMB` will be internal and `Basic...AMB` contracts will call it to perform initialization for common parameters.

Here is an example to illustrate what I mean:
```solidity
contract BasicAMB is BasicBridge {
    bytes32 internal constant SUBSIDIZED_MODE = keccak256(abi.encodePacked("AMB-subsidized-mode"));
    bytes32 internal constant DEFRAYAL_MODE = keccak256(abi.encodePacked("AMB-defrayal-mode"));
    bytes32 internal constant MAX_GAS_PER_TX = keccak256(abi.encodePacked("maxGasPerTx"));

    function _initialize(
        address _validatorContract,
        uint256 _maxGasPerTx,
        uint256 _gasPrice,
        address _owner
    ) internal {
    . . .
    }

    function getBridgeMode() external pure returns (bytes4 _data) {
        return bytes4(keccak256(abi.encodePacked("arbitrary-message-bridge-core")));
    }

    function maxGasPerTx() public view returns (uint256) {
        return uintStorage[MAX_GAS_PER_TX];
    }

     function setMaxGasPerTx(uint256 _maxGasPerTx) external onlyOwner {
        require(_maxGasPerTx > 0);
        uintStorage[MAX_GAS_PER_TX] = _maxGasPerTx;
    }

contract BasicForeignAMB is BasicAMB, MessageDelivery, MessageProcessor {
    bytes32 internal constant FOREIGN_TO_HOME_MODE = keccak256(abi.encodePacked("foreignToHomeMode"));

    function initialize(
        address _validatorContract,
        uint256 _maxGasPerTx,
        uint256 _gasPrice,
        address _owner
    ) public returns (bool) {
        require(!isInitialized());
        _initialize(_validatorContract, _maxGasPerTx, _gasPrice, _requiredBlockConfirmations, _owner);

        bytesStorage[FOREIGN_TO_HOME_MODE] = SUBSIDIZED_MODE;
        setInitialize();

        emit ForeignToHomeModeChanged(SUBSIDIZED_MODE);

        return isInitialized();
    }

     function setSubsidizedModeForForeignToHome() external onlyOwner {
        bytes32Storage[FOREIGN_TO_HOME_MODE] = SUBSIDIZED_MODE;
        emit ForeignToHomeModeChanged(SUBSIDIZED_MODE);
    }

     function setDefrayalModeForForeignToHome() external onlyOwner {
        bytes32Storage[FOREIGN_TO_HOME_MODE] = DEFRAYAL_MODE;
        emit ForeignToHomeModeChanged(DEFRAYAL_MODE);
    }

     function foreignToHomeMode() public view returns (bytes) {
        return bytes32Storage[FOREIGN_TO_HOME_MODE];
    }
. . .
. . .
}
```

Note, that I changed the data type for constants `SUBSIDIZED_MODE` and `DEFRAYAL_MODE` to reduce the gas usage. 
I also added event to notify about the mode change and make the mode by default `SUBSIDIZED_MODE` since the defrayal mode is not ready yet.

By moving the corresponding method to handle the mode to specific contract we also reduce gas usage on the deployment stage.

patitonar: My understanding is that messages that are transferred from home to be executed on foreign could be set to one mode, and messages that are trasferred from foreign to be executed on Home could be set to the same or another mode. Is this correct?

If this is true, for example on Home side, the contract needs to know how to pass the message to foreign when `requireToPassMessage` is called. So in this case it needs to know the mode of the contract on other network to know how will be the message processed. This is done on `MessageDelivery`.

https://github.com/poanetwork/poa-bridge-contracts/blob/c929d2c4083852ff66967164983818495226923f/contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol#L8-L12

On another example, Home receives a message to process from foreign. In this case it needs to know its own mode to be able to run the proper checks. This is done on `MessageProcessor` 

https://github.com/poanetwork/poa-bridge-contracts/blob/c929d2c4083852ff66967164983818495226923f/contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol#L36-L48

Definitions from Home contract:

https://github.com/poanetwork/poa-bridge-contracts/blob/c929d2c4083852ff66967164983818495226923f/contracts/upgradeable_contracts/arbitrary_message/HomeAMB.sol#L13-L23

So at least the way is implemented now it looks like the contract needs to know its own mode too, unless I'm missing something or some way to avoid this.

>Note, that I changed the data type for constants SUBSIDIZED_MODE and DEFRAYAL_MODE to reduce the gas usage.
I also added event to notify about the mode change and make the mode by default SUBSIDIZED_MODE since the defrayal mode is not ready yet.

Thanks, I'll apply these changes

contracts/mocks/Box.sol
```diff
@@ -0,0 +1,81 @@
...
+        for (uint256 i = 0; i < 1000; i++) {
+            a = a + i;
+        }
+    }
+
+    function methodWillFailOnOtherNetwork(address _bridge, address _executor) public {
+        bytes4 methodSelector = this.methodWillFail.selector;
+        bytes memory encodedData = abi.encodeWithSelector(methodSelector);
+        MessageDelivery bridge = MessageDelivery(_bridge);
+        bridge.requireToPassMessage(_executor, encodedData, 141647);
```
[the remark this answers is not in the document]
patitonar: Thanks, implemented here https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/c929d2c4083852ff66967164983818495226923f

contracts/mocks/Box.sol
```diff
@@ -0,0 +1,81 @@
...
+        bridge.requireToPassMessage(_executor, encodedData, 821254, _oracleGasPriceSpeed);
+    }
+
+    function withdrawFromDepositOnOtherNetworkGasPrice(
+        address _recipient,
+        address _bridge,
+        address _executor,
+        uint256 _gasPrice
+    ) public {
+        bytes4 methodSelector = this.withdrawFromDeposit.selector;
```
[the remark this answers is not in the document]
patitonar: Thanks, moved the method to `IAMB` https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/4ad3ee6e7498290544eac390ab56cbb66e99698b

contracts/libraries/ArbitraryMessage.sol
```diff
@@ -0,0 +1,147 @@
+pragma solidity 0.4.24;
+
+import "../interfaces/IBridgeValidators.sol";
+
+library ArbitraryMessage {
+    function recoverAddressFromSignedMessage(bytes signature, bytes message) internal pure returns (address) {
```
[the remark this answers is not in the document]
patitonar: Actually only `hasEnoughValidSignatures` and `hashMessage` were different. To avoid code duplication, I added a new parameter in the methods to properly handle each case https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/d531eafc76be8fee215c0e1d80a278da868a75ec

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
patitonar: >Note, that I changed the data type for constants SUBSIDIZED_MODE and DEFRAYAL_MODE to reduce the gas usage.

Looking at the code, our `EternalStorage` does not provide a`bytes32Storage` mapping. To avoid modifying the storage do you think we can use `uint256` instead of `bytes32` so we can store the information on `uintStorage` and also save gas compairing with the previous `bytes` definition? In that case the variables could be defined as 
```solidity
    uint256 internal constant SUBSIDIZED_MODE = 0;
    uint256 internal constant DEFRAYAL_MODE = 1;
```

akolotov: > In this case it needs to know its own mode to be able to run the proper checks.

Thanks for clarification I was not aware about these checks.

> we can store the information on uintStorage and also save gas compairing with the previous bytes definition

yes, let's use their integer representation then.

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
```diff
@@ -0,0 +1,137 @@
...
+        require(!isAlreadyProcessed(signed));
+        // the check above assumes that the case when the value could be overflew will not happen in the addition operation below
+        signed = signed + 1;
+
+        setNumAffirmationsSigned(hashMsg, signed);
+
+        emit SignedForAffirmation(msg.sender, hashMsg);
+
+        if (signed >= requiredSignatures()) {
+            setNumAffirmationsSigned(hashMsg, markAsProcessed(signed));
```
akolotov: We mark here that the message has been processed. But then we do the same in `processMessage`:
```solidity
    require(!messageProcessed(txHash));
    setMessageProcessed(txHash, true);
```
So, we spend extra gas for the same action.

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
```diff
@@ -0,0 +1,137 @@
...
+        // ensure that `signature` is really `message` signed by `msg.sender`
+        require(msg.sender == Message.recoverAddressFromSignedMessage(signature, message, true));
+        bytes32 hashMsg = keccak256(abi.encodePacked(message));
+        bytes32 hashSender = keccak256(abi.encodePacked(msg.sender, hashMsg));
+
+        uint256 signed = numMessagesSigned(hashMsg);
+        require(!isAlreadyProcessed(signed));
+        // the check above assumes that the case when the value could be overflew
+        // will not happen in the addition operation below
+        signed = signed.add(1);
```
akolotov: We don't use `safeMath` in the similar handling within `executeAffirmation`. Let's keep it consistent.

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
patitonar: Applied changes https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/e7cf4fa8ab20a8a1b3f8b92880caeeb6f5dbee9a

contracts/libraries/Message.sol
```diff
@@ -68,40 +68,49 @@ library Message {
         return 104;
     }
 
-    function recoverAddressFromSignedMessage(bytes signature, bytes message) internal pure returns (address) {
+    function recoverAddressFromSignedMessage(bytes signature, bytes message, bool variableMsgLength)
```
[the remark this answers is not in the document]
patitonar: Thanks for the suggestion, renamed the parameter https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/6d96e6cd8855dfa969c328ecb0bed619f3c89dff

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -0,0 +1,81 @@
...
+    function foreignToHomeMode() public view returns (bytes) {
+        return bytesStorage[FOREIGN_TO_HOME_MODE];
+    }
+
+    function maxGasPerTx() public view returns (uint256) {
+        return uintStorage[MAX_GAS_PER_TX];
+    }
+
+    function setMaxGasPerTx(uint256 _maxGasPerTx) external onlyOwner {
+        require(_maxGasPerTx > 0);
```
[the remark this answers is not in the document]
patitonar: Yes I think we can use this parameter for such behavior https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/ae8c0596d3b6d06cebba9f4d439de865be395040

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
```diff
@@ -0,0 +1,23 @@
...
+import "./MessageProcessor.sol";
+
+contract BasicForeignAMB is BasicAMB, MessageDelivery, MessageProcessor {
+    function executeSignatures(bytes _data, uint8[] vs, bytes32[] rs, bytes32[] ss) external {
+        Message.hasEnoughValidSignatures(_data, vs, rs, ss, validatorContract(), true);
+
+        processMessage(_data, true);
+    }
+
+    function relayedMessages(bytes32 _txHash) public view returns (bool) {
```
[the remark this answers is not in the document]
patitonar: Good catch! moved the common logic here https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/cf3e507b4799b6b4b426b9fa2956d6f5abd994d4

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
patitonar: Fixed here https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/1613a10420cbe938a23a38847a4235aec0af494b

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
patitonar: Made a refactor to avoid `processMessage` checking for this. Please take a look https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/7f8307554658a61a024de086e2cc65b40ec149b1

akolotov: should we really have two `processMessage` methods?

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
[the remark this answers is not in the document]
akolotov: does it make sense to add a test to check that the message relay is not possible if `maxGasPerTx` is zero?

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -47,27 +52,27 @@ contract BasicAMB is BasicBridge {
...
     }
 
     function setDefrayalModeForHomeToForeign() external onlyOwner {
-        bytesStorage[HOME_TO_FOREIGN_MODE] = DEFRAYAL_MODE;
+        uintStorage[HOME_TO_FOREIGN_MODE] = DEFRAYAL_MODE;
     }
 
     function setSubsidizedModeForForeignToHome() external onlyOwner {
-        bytesStorage[FOREIGN_TO_HOME_MODE] = SUBSIDIZED_MODE;
+        uintStorage[FOREIGN_TO_HOME_MODE] = SUBSIDIZED_MODE;
```
akolotov: does it make sense to raise the event `ForeignToHomeModeChanged` here?

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -47,27 +52,27 @@ contract BasicAMB is BasicBridge {
     }
 
     function setSubsidizedModeForHomeToForeign() external onlyOwner {
-        bytesStorage[HOME_TO_FOREIGN_MODE] = SUBSIDIZED_MODE;
+        uintStorage[HOME_TO_FOREIGN_MODE] = SUBSIDIZED_MODE;
```
akolotov: does it make sense to raise the event `HomeToForeignModeChanged` here?

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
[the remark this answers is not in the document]
patitonar: Added a test case for this https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/87d9cc97c8c2d2dbd21bdcb686fdfe858414cdf1

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
patitonar: Moved the logic so we have only one `processMessage` and now `BasicHomeAMB` works similar to `BasicForeignAMB` https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/a23dfba06c5eb4f48c00a855231b61160e25eeb9

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
patitonar: Yes, updated the setter for this mode to emit the event https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/f9efd5239261b48d4842d1b81e96fbfcd028b98a

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
patitonar: Added events for the setters https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/cd236e08095fd254a32e661bde88a86d394824f1

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
```diff
@@ -0,0 +1,27 @@
...
+
+import "../../libraries/Message.sol";
+import "../../libraries/ArbitraryMessage.sol";
+import "./BasicAMB.sol";
+import "./MessageDelivery.sol";
+import "./MessageProcessor.sol";
+import "../MessageRelay.sol";
+
+contract BasicForeignAMB is BasicAMB, MessageRelay, MessageDelivery, MessageProcessor {
+    function executeSignatures(bytes _data, uint8[] vs, bytes32[] rs, bytes32[] ss) external {
```
akolotov: I would like to get you opinion here. If we consider as arrays are encoded to be passed as method parameters you will see the following (I took the previous implementation of `executeSignatures` where the message was at the end of the parameters list):
* The first 32 bytes point to the place where `uint8[] vs` is stored.
* The second 32 bytes point to the place where `bytes32[] rs` is stored.
* The third 32 bytes point to the place where `bytes32[] ss` is stored.
* The fourth 32 bytes point to the place where the message begins.
* `uint8[] vs` takes 4*32 bytes, where the first 32 bytes store 3.
* `bytes32[] rs` takes 4*32 bytes, where the first 32 bytes store 3.
* `bytes32[] ss` takes 4*32 bytes, where the first 32 bytes store 3.

This is an [example](https://blockscout.com/eth/mainnet/tx/0x5b0b29bbe1604d5e9bad3bf9af2a615eaffdd946b2569dbad50a4c6937936b1b):
```
0x232a2c1d
000 - 031: 0000000000000000000000000000000000000000000000000000000000000080 - 128
032 - 063: 0000000000000000000000000000000000000000000000000000000000000100 - 256
064 - 095: 0000000000000000000000000000000000000000000000000000000000000180 - 384
096 - 127: 0000000000000000000000000000000000000000000000000000000000000200 - 512
128 - 159: 0000000000000000000000000000000000000000000000000000000000000003 - 3
160 - 191: 000000000000000000000000000000000000000000000000000000000000001c
192 - 223: 000000000000000000000000000000000000000000000000000000000000001b
224 - 255: 000000000000000000000000000000000000000000000000000000000000001c
256 - 287: 0000000000000000000000000000000000000000000000000000000000000003 - 3
288 - 319: e7dfc36bdc721319d943a2d70fcedfac456ca9395ec50404c24178fe08b48837
320 - 351: b85ef14a1a117d9a0781eb07dd0e53e445f14d24629682b7413f23365dcb4f8e
352 - 383: 13bbac2ed6b40a0be695f25b125f610ef6305bfea1cb7651b2d2cb89c45920d9
384 - 415: 0000000000000000000000000000000000000000000000000000000000000003 - 3
416 - 447: 0fba05cf0ba056230150d84975c46bfa961e3181e12efdaa4614de332f6ecf45
448 - 479: 12c95da4c7034a92ad6ee577925e9d9ebf48650c992d5f016193324d4766b56a
480 - 511: 017a79312986afea6350b97a48aec7a6b9956ec8eef2b6e9c48a05da709ef59f
512 - 543: 0000000000000000000000000000000000000000000000000000000000000068 - 104
b96432f3307a72266ab210a8536726d666aa54d0
0000000000000000000000000000000000000000000000001aec43b8b0420000
6eef8dc1865bd54ece19ea9be39996db7e5a44c592f74be7f28d573d2dbf8d99
4aa42145aa6ebf72e164c9bbc74fbd3788045016000000000000000000000000000000000000000000000000
```

Do you think if we pack the arrays before the sending will it consume less gas? I am interesting how much gas it could take to unpack them.

For example, we can use the following method to pack the data: `length + vs1 + vs2 + vs3 + rs1 + rs2 + rs3 + ss1 + ss2 + ss3`. If we pack length and `vs` components as `uint8` it will reduce the transaction input by `3*32+60=156` bytes in case of 3 signatures. It consumes 816 gas.

contracts/upgradeable_contracts/arbitrary_message/ForeignAMB.sol
```diff
@@ -0,0 +1,28 @@
+pragma solidity 0.4.24;
+
+import "./BasicForeignAMB.sol";
+
+contract ForeignAMB is BasicForeignAMB {
+    event UserRequestForAffirmation(bytes encodedData);
+    event RelayedMessage(address sender, address executor, bytes32 transactionHash, bool status);
+
+    function getMaxGasPerTx() internal returns (uint256) {
```
akolotov: Could we invoke `maxGasPerTx` directly instead of `getMaxGasPerTx` in that places where it is used?

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -0,0 +1,43 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+contract MessageDelivery {
+    using SafeMath for uint256;
+
+    function requireToPassMessage(address _contract, bytes _data, uint256 _gas) public {
+        require(isMessageDeliverySubsidizedMode());
+        require(_gas >= getMinimumGasUsage(_data) && _gas <= getMaxGasPerTx());
```
akolotov: can we use here `maxGasPerTx`?

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
akolotov: No, it is not necessary since this transfer assumes to just cover the message sender expenses to send this transaction

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    function _passMessage(address _sender, address _contract, bytes _data, uint256 _gas) internal returns (bool) {
+        if (_contract == address(this)) {
+            //Special check to handle invocation of withdrawFromDeposit
+            if (isWithdrawFromDepositSelector(_data)) {
+                setAccountForAction(_sender);
+            }
+        }
+        setMessageSender(_sender);
+        bool status = _contract.call.gas(_gas)(_data);
```
akolotov: Is it safe if we don't handle the case when the method returns some data after invocation?

contracts/libraries/ArbitraryMessage.sol
```diff
@@ -0,0 +1,82 @@
...
+    // so we can and have to start reading recipient at offset 20 instead of 32.
+    // if we were to read at 32 the address would contain part of value and be corrupted.
+    // when reading from offset 20 mload will read 12 zero bytes followed
+    // by the 20 recipient address bytes and correctly convert it into an address.
+    // this saves some storage/gas over the alternative solution
+    // which is padding address to 32 bytes and reading recipient at offset 32.
+    // for more details see discussion in:
+    // https://github.com/paritytech/parity-bridge/issues/61
+
+    function unpackData(bytes _data, bool applyDataOffset)
```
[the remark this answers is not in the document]
patitonar: To get the `data` parameter from the bridged message in `unpackData` the following line is used to copy values directly from `calldata`:

```solidity
calldatacopy(dataptr, add(add(68, srcdataptr), dataOffset), datasize)
```

While implementing this I found that when `unpackData` was called on home side from the method `function executeAffirmation(bytes message) external` it was working correctly. But when it was being called from foreign side from method `function executeSignatures(uint8[] vs, bytes32[] rs, bytes32[] ss, bytes message)` the `data` parameter from the message was being filled with incorrects values. Also the values assigned to `data` were different depending on the amount of validator signatures I used on every call. 

So that's why I changed the order of the parameters to:
`function executeSignatures(bytes _data, uint8[] vs, bytes32[] rs, bytes32[] ss) external`

After that change, independently of the numbers of validator signatures, the `data` parameter was being filled with the same incorrect values on every call. I found that there was an offset of 9 positions on the values that were being read from `calldata`, which also could be different depending on the `dataType` of the message (if gas prices values or gas price oracle speed were in the message).

I'm not sure exactly how the parameters from the external methods are managed for `calldata` but it seems to be affecting the position of the information inside `calldata` that's why the previous changes (parameter order change and `applyDataOffset` parameter) were introduced to make `unpackData` method work for home and foreign methods.

contracts/libraries/ArbitraryMessage.sol
```diff
@@ -0,0 +1,82 @@
...
+
+import "../interfaces/IBridgeValidators.sol";
+import "./Message.sol";
+
+library ArbitraryMessage {
+    // layout of message :: bytes:
+    // offset  0: 32 bytes :: uint256 - message length
+    // offset 32: 20 bytes :: address - sender address
+    // offset 52: 20 bytes :: address - executor contract
+    // offset 72: 32 bytes :: bytes32 txHash
```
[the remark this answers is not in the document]
patitonar: The reason to inject the transactions hash into the message is to have an unique message that is mark as processed so it can't be processed again. The same message could be generated from an user multiple times so adding the transactions hash make it possible to continue processing the new messages.  Also this parameter is included on the events `AffirmationCompleted` and `RelayedMessage` for traceability.

I first considered to add the `txHash` at the end of the message, but to avoid calculations in `unpackData` method to know in which position the `txHash` will be, since it will depend on the `dataType`, I added it in the middle similar to the message that is constructed based on `UserRequestForSignature` as you mention. Adding the `txHash` at the beginning is a good option too, If you think it will be better I can change it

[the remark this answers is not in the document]
akolotov: > The reason to inject the transactions hash into the message is to have an unique message that is mark as processed so it can't be processed again.

Yes, I know this.

> Adding the txHash at the beginning is a good option too, If you think it will be better I can change it

Let's do this in order to simplify the oracles' work to compose the message.

contracts/libraries/ArbitraryMessage.sol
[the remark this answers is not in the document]
akolotov: Hmm. Quite interesting behavior. Let's have it as is so far. Probably later we can investigate it deeply.

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
patitonar: This looks interesting, we can try to run some tests to know if it worth packing the data and unpacking it in the contracts based on the gas usage. This could be used for executing signatures for amb bridge but also for other bridge modes, does it makes sense to create a separate issue for this?

akolotov: I think we can implement this approach in the AMB first. So far, there is no necessity to optimize the original token bridge contracts.

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
patitonar: I'm not sure I fully understood the question, `contract.call` will return `true` if execution is correct and `false` if it is reverted and we show this result in the emitted event later. If we want to get the returned data from the method we should use the assembly call, but I'm not sure why we need it

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    function setMessagesSigned(bytes32 _hash, bool _status) internal {
+        boolStorage[keccak256(abi.encodePacked("messagesSigned", _hash))] = _status;
+    }
+
+    function messages(bytes32 _hash) internal view returns (bytes) {
+        return bytesStorage[keccak256(abi.encodePacked("messages", _hash))];
+    }
+
+    function signatures(bytes32 _hash) internal view returns (bytes) {
```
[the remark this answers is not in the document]
patitonar: Thanks, removed https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/b4810f5698e9c78b790390a443ead15f1a0c51d4

contracts/upgradeable_contracts/arbitrary_message/ForeignAMB.sol
patitonar: Yes, that makes sense. https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/e1eb585dd20c84e8686717424225c25bfc1ea3d7

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
patitonar: Fixed here https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/e1eb585dd20c84e8686717424225c25bfc1ea3d7

contracts/libraries/ArbitraryMessage.sol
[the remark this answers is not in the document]
patitonar: Update message to include `txHash` in the beginning https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/1987b9ab285577ef263bece6488b48e204879455

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
akolotov: If you look at the `call` function [description](https://solidity.readthedocs.io/en/v0.5.11/units-and-global-variables.html#members-of-address-types) you will see that it returns `(bool, bytes memory)`. But in the discussed code we are expecting only a bool value. So, the question is if the calling method returns something will this be skipped successfully even if we don't allocate memory for this data explicitly.

patitonar: Oh I see. Actually, what you mention about `call` function is valid for solidity version > `0.5.0`. For [version `0.4.24`](https://solidity.readthedocs.io/en/v0.4.24/units-and-global-variables.html#address-related) `call` function only returns `(bool)`, so I think we don't need to worry about a possible returned data

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
patitonar: Following the idea of packing data as `length + vs1 + vs2 + vs3 + rs1 + rs2 + rs3 + ss1 + ss2 + ss3`
I added an implementation to unpack signature data which I combined with `hasEnoughValidSignatures` method to avoid executing two `for` loops. Probably this can be improved to save more gas.

```solidity
function hasEnoughValidSignaturesAMB(
    bytes _message,
    bytes _signatures,
    IBridgeValidators _validatorContract,
    bool isAMBMessage
) internal view {
    require(isAMBMessage || (!isAMBMessage && isMessageValid(_message)));
    uint256 requiredSignatures = _validatorContract.requiredSignatures();
    uint8 amount;
    assembly {
        amount := mload(add(_signatures, 1))
    }
    require(amount >= requiredSignatures);
    bytes32 hash = hashMessage(_message, isAMBMessage);
    address[] memory encounteredAddresses = new address[](requiredSignatures);

    for (uint256 i = 0; i < requiredSignatures; i++) {
        uint8 v;
        bytes32 r;
        bytes32 s;
        uint256 posr = 33 + amount + 32 * i;
        uint256 poss = posr + 32 * amount;
        assembly {
            v := mload(add(_signatures, add(2, i)))
            r := mload(add(_signatures, posr))
            s := mload(add(_signatures, poss))
        }

        address recoveredAddress = ecrecover(hash, v, r, s);
        require(_validatorContract.isValidator(recoveredAddress));
        require(!addressArrayContains(encounteredAddresses, recoveredAddress));
        encounteredAddresses[i] = recoveredAddress;
    }
}
```

To compare the gas usage, I added a new method and to simplify the comparison, I removed all the logic for processing the message and only tested checking the signatures:

Method with array signatures:
```solidity
function executeSignatures(bytes _data, uint8[] vs, bytes32[] rs, bytes32[] ss) external {
    Message.hasEnoughValidSignatures(_data, vs, rs, ss, validatorContract(), true);
}
```

Method with bytes signatures:
```solidity
function executeSignaturesAMB(bytes _data, bytes _signatures) external {
    Message.hasEnoughValidSignaturesAMB(_data, _signatures, validatorContract(), true);
}
```

I run tests with signatures from 1, 3 and 5 validators, the following were the gas consumed by the transactions:

|Method | 1 | 3 | 5 |
|-----|-----|-----|-----|
array signatures | 50777 | 77334 | 107244
bytes signatures | 49788 | 75624 | 105321

From the results, it looks like packing the signatures has a little better gas usage. Maybe if we can optimize more the method `hasEnoughValidSignaturesAMB` we can get a better result.

Do you think it is worth implementing this logic? Logic on oracle side will need to be updated to reflect this change for AMB mode

akolotov: Cool! Thanks for research! What your expectations how much it could take to introduce changes on both oracle and contracts is?

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
akolotov: Thanks! Sorry for my inattention.

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
patitonar: Well, this shouldn't take big efforts. To perform this tests I had to create some helper methods to pack the signature data so I can reuse this for the oracle (plus adding some unit tests) and on the contracts I need to properly include the new already implemented method and update the units tests

patitonar: Applied the changes on the contracts side https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/0309b3d3739bd9c21b764c41f17e6c18cc506ec3

akolotov: Thanks!

patitonar:
Based on comment https://github.com/poanetwork/poa-bridge-contracts/issues/267#issuecomment-522066864 I implemented https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/b6f06e4b003a4417a84ea9757203b04005eccc17

patitonar:
@akolotov Fixed the conflicts

contracts/mocks/IAMB.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity 0.4.24;
+
+interface IAMB {
+    function messageSender() external view returns (address);
+    function transactionHash() external view returns (bytes32);
+    function withdrawFromDeposit(address _recipient) external;
+    function requireToPassMessage(address _contract, bytes _data, uint256 _gas) public;
```
akolotov: We need to leave only the methods that work in the `subsidized` mode.

contracts/mocks/Box.sol
```diff
@@ -0,0 +1,72 @@
...
+        IAMB(_bridge).requireToPassMessage(_executor, encodedData, 1000);
+    }
+
+    function setValueOnOtherNetwork(uint256 _i, address _bridge, address _executor) public {
+        bytes4 methodSelector = this.setValue.selector;
+        bytes memory encodedData = abi.encodeWithSelector(methodSelector, _i);
+        IAMB(_bridge).requireToPassMessage(_executor, encodedData, 141647);
+    }
+
+    function setValueOnOtherNetworkGasPrice(uint256 _i, address _bridge, address _executor, uint256 _gasPrice) public {
```
akolotov: We need to leave only the methods that work in the `subsidized` mode.

contracts/mocks/IAMB.sol
patitonar: Thanks. Fixed https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/2655b4dbdabf0f6dd99ec074e50db3e5628ec719

contracts/mocks/Box.sol
patitonar: Done https://github.com/poanetwork/poa-bridge-contracts/pull/77/commits/2655b4dbdabf0f6dd99ec074e50db3e5628ec719

PR #80, opened 2018-09-17
Add bridge mode field and interfaces version support

### LINKED ISSUE #67: A field in bridge contracts to support type of bridge and version tracking
Currently the bridge contracts does not provide any facility to get information about interfaces supported by particular deployed contract.
The functionality to provide mode of the bridge by contracts will be very useful in order to allow Dapp like `bridge-ui` and `bridge-monitor` (or even `bridge-nodejs`) to adjust automatically their functionality without additional manual configuration.

The functionality to provide list of supported interfaces by the bridge by contracts will be useful in order to load corresponding version of ABIs or turn on/off invocation of corresponding methods from contracts.

So, it make sense to consider to add a method `getBridgeMode()` to return the mode of the bridge and  a method `getBridgeInterfacesVersion()`. An integer value returned by `getBridgeInterfacesVersion()` should be increased manually in the repo every time when backward compatibility is broken: selectors of methods or events were changed, functionality of the methods with the same name and the same list of parameters modified dramatically.

patitonar:
Closes #67 

Add `getBridgeMode` method on bridge contracts and add `getBridgeInterfacesVersion`  method on bridge, ERC677 token  and validators contract.

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -14,7 +14,7 @@ contract ERC677BridgeToken is
     BurnableToken,
     MintableToken {
 
-    Version.Version public getTokenInterfacesVersion = Version.Version(1, 0, 0);
+    Version.Version public getTokenInterfacesVersion = Version.Version(2, 0, 0);
 
     constructor(
         string _name,
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -14,7 +14,7 @@ contract ERC677BridgeToken is
     BurnableToken,
     MintableToken {
 
-    Version.Version public getBridgeInterfacesVersion = Version.Version(1, 0, 0);
+    Version.Version public getTokenInterfacesVersion = Version.Version(1, 0, 0);
 
     constructor(
         string _name,
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -14,7 +14,7 @@ contract ERC677BridgeToken is
     BurnableToken,
     MintableToken {
 
-    Version.Version public getInterfacesVersion = Version.Version(1, 0, 0);
+    Version.Version public getBridgeInterfacesVersion = Version.Version(1, 0, 0);
 
     constructor(
         string _name,
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -5,6 +5,7 @@ import "openzeppelin-solidity/contracts/token/ERC20/MintableToken.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/DetailedERC20.sol";
 import "./IBurnableMintableERC677Token.sol";
 import "./ERC677Receiver.sol";
+import "./libraries/Version.sol";
 
 
 contract ERC677BridgeToken is
@@ -12,6 +13,9 @@ contract ERC677BridgeToken is
     DetailedERC20,
     BurnableToken,
     MintableToken {
+
+    Version.Version public getInterfacesVersion = Version.Version(1, 0, 0);
+
     constructor(
         string _name,
         string _symbol,
--- a/contracts/upgradeable_contracts/BasicBridge.sol
+++ b/contracts/upgradeable_contracts/BasicBridge.sol
@@ -16,14 +16,6 @@ contract BasicBridge is EternalStorage, Validatable {
     event RequiredBlockConfirmationChanged(uint256 requiredBlockConfirmations);
     event DailyLimitChanged(uint256 newLimit);
 
-    function bridgeMode() public view returns(bytes) {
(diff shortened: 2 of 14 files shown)


contracts/ERC677BridgeToken.sol
```diff
@@ -5,13 +5,17 @@ import "openzeppelin-solidity/contracts/token/ERC20/MintableToken.sol";
...
+import "./libraries/Version.sol";
 
 
 contract ERC677BridgeToken is
     IBurnableMintableERC677Token,
     DetailedERC20,
     BurnableToken,
     MintableToken {
+
+    Version.Version public getBridgeInterfacesVersion = Version.Version(1, 0, 0);
```
akolotov: `getTokenInterfacesVersion` should be here.

Also, it seems that the code was changed significantly from the version we have in https://github.com/poanetwork/poa-bridge-contracts/releases/tag/1.0. So, lets announce the version `2.0.0` here.

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -4,10 +4,14 @@ import "./Ownable.sol";
 import "../IBridgeValidators.sol";
 import "../libraries/SafeMath.sol";
 import "../upgradeability/EternalStorage.sol";
+import "../libraries/Version.sol";
 
 
 contract BridgeValidators is IBridgeValidators, EternalStorage, Ownable {
     using SafeMath for uint256;
+
+    Version.Version public getBridgeInterfacesVersion = Version.Version(1, 0, 0);
```
akolotov: `getBrdigeValidatorsInterfacesVersion` should be here.

Also, it seems that the code was changed slightly from the version we have in https://github.com/poanetwork/poa-bridge-contracts/releases/tag/1.0. But lets inline the version of validators contract with the version of bridge contracts - `2.0.0` should be here.

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
```diff
@@ -11,6 +11,8 @@ import "../ERC677Bridge.sol";
 
 contract HomeBridgeErcToErc is ERC677Receiver, EternalStorage, BasicBridge, BasicHomeBridge, ERC677Bridge {
 
+    bytes internal constant BRIDGE_MODE = hex"ba4690f5"; // 4 bytes of keccak256('erc-to-erc-core')
```
akolotov: This variable is not needed, I believe.

contracts/upgradeable_contracts/BasicBridge.sol
```diff
@@ -2,12 +2,16 @@ pragma solidity 0.4.24;
...
 import "../libraries/SafeMath.sol";
+import "../libraries/Version.sol";
 import "./Validatable.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20Basic.sol";
 
 
 contract BasicBridge is EternalStorage, Validatable {
     using SafeMath for uint256;
+
+    Version.Version public getBridgeInterfacesVersion = Version.Version(1, 0, 0);
```
akolotov: According to the contract releases (https://github.com/poanetwork/poa-bridge-contracts/releases), lets have `2.0.0` here.

contracts/ERC677BridgeToken.sol
patitonar: Updated name to `getTokenInterfacesVersion` and version to `2.0.0`

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Updated name to `getBridgeValidatorsInterfacesVersion` and version to `2.0.0`

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
patitonar: Thanks! I removed the variable

contracts/upgradeable_contracts/BasicBridge.sol
patitonar: Updated version to `2.0.0`

fvictorio:
It seems to me that we could have some "mixin" contract for the version, something like:

```solidity
contract BridgeVersion {
  Version.Version public getVersion = Version.Version(2, 0, 0);
}
```

Unless there's reason to think that the different contracts could change their versions independently.

Similarly for the bridge modes:

```solidity
contract NativeToErcMode {
  function getBridgeMode() public pure returns(bytes4 _data) {
    return bytes4(keccak256(abi.encodePacked("native-to-erc-core")));
  }
```

and something similar for `ErcToErc`.

But I'm OK with this anyway.

akolotov:
If we consider the process how bugs could be fixed in the production, the different sides of the bridge could have different `patch` component indeed. For example, in the past we had a case when the fix was applied to the contract on `Foreign` side only.

PR #91, opened 2018-10-03
Add gas price on initialization of Foreign erc-to-erc mode

### LINKED ISSUE #86: Default gas price is not initialized for ForeignBridge contract in `erc-to-erc` mode
Due to incorrect copy of ForeingBridge implementation when `erc-to-erc` bridge mode was introduced, the following line was missed:

https://github.com/poanetwork/poa-bridge-contracts/blob/fef87310b523be908b16295d777fd0215878dad8/contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol#L35

So, it caused uninitialized `gasPrice` value in the storage after deployment and configuration of ForeignBridge contract:

https://github.com/poanetwork/poa-bridge-contracts/blob/fef87310b523be908b16295d777fd0215878dad8/contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol#L15-L29

patitonar:
Closes #86

### DIFF
--- a/contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
+++ b/contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
@@ -15,15 +15,18 @@ contract ForeignBridgeErcToErc is BasicBridge, BasicForeignBridge {
     function initialize(
         address _validatorContract,
         address _erc20token,
-        uint256 _requiredBlockConfirmations
+        uint256 _requiredBlockConfirmations,
+        uint256 _gasPrice
     ) public returns(bool) {
         require(!isInitialized(), "already initialized");
         require(_validatorContract != address(0), "address cannot be empty");
         require(_requiredBlockConfirmations != 0, "requiredBlockConfirmations cannot be 0");
+        require(_gasPrice > 0);
         addressStorage[keccak256(abi.encodePacked("validatorContract"))] = _validatorContract;
         setErc20token(_erc20token);
         uintStorage[keccak256(abi.encodePacked("deployedAtBlock"))] = block.number;
         uintStorage[keccak256(abi.encodePacked("requiredBlockConfirmations"))] = _requiredBlockConfirmations;
+        uintStorage[keccak256(abi.encodePacked("gasPrice"))] = _gasPrice;
         setInitialize(true);
         return isInitialized();
     }
--- a/contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
+++ b/contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
@@ -18,9 +18,9 @@ contract ForeignBridgeErcToErc is BasicBridge, BasicForeignBridge {
         uint256 _requiredBlockConfirmations,
         uint256 _gasPrice
     ) public returns(bool) {
-        require(!isInitialized(), "already initialized");
-        require(_validatorContract != address(0), "address cannot be empty");
-        require(_requiredBlockConfirmations != 0, "requiredBlockConfirmations cannot be 0");
+        require(!isInitialized());
+        require(_validatorContract != address(0));
+        require(_requiredBlockConfirmations != 0);
         require(_gasPrice > 0);
         addressStorage[keccak256(abi.encodePacked("validatorContract"))] = _validatorContract;
         setErc20token(_erc20token);
--- a/test/erc_to_erc/foreign_bridge.test.js
+++ b/test/erc_to_erc/foreign_bridge.test.js
@@ -32,6 +32,11 @@ contract('ForeignBridge_ERC20_to_ERC20', async (accounts) => {
 
       await foreignBridge.initialize(validatorContract.address, token.address, requireBlockConfirmations).should.be.rejectedWith(INVALID_ARGUMENTS);
 
+      await foreignBridge.initialize(ZERO_ADDRESS, token.address, requireBlockConfirmations, gasPrice).should.be.rejectedWith(ERROR_MSG);
+      await foreignBridge.initialize(validatorContract.address, ZERO_ADDRESS, requireBlockConfirmations, gasPrice).should.be.rejectedWith(ERROR_MSG);
+      await foreignBridge.initialize(validatorContract.address, token.address, 0, gasPrice).should.be.rejectedWith(ERROR_MSG);
+      await foreignBridge.initialize(validatorContract.address, token.address, requireBlockConfirmations, 0).should.be.rejectedWith(ERROR_MSG);
+
       await foreignBridge.initialize(validatorContract.address, token.address, requireBlockConfirmations, gasPrice);
 
       token.address.should.be.equal(await foreignBridge.erc20token());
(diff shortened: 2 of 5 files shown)


test/erc_to_erc/foreign_bridge.test.js
```diff
@@ -28,14 +29,16 @@ contract('ForeignBridge_ERC20_to_ERC20', async (accounts) => {
       '0'.should.be.bignumber.equal(await foreignBridge.deployedAtBlock())
       false.should.be.equal(await foreignBridge.isInitialized())
       '0'.should.be.bignumber.equal(await foreignBridge.requiredBlockConfirmations())
-      await foreignBridge.initialize(validatorContract.address, token.address, requireBlockConfirmations);
+      await foreignBridge.initialize(validatorContract.address, token.address, requireBlockConfirmations, gasPrice);
```
akolotov: Does it make sense to test that the bridge will not be initialized if incorrect parameters were sent to `initialize()` method?

contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
```diff
@@ -15,15 +15,18 @@ contract ForeignBridgeErcToErc is BasicBridge, BasicForeignBridge {
     function initialize(
         address _validatorContract,
         address _erc20token,
-        uint256 _requiredBlockConfirmations
+        uint256 _requiredBlockConfirmations,
+        uint256 _gasPrice
     ) public returns(bool) {
         require(!isInitialized(), "already initialized");
         require(_validatorContract != address(0), "address cannot be empty");
```
akolotov: Let's be consistent and return `revert` reason in all `require` in the `initialize` method.

My suggestion is to use values as numbers instead of strings since it should reduce the gas usage. You could do simple experiment in order to determine how gas consumption is reduced if a number will be used in `require(_validatorContract != address(0), "address cannot be empty");` for example. You need to consider how the gas usage differs for both the contract deployment and the method invocation). If the difference is significant let's use numbers.

patitonar: I ran some tests to determine gas consumption, I deployed and call `initialize` method with the following variations on **Kovan** network:
- No reason: `require(_validatorContract != address(0));`
- Reason with number:  `require(_validatorContract != address(0), "1");`
- Reason with text: `require(_validatorContract != address(0), "address cannot be empty");`

**Results**:

Deploy:
- No reason - Reason with number: `+ 22701`
- Reason with number - Reason with text: `+ 1344`

Initialize method failed (called with address(0) for validatorContract parameter):
- No reason - Reason with number: `+ 102`
- Reason with number - Reason with text: `+ 0`

Initialize method success:
- No reason - Reason with number: `+ 0`
- Reason with number - Reason with text: `+ 0`

From the results, it seems that avoiding using reason on `require` saves a lot of gas on deployment and very little on calling the method. Changing a full text to a number have some impact on deployment I guess depending on the string length.

The other contracts doesn't use reason on `require`, except foreign erc20-to-native which was built based on this contract. Also it seems that reason string is not yet fully supported on toolings (https://github.com/ethereum/web3.js/issues/1707) so I think that it could be a good idea to remove the reason string from these two contracts. What do you think @akolotov ? 

**Details from the tests:**

`require(_validatorContract != address(0), "address cannot be empty");`

Deploy
https://kovan.etherscan.io/tx/0x42cb3448e31c2a873b77869759ddc92698cf9d89004e3d989a6952eb392f9d07
Gas Used By Transaction:2530163

Success initialization
https://kovan.etherscan.io/tx/0xe9d150b24edeb3007158d1997b77f19c690ab005c2e1e97deb0638e8799e581a
Gas Used By Transaction:152124

Failed initialization with zero_address on validators
https://kovan.etherscan.io/tx/0x1edcb0d738807b049df726c00ecb53b85b6d12292f9d2c83f252489e3e845f03
Gas Used By Transaction:26487

--------------------------

`require(_validatorContract != address(0), "1");`

Deploy
https://kovan.etherscan.io/tx/0x44690aa21eca360a28b85b6bacae44e8f11a603d28859f6115d70232df461295
Gas Used By Transaction:2528819

Success initialization
https://kovan.etherscan.io/tx/0x31b57c073d60820905a6b5db1e01a6cf28106f81a57862e23349d45df0fd2057
Gas Used By Transaction:152124

Failed initialization with zero_address on validators
https://kovan.etherscan.io/tx/0x74fbff2f43ba2d5dfe3b09e4c100265a15612d7bc6c20774b5c0a798d168f167
Gas Used By Transaction:26487

--------------------------

`require(_validatorContract != address(0));`

Deploy
https://kovan.etherscan.io/tx/0xbd67b40a0fba02aeac507535c8a33b61530733d56cde4233bc09b3b2532fba63
Gas Used By Transaction:2506118

Success initialization
https://kovan.etherscan.io/tx/0x539066f85a9a8d392f61bf15142d3322d49d669ad5bf336bc3fc1060babd846e
Gas Used By Transaction:152124

Failed initialization with zero_address on validators
https://kovan.etherscan.io/tx/0xc7f62f1abdaff1362a1baf282ecdc00fdd992ccb6b19b66580bd08522ec8d38e
Gas Used By Transaction:26385

test/erc_to_erc/foreign_bridge.test.js
patitonar: Done

test/erc_to_erc/foreign_bridge.test.js
```diff
@@ -29,6 +29,9 @@ contract('ForeignBridge_ERC20_to_ERC20', async (accounts) => {
       '0'.should.be.bignumber.equal(await foreignBridge.deployedAtBlock())
       false.should.be.equal(await foreignBridge.isInitialized())
       '0'.should.be.bignumber.equal(await foreignBridge.requiredBlockConfirmations())
+
+      await foreignBridge.initialize(validatorContract.address, token.address, requireBlockConfirmations).should.be.rejectedWith(INVALID_ARGUMENTS);
```
akolotov: Does it makes sense to cover scenarios?
* `validatorContract` is zero, or it is not a contract address;
* `token` is zero, or it is not a contract address;
* `requireBlockConfirmations` is zero;
* `gasPrice` is zero.

contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
akolotov: Great! I vote for removing the revert reason.

test/erc_to_erc/foreign_bridge.test.js
patitonar: Added tests to cover those scenarios except when `validatorContract` and `token` are not contract address. Those kind of checks are not performed on any bridge contract on initialization. Should we open an issue to add those checks?

contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
patitonar: Done

test/erc_to_erc/foreign_bridge.test.js
akolotov: > Should we open an issue to add those checks?

Yes, please open it.

patitonar:
@akolotov can we merge this PR?

Issue #47, opened 2018-06-20
Gas usage benchmarks for bridge contracts


akolotov:
Since it is critical to have clear understanding how much gas consumed by the bridge contracts in different scenarios and different deployments it is necessary to perform the corresponding measurements.

Goals are the following:
1. To prepare a sheet in the bridge documentation with description of gas consumed on Home and Foreign side in different scenarios.
2. To identify potential places in the code where gas consumption could be optimized.

patitonar:
@akolotov As a first step to get knowledge of gas consumed by bridge contracts I think we can add https://github.com/cgewecke/eth-gas-reporter which is a library to generate gas usage reports of methods and deployments, based on our unit tests. What do you think?

akolotov:
Looks interesting. Could you estimate how much effort it will take to integrate this? I would avoid spending much time on this issue. From my perspective it is one-time-run every time we add new bridge mode and it will not take more than 2 days.

patitonar:
Installation is pretty straightforward, it took me 10 minutes. If you are OK I can open a PR for this.
Here is an example of the output for `refactor_v1` branch.

```
·------------------------------------------------------------------------------------------|-----------------------------------·
|                                           Gas                                            ·  Block limit: 17592186044415 gas  │
·······················································|···································|····································
|  Methods                                             ·            1 gwei/gas             ·          205.39 usd/eth           │
·····························|·························|···········|···········|···········|··················|·················
|  Contract                  ·  Method                 ·  Min      ·  Max      ·  Avg      ·  # calls         ·  usd (avg)     │
·····························|·························|···········|···········|···········|··················|·················
|  BridgeValidators          ·  addValidator           ·        -  ·        -  ·    53178  ·               1  ·          0.01  │
·····························|·························|···········|···········|···········|··················|·················
|  BridgeValidators          ·  initialize             ·   187738  ·   218775  ·   212568  ·               5  ·          0.04  │
·····························|·························|···········|···········|···········|··················|·················
|  BridgeValidators          ·  removeValidator        ·        -  ·        -  ·    24400  ·               3  ·          0.01  │
·····························|·························|···········|···········|···········|··················|·················
|  BridgeValidators          ·  setRequiredSignatures  ·        -  ·        -  ·    29790  ·               9  ·          0.01  │
·····························|·························|···········|···········|···········|··················|·················
[...]

PR #99, opened 2018-10-18
Add eth-gas-reporter for unit tests

patitonar:
Add [eth-gas-reporter](https://github.com/cgewecke/eth-gas-reporter) for gas usage reports on unit tests.

Related to #47


patitonar:
@akolotov unit test were using 3 validators and 2 required signatures for `submitSignature`, `executeAffirmation` and `executeSignatures`. I added new tests with 5 validators and 3 required signatures. Also I added gas usage report on README file under `test` section. Other alternatives are to move the report to a separate file y linked it from README file or put the report on its own section.

README.md
```diff
@@ -62,6 +62,127 @@ Please the [README.md](deploy/README.md) in the `deploy` folder for instructions
 npm test
 ```
 
+##### Gas usage report
```
akolotov: This information is hard to analyse especially if you are not familiar with the bridge contracts code execution flow. Please take a look at the review comment to consider more practical example of the report that it make sense to add to the README.

patitonar:
@akolotov I added a summarized gas consumption section for `Native-to-ERC` mode, if you think it is OK I can move forward with the other modes. Also since it is a lot of information, I think it will be best to move this section to a different file and add a link to it on the `README`

akolotov:
>  I think it will be best to move this section to a different file and add a link to it on the README
Right. Let's create a separate file. 
I am OK with `Native-to-ERC` section. Good job!

Issue #111, opened 2018-11-23
Changes in security roles 


akolotov:
The current security model assumes existence of three roles:
- an account with permissions to upgrade the bridge contract
- an account with permissions to upgrade the bridge validator contract
- an account with permissions to change bridge parameters and set of validators

After internal discussion it was decided to improve the model.
It still will based on three roles but they will have different meaning:
- "Upgradability admin": an account with permissions to upgrade the bridge contract and the bridge validator contract
- "Bridge admin": an account with permissions to change bridge parameters
- "Validators admin": an account with permissions to change set of validators and number of required signatures

Since the role "Upgradability admin" is most permissive, the method `claimTokens` must be allowed to be called only by an account with this role.

Since these changes related to security they needs to be done for both 1.x and 2.x versions of contracts.

The deployed contracts for the POA-POA20 bridge must be upgraded to apply this security model.

The deployed contracts for the DAI-xDAI bridge must be upgraded to apply this security model as well.

akolotov:
Addressed
  * in #116 for the POA bridge
  * in #117 for current master branch

Issue #110, opened 2018-11-22
ERC-TO-NATIVE: extend limits functionality


akolotov:
The current implementation of the bridge contracts in `erc-to-native` mode allows to an user to operates within specified limits (daily limit and limits per transaction). But it still allows to operates bridge validators without any limitation. So, if a set of validators will be compromised it could lead to a situation when funds locked on the bridge can be released by sending signatures of compromised validators.

In order to reduce the risk of the attack it is recommended to extend the functionality of limits.

So, we will have the following parameters on both bridges side:
**for Home:**
  HOME_DAILY_LIMIT - maximum number of coins that can be burnt daily
  HOME_MAX_AMOUNT_PER_TX - maximum number of coins that can be burnt in one transaction
  HOME_MIN_AMOUNT_PER_TX - minimum number of coins that can be burnt in one transaction
  FOREIGN_DAILY_LIMIT - maximum number of coins that can be **minted** daily
  FOREIGN_MAX_AMOUNT_PER_TX - maximum number of coins that can be **minted** in one transaction

**for Foreign:**
  FOREIGN_MAX_AMOUNT_PER_TX - maximum number of tokens that can be **locked** in one transaction - to be used by the Bridge UI (or a contract in the Foreign network) to inform users that the value of the bridge transfer is above the limit. 
  HOME_DAILY_LIMIT - maximum number of tokens that can be **unlocked** daily
  HOME_MAX_AMOUNT_PER_TX - maximum number of tokens that can be **unlocked** in one transaction

The functionality for `HOME_DAILY_LIMIT`, `HOME_MAX_AMOUNT_PER_TX` and `HOME_MIN_AMOUNT_PER_TX` already exists in the HomeBridge contract
Implementation of the function `executeAffirmation` needs to be changed as so it does not accept incoming requests if they are above `FOREIGN_DAILY_LIMIT` and `FOREIGN_MAX_AMOUNT_PER_TX` limits. Instead of this the value of the assets must be stored in a variable. This variable will be used by the Bridge Monitor to calculate the balance difference properly. An event (e.g. `DailyLimitExceeded` or `MaxAmountLimitExceeded`) must be raised to notify that the operation was not successful. 

So far there is no functionality covering limits in the ForeignBridge contract.
The following function needs to be changed:
[...]

PR #109, opened 2018-11-22
Extend Native-ERC20 limits functionality on contracts v1.1

patitonar:
Relates to issue #108

Gist for deployed HomeBridge: https://gist.github.com/patitonar/df585134034f5b6fbe01928fc6e00e36
Gist for deployed ForeignBridge: https://gist.github.com/patitonar/0ff940d070ef5c4d3dbd70e44eeae361

PR #116, opened 2018-12-04
Update security roles on v1 contracts

patitonar:
Relates to #111 

Applied the new security roles, updated deploy script and tests.

Added `upgradeSecurityRoles` method to be called using `upgradeToAndCall` for already deployed bridge contracts that has not an owner set on initialization.


akolotov:
May you prepare new gists based on the contracts you prepared in #109?

contracts/upgradeable_contracts/Validatable.sol
```diff
@@ -0,0 +1,24 @@
...
+    function validatorContract() public view returns(IBridgeValidators) {
+        return IBridgeValidators(addressStorage[keccak256("validatorContract")]);
+    }
+
+    modifier onlyValidator() {
+        require(validatorContract().isValidator(msg.sender));
+        _;
+    }
+
+    function upgradeSecurityRoles(address _newOwner) public {
```
akolotov: It is not safe to do since anyone who run this method after the contract upgrade first could set up the owner and capture control on the bridge contract. Consider to add the modifier `onlyProxyOwner`.

patitonar: Thanks. Fixed

patitonar:
I will prepare gist for the contracts. Should I also test upgrading them similar as we did on #109 ?

contracts/upgradeable_contracts/Validatable.sol
akolotov: Should this method be included in git at all? It is needed just for upgrade in case of already deployed contracts (MakerDAO-xDAI).

akolotov:
> I will prepare gist for the contracts. Should I also test upgrading them similar as we did on #109 ?

yes. And it should be extended with tests that we can send back tokens that exceed the limit.

contracts/upgradeable_contracts/Validatable.sol
patitonar: The intention was to provide a way to upgrade already deployed contracts. It makes sense to not include it in git. I'll remove it and add it only on gist files

patitonar: Done

patitonar:
Here are the gists:

- Home: https://gist.github.com/patitonar/8caa3631e5a89b85ff3269a27c429813
- Foreign: https://gist.github.com/patitonar/b1356d83fd199d934a6408bdc099ce6f

Next step is to run the suggested tests

patitonar:
> yes. And it should be extended with tests that we can send back tokens that exceed the limit.

Actually on v1.x contracts I don't think we have the ability to send back tokens that exceed the limit. From changes on #109, transactions that exceeded the limit are reverted. However, such kind of test can be applied to `ERC-TO-NATIVE` mode on changes from #110. Is that what you had on mind?

akolotov:
> Is that what you had on mind?

Oh! Sorry. It should be for another issue, for #110. So, for #116 you need to test the following:
- deploy the contract with three different accounts as the admins
- make some two ways transactions
- upgrade the contract, set new owner for the bridge contract
- make some two ways transactions
- reset blocks and restart the bridge to check that the storage was not affected and no double spend happens
- check that the bridge admin works as expected and it cannot perform operations with validators contract, cannot upgrade both the bridge and validator contracts, cannot claim tokens.
- check that the validator admin works as expected and it cannot perform operations with the bridge contract, cannot upgrade both the bridge and validator contracts 
- check that the upgradability admin is able to perform upgrade, can claim tokens.

patitonar:
I run the tests and found one issue.
The issue is related to methods on Bridge contract using modifier `onlyProxyOwner`. 
When I tried to run `upgradeToAndCall` with correct account to upgrade contract and calling `upgradeSecurityRoles` it failed.
I removed the `onlyProxyOwner` modifier and it worked correctly.
```solidity
function upgradeSecurityRoles(address _newOwner) public onlyProxyOwner {
    require(owner() == address(0));
    require(_newOwner != address(0));
    setOwner(_newOwner);
}
```

Then when testing the upgradability admin role, I wasn't able to claim tokens. 
This method uses `onlyProxyOwner` too. 

```solidity
function claimTokens(address _token, address _to) external onlyProxyOwner {
    ...
}
```

I run some tests to understand what was causing the issue with `onlyProxyOwner`. Which is defined on `OwnedUpgradeabilityProxy` as:

```solidity
modifier onlyProxyOwner() {
    require(msg.sender == proxyOwner());
    _;
}
```

I found that methods defined on Bridge contracts that calls `proxyOwner()` the returned value is `0x0000000000000000000000000000000000000000`.
But if you call the method `EternalStorageProxy.methods.proxyOwner()` the returned value is the correct.
I don't know exactly what is the cause of this but it seems some kind of issue of the bridge contract on accessing the state on the proxy contract.

Any ideas or suggestion on how to fix this or any other approach on how can we avoid this situation?

Besides this issue, the new roles worked as expected.

patitonar:
@akolotov Found the issue, it was related to the way in how `Validatable` was defined.
https://github.com/poanetwork/poa-bridge-contracts/blob/f6ba43cd2d3c230826e76d67fafa7a8cfa27c6ef/contracts/upgradeable_contracts/Validatable.sol#L8

Fixing the order of inheritance made the trick
```solidity
contract Validatable is OwnedUpgradeabilityProxy, EternalStorage, Ownable {

}
```
I pushed the fix, and updated the gists files. Also, I was able to re-test the steps that had issues, and now it worked as expected on the steps that used to fail.

akolotov:
Great! Do you you know exactly the reason why the order matters?

patitonar:
Unfortunately, I don't know the exact reason. I just tried having the same order as `EternalStorageProxy` does.
https://github.com/poanetwork/poa-bridge-contracts/blob/1f61c695df23b2a8f79d96e138c6bb5d8cd5b592/contracts/upgradeability/EternalStorageProxy.sol#L13

akolotov:
Honestly, I do not see any reasons why `Validatable` should inherit something besides `EternalStorage`.

patitonar:
We still need `OwnedUpgradeabilityProxy` for methods executed by `onlyProxyOwner` and `Ownable` for setting an owner and for methods that require `onlyOwner`. Do you have any suggestion on how to refactor  this?

akolotov:
OK. I see. They are needed for `upgradeSecurityRoles` only. Since this method is specific for the particular upgrade my suggestion is to move it to `HomeBridge` and `ForeignBridge` implementations. Will it work?

patitonar:
`onlyProxyOwner` is also used by `claimTokens` method which is on `ForeignBridge`. So yes, I think moving `OwnedUpgradeabilityProxy` inheritance to `HomeBridge` and `ForeignBridge` will work. Regarding `Ownable`, `onlyOwner` is used on `BasicBridge`, `HomeBridge` and `ForeignBridge`. So I think moving it to `BasicBridge` is the best option. Is this ok?

patitonar:
Updated the contracts with proposed changes. Also gists files were updated.

akolotov:
> Updated the contracts with proposed changes. Also gists files were updated.

Did you perform tests as described above after the changes?

patitonar:
I updated the changes with solution similar to v2 contracts here https://github.com/poanetwork/poa-bridge-contracts/pull/116/commits/76f7793b3f4d4aeba578de9d74bb088e2ee4f63d 
Also updated gist files.
Home: https://gist.github.com/patitonar/8caa3631e5a89b85ff3269a27c429813
Foreign: https://gist.github.com/patitonar/b1356d83fd199d934a6408bdc099ce6f
 
If solution is OK, next step is to test the upgrade with this solution.

akolotov:
@patitonar I have started reviewing the gist files you prepared and will provide the feedback soon.

patitonar:
@akolotov OK, let me know if there is any change we should perform.

I've tested the upgrade with the proposed changes, and everything worked as expected, no issues were found. State wasn't affected after the upgrades and bridge continue to work as normal.

I started with version v1.0 of contracts, then upgrade them to new version implemented on https://github.com/poanetwork/poa-bridge-contracts/pull/109
and then upgrade them to the version proposed on this PR.

These were the steps for the upgrade of new roles:
    - Call `upgradeTo` from proxy contract
    - Call `upgradeSecurityRoles(newBridgeOwner)` on bridge contract
    - Transfer other ownership to other accounts for testing

akolotov:
@patitonar 

Since we didn't touch `initialize` method in the previous upgrade in both contracts, my suggestion is to not touch it in this upgrade as well. It is also important if someone would like to decode the transaction with `initialize` invocation - the number and meaning of arguments will be the same.

Since we know in advance which owner will be used during the upgrade we could simplify the upgrade procedure if we use in `HomeBridge`
```solidity
function upgradeFrom3To4() public  {
    require(owner() == address(0));
    setOwner(validatorContract().owner());
}
```
and in `ForeignBridge`
```solidity
function upgradeFrom2To3() public  {
    require(owner() == address(0));
    setOwner(validatorContract().owner());
}
```
It will allow to call the methods in `upgrateToAndCall`.

As you see I changed the name of the call in order to have explicit correspondence between the version which is used in `upgradeToAndCall` method and the actions that is going to be made as part of the upgrade.

patitonar:
@akolotov Updated the gist, please check the last changes.

Home: https://gist.github.com/patitonar/8caa3631e5a89b85ff3269a27c429813
Foreign: https://gist.github.com/patitonar/b1356d83fd199d934a6408bdc099ce6f 

Should we run a new test with these updated Gist files?

akolotov:
Yes, please. We cannot use untested changes for upgrade of the bridge in the production.

akolotov:
Thank you! Did you test that the bridge parameters can be changed by the old owner of validator contract after changing the owner of the validator contract to some another account?

patitonar:
Yes. After the upgrade I tested the following steps for that case:
- The new bridge owner can update parameters on the bridge contract
- Validator's owner call `transferOwnership` to a new account
- Bridge owner can still update parameters on the bridge contract
- New Validator's owner  can not update parameters on the bridge contract

akolotov:
Thanks! I will initiate the process to upgrade the contracts of the POA bridge.

akolotov:
The upgraded version of contracts was applied to the ETH mainnet and the POA core:
* [Gist for HomeBridge](https://gist.github.com/patitonar/8caa3631e5a89b85ff3269a27c429813)
  * [Deployed contract in the POA core](https://blockscout.com/poa/core/address/0x2f4f4a52295940c576417d29f22eeb92b440ec89/transactions)
* [Gist for ForeignBridge](https://gist.github.com/patitonar/b1356d83fd199d934a6408bdc099ce6f)
  * [Deployed contract in the ETH Mainnet](https://etherscan.io/address/0x1bb21b6edb83e3db63bde9f51616da6909a368d2)

PR #117, opened 2018-12-11
Update security roles on v2 contracts

### LINKED ISSUE #111: Changes in security roles

patitonar:
Closes #111 

Since we are going to perform one upgrade on deployed contracts of ERC20-To-Native, this PR includes changes from #112.  

To review specific changes from this PR, refer to commit https://github.com/poanetwork/poa-bridge-contracts/commit/abbde316d022bf3e7c143bab6bbccf07bc8b312f

Once we agree on changes, these are the next steps:
- Prepare gists files for upgrade
- Test performing an upgrade of deployed contract
- Update deploy script, `.env.example` and `deploy/README.md` file


contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
```diff
@@ -34,6 +36,7 @@ contract ForeignBridgeNativeToErc is ERC677Receiver, BasicBridge, BasicForeignBr
         uintStorage[keccak256(abi.encodePacked("minPerTx"))] = _minPerTx;
         uintStorage[keccak256(abi.encodePacked("gasPrice"))] = _foreignGasPrice;
         uintStorage[keccak256(abi.encodePacked("requiredBlockConfirmations"))] = _requiredBlockConfirmations;
+        setOwner(_owner);
```
akolotov: Should `claimTokensFromErc677` be also allowed only for upgradability admin?

patitonar: Yes, also two other `claimTokens` overrides needed to be updated too. Fixed them on https://github.com/poanetwork/poa-bridge-contracts/pull/117/commits/adb81cf1ca663b668d4015c512287a0750323ff2

akolotov: Oh! Thanks!
And seams that we found a bug: the Home contact in `erc20-to-erc20` mode does not have 
```solidity
    function claimTokensFromErc677(address _token, address _to) external onlyProxyOwner {
        erc677token().claimTokens(_token, _to);
    }
```
similarly to the Foreign contract in `native-to-erc20` mode.

patitonar: Should we address this bug on a separate issue?

akolotov: Yes, please create an issue for this.

patitonar: Done. Created https://github.com/poanetwork/poa-bridge-contracts/issues/118

patitonar:
Here are the Gist files for each contract. I also described one method for each contract to be called when upgrading the contract to make the necessary updates for limits and new Owner role. Please check the values of the limits and update them as needed, also regarding the Owner, another option is to receive the new owner address as a parameter instead of using current validatorContract's owner.

Once we agree on the details of methods, I'll add them to the gist files.

Home Bridge Gist: https://gist.github.com/patitonar/39ac4d337420864bc912123f72bc41f7
```solidity
// Method for Home Bridge
function upgradeToV220() public onlyProxyOwner {
  require(!boolStorage[keccak256(abi.encodePacked("isUpgradedToV220"))]);
  setOwner(validatorContract().owner());
  uintStorage[keccak256(abi.encodePacked("executionDailyLimit"))] = 1000000 ether;
  emit ExecutionDailyLimitChanged(3000000 ether);
  uintStorage[keccak256(abi.encodePacked("executionMaxPerTx"))] = 100000 ether;
  boolStorage[keccak256("isUpgradedToV220")] = true;
}
```

Foreign Bridge Gist: https://gist.github.com/patitonar/36a6e03c3ca13b912064b7742b5c3f5c
```solidity
// Method for Foreign Bridge
function upgradeToV220() public onlyProxyOwner {
  require(!boolStorage[keccak256(abi.encodePacked("isUpgradedToV220"))]);
  setOwner(validatorContract().owner());
  uintStorage[keccak256(abi.encodePacked("maxPerTx"))] = 100000 ether;
  uintStorage[keccak256(abi.encodePacked("executionDailyLimit"))] = 1000000 ether;
  emit ExecutionDailyLimitChanged(3000000 ether);
  uintStorage[keccak256(abi.encodePacked("executionMaxPerTx"))] = 100000 ether;
  boolStorage[keccak256("isUpgradedToV220")] = true;
}
```

akolotov:
Thanks!
1. 
```
emit ExecutionDailyLimitChanged(3000000 ether);
```
Should be with the value 1000000 ether as well in both methods.

2. 
> another option is to receive the new owner address as a parameter

Let's do in this way.

patitonar:
Fixed the methods and added them into the Gist files. Here are the updated version:

```solidity
// Method for Home Bridge
function upgradeToV220(address _newOwner) public onlyProxyOwner {
  require(!boolStorage[keccak256(abi.encodePacked("isUpgradedToV220"))]);
  setOwner(_newOwner);
  uintStorage[keccak256(abi.encodePacked("executionDailyLimit"))] = 1000000 ether;
  emit ExecutionDailyLimitChanged(1000000 ether);
  uintStorage[keccak256(abi.encodePacked("executionMaxPerTx"))] = 100000 ether;
  boolStorage[keccak256("isUpgradedToV220")] = true;
}
```

```solidity
// Method for Foreign Bridge
function upgradeToV220(address _newOwner) public onlyProxyOwner {
    require(!boolStorage[keccak256(abi.encodePacked("isUpgradedToV220"))]);
    setOwner(_newOwner);
    uintStorage[keccak256(abi.encodePacked("maxPerTx"))] = 100000 ether;
    uintStorage[keccak256(abi.encodePacked("executionDailyLimit"))] = 1000000 ether;
    emit ExecutionDailyLimitChanged(1000000 ether);
    uintStorage[keccak256(abi.encodePacked("executionMaxPerTx"))] = 100000 ether;
    boolStorage[keccak256("isUpgradedToV220")] = true;
}
```

Next step is to test performing an upgrade of the contracts, test that we can send back tokens that exceed the limit and the new security roles.

patitonar:
@akolotov I found an issue while testing. This same issue affects security roles upgrades also v1 contracts.

Previouly I described an issue regarding `onlyProxyOwner` modifier https://github.com/poanetwork/poa-bridge-contracts/pull/116#issuecomment-445297078

The solution implemented to that issue, was to change the order of inheritance between `OwnedUpgradeabilityProxy` and `EternalStorage`.
However, that solution is not correct because it caused the loss of the state of the contract while upgrading the contracts in the last test (because of changing the structure of the store). Also, accessing to the `proxyOwner` value was not correct since it was read from the contract (initialized when deploying new contract version and was the same address) and not from proxy contract.
So I think this change should be reverted.

The `proxyOwner` variable is stored in a separte variable on the proxy contract and not on the `EternalStorage`, so I think is not possible to access that value from Bridge contracts.
Here are some documentation about `EternalStorageProxy` strategy that we are using.
https://blog.zeppelinos.org/proxy-patterns/
https://github.com/zeppelinos/labs/tree/master/upgradeability_using_eternal_storage

Given this, I think a possible solution is to introduce an admin role for Bridge contracts (similar to `Ownable`), that could be set on `initialize` method too.
And also update the method `upgradeToV220` to include setting the address for the admin role.
The bad side of this approach, is that if proxyOwner is changed on the proxy contract, this admin role would have to also be updated manually.

Does this solution sound good? Any other option you can think of?

akolotov:
> Given this, I think a possible solution is to introduce an admin role for Bridge contracts (similar to Ownable), that could be set on initialize method too.
And also update the method upgradeToV220 to include setting the address for the admin role.
The bad side of this approach, is that if proxyOwner is changed on the proxy contract, this admin role would have to also be updated manually.

Honestly, I don't like it, Let me think more about this.

akolotov:
@patitonar I have an idea. I did not tried this but what address will be assigned to `this` if you try to access this variable from a method in the bridge contract? If this address is the same as address of the proxy contract (not the bridge contract implementation), you can try to call `address(this).proxyOwner()` or use the `call` functionality explicitly.

Could you try this?

patitonar:
@akolotov Good idea! I tried that and `this` refers to the proxy contract. It seems to work. Please check https://github.com/poanetwork/poa-bridge-contracts/pull/117/commits/ac0e671aeb7648942f0930a53cd386ed3e7ced2c 
If implementation is Ok, I can update gists and test it properly.

contracts/upgradeable_contracts/OwnedUpgradeability.sol
```diff
@@ -0,0 +1,16 @@
...
+import "../IOwnedUpgradeabilityProxy.sol";
+
+
+contract OwnedUpgradeability {
+
+    function upgradeabilityAdmin() public view returns (address) {
+        return IOwnedUpgradeabilityProxy(this).proxyOwner();
+    }
+
+    modifier onlyProxyOwner() {
```
akolotov: I would avoid of using the same name for different implementation of modifiers since it could to issues in the future. May we have the name like 'onlyIfOwnerOfProxy' and add a comment in here why we need two different modifier for the same functionality?

patitonar: Thanks, Fixed

akolotov: I am OK with the changes.

akolotov:
We can proceed with testing before deployment to the real chains.
Please use https://etherscan.io/address/0x710d6eC2b0948dEf1f423Ec77B51b6a55847D2C5#code and https://blockscout.com/poa/dai/address/0x37d5b903a8fbd0d9f44f9f719fc33603189f2ff5/contracts for the baseline of the contract before testing the upgrade.

patitonar:
I performed the test with the new roles and the new limits functionality. Everything worked as expected. Only one issue was found regarding functionality (setExecutionMaxPerTx) and was fixed.
- Bridge admin // HomeBridge.owner()
    - Can update parameter on bridge contract
    - Can not update parameter on validators contract
    - Can not upgrade bridge contract nor validators contract
    - Can not claim tokens
- Validators admin // BridgeValidators.owner()
    - Can update parameter on validators contracts
    - Can not update parameter on bridge contract
    - Can not upgrade bridge contract nor validators contract
- Upgradability admin // HomeBridgeEternalStorageProxy.proxyOwner()
    - Can upgrade bridge contract
    - Can claim tokens
    - Can fix assets above limits
- Upgradability admin // BridgeValidatorsEternalStorageProxy.proxyOwner()
    - Can upgrade validators contract

- State wasn't affected on contract upgrade.
- About new limits functionality:
    - Sent a Tx above daily execution limit
    - Called fixAssetsAboveLimits(txHash, true) from Upgradability admin, which emitted `UserRequestForSignature`
    - out of limit amount was reduced
    - couldn't execute fixAssetsAboveLimits for the same txHash
    - `UserRequestForSignature` was processed by Bridge

For the contract upgrade, I had to call `upgradeTo` and then call method `upgradeToV220`. I wasn't able to call `upgradeToAndCall`.

The reason is that `upgradeToV220` can be executed only by `proxyOwner`. On `upgradeToAndCall` when executing the method `upgradeToV220` the msg.sender is the address of the proxy contract and not the upgradability admin address.

https://github.com/poanetwork/poa-bridge-contracts/blob/1f61c695df23b2a8f79d96e138c6bb5d8cd5b592/contracts/upgradeability/OwnedUpgradeabilityProxy.sol#L69-L72


Are the upgrade steps OK or should we refactor `upgradeToV220` method to split logic between setters of new limits (just `public ` method to be called with `upgradeToAndCall`) and logic for setting the new bridge owner (`public onlyIfOwnerOfProxy` method to be called after upgrade) ?

akolotov:
> Only one issue was found regarding functionality (setExecutionMaxPerTx) and was fixed.

Was the gist updated with the fix?

patitonar:
> Was the gist updated with the fix?

Yes, both Gist files were updated. On `Revisions` section you can check the updates.

Home Bridge Gist: https://gist.github.com/patitonar/39ac4d337420864bc912123f72bc41f7
Foreign Bridge Gist: https://gist.github.com/patitonar/36a6e03c3ca13b912064b7742b5c3f5c

akolotov:
> Are the upgrade steps OK or should we refactor upgradeToV220 method to split logic between setters of new limits (just public method to be called with upgradeToAndCall) and logic for setting the new bridge owner (public onlyIfOwnerOfProxy method to be called after upgrade) ?

No, that's OK since the upgrade procedure is not trivial anyway:
1. Upgrade contracts
2. Set limits and owner for the bridge contracts on both sides
3. Set a new owner for the validator contract
4. Set a new upgrade admin for the validator contract

akolotov:
So, let's finish changes for the contracts v1.x and distribute the changes to other modes of the bridge.

patitonar:
@akolotov As you mentioned, correct version of `upgradeToV220` should use `onlyIfOwnerOfProxy`.

Home:
```solidity
    function upgradeToV220(address _newOwner) public onlyIfOwnerOfProxy {
        require(!boolStorage[keccak256(abi.encodePacked("isUpgradedToV220"))]);
        setOwner(_newOwner);
        uintStorage[keccak256(abi.encodePacked("executionDailyLimit"))] = 1000000 ether;
        emit ExecutionDailyLimitChanged(1000000 ether);
        uintStorage[keccak256(abi.encodePacked("executionMaxPerTx"))] = 100000 ether;
        boolStorage[keccak256("isUpgradedToV220")] = true;
    }
```

Foreign:
```solidity
    function upgradeToV220(address _newOwner) public onlyIfOwnerOfProxy {
        require(!boolStorage[keccak256(abi.encodePacked("isUpgradedToV220"))]);
        setOwner(_newOwner);
        uintStorage[keccak256(abi.encodePacked("maxPerTx"))] = 100000 ether;
        uintStorage[keccak256(abi.encodePacked("executionDailyLimit"))] = 1000000 ether;
        emit ExecutionDailyLimitChanged(1000000 ether);
        uintStorage[keccak256(abi.encodePacked("executionMaxPerTx"))] = 100000 ether;
        boolStorage[keccak256("isUpgradedToV220")] = true;
    }
```

These correct versions of `upgradeToV220` were used while I performed the tests. 

What I think happened is that, after fixing the issue that was found during the test (on method `setExecutionMaxPerTx`) when updating the file that was going to be used to update the Gist file, I may have accidentally overwritten `upgradeToV220` with the older version of the method, that used `onlyProxyOwner`.

To avoid this situation next time I think we should:
- Push changes to Gist file first, before running the test.
- Review carefully the updates performed to Gist files.
- If any fix was made after a test, update the Gist file, and run a new test with it.

I have fixed the methods `upgradeToV220` on the Gist Files. Let me know which should be the next steps here.

akolotov:
Ok, thanks! We already deployed the contracts to the ETH mainnet and the xDai chain.

patitonar:
@akolotov Applied changes to other modes, updated deploy scripts,  `.env.example` and README file

akolotov:
The upgraded version of contracts was applied to the ETH mainnet and the xDai chain:
* [Gist for HomeBridgeErcToNative](https://gist.github.com/patitonar/39ac4d337420864bc912123f72bc41f7)
  * [Deployed contract in the xDai chain](https://blockscout.com/poa/dai/address/0xcc74f1ffafb9e8cb55076c21be9af6cb0ccfc6d7/transactions)
* [Gist for ForeignBridgeErcToNative](https://gist.github.com/patitonar/36a6e03c3ca13b912064b7742b5c3f5c)
  * [Deployed contract in the ETH Mainnet](https://etherscan.io/address/0x0d3726e5a9f37234d6b55216fc971d30f150a60f)

PR #129, opened 2018-12-26
Extend ERC677BridgeToken contract for `xDai-POA` bridge

varasev:
These changes add `mintReward`, `stake` and `withdraw` functions which are used by [DPoS contracts](https://github.com/poanetwork/pos-contracts/tree/master/contracts) ([`BlockReward`](https://github.com/poanetwork/pos-contracts/blob/13fd4a1c0bb40f97afcef5f2136b851d955dbb88/contracts/BlockRewardAuRa.sol#L44-L56) and [`ValidatorSet`](https://github.com/poanetwork/pos-contracts/blob/13fd4a1c0bb40f97afcef5f2136b851d955dbb88/contracts/abstracts/ValidatorSetBase.sol#L128-L144)).

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -14,6 +14,8 @@ contract ERC677BridgeToken is
     MintableToken {
 
     address public bridgeContract;
+    address public blockRewardContract;
+    address public validatorSetContract;
 
     event ContractFallbackCallFailed(address from, address to, uint value);
 
@@ -28,8 +30,28 @@ contract ERC677BridgeToken is
         bridgeContract = _bridgeContract;
     }
 
+    function setBlockRewardContract(address _blockRewardContract) onlyOwner public {
+        require(_blockRewardContract != address(0) && isContract(_blockRewardContract));
+        blockRewardContract = _blockRewardContract;
+    }
+
+    function setValidatorSetContract(address _validatorSetContract) onlyOwner public {
+        require(_validatorSetContract != address(0) && isContract(_validatorSetContract));
+        validatorSetContract = _validatorSetContract;
+    }
+
     modifier validRecipient(address _recipient) {
         require(_recipient != address(0) && _recipient != address(this));
+        _;
+    }
+
+    modifier onlyBlockRewardContract() {
+        require(msg.sender == blockRewardContract);
+        _;
+    }
+
+    modifier onlyValidatorSetContract() {
+        require(msg.sender == validatorSetContract);
         _;
     }
 
@@ -104,5 +126,33 @@ contract ERC677BridgeToken is
         require(token.transfer(_to, balance));
     }
 
+    function mintReward(address[] _receivers, uint256[] _rewards) external onlyBlockRewardContract {
+        for (uint256 i = 0; i < _receivers.length; i++) {
+            address to = _receivers[i];
+            uint256 amount = _rewards[i];
+
+            // Mint `amount` for `to`
+            totalSupply_ = totalSupply_.add(amount);
+            balances[to] = balances[to].add(amount);
+            emit Mint(to, amount);
+            emit Transfer(address(0), to, amount);
+        }
+    }
+
+    function stake(address _staker, uint256 _amount) external onlyValidatorSetContract {
+        // Transfer `_amount` from `_staker` to `validatorSetContract`
+        require(_amount <= balances[_staker]);
+        balances[_staker] = balances[_staker].sub(_amount);
+        balances[validatorSetContract] = balances[validatorSetContract].add(_amount);
+        emit Transfer(_staker, validatorSetContract, _amount);
+    }
+
+    function withdraw(address _staker, uint256 _amount) external onlyValidatorSetContract {
+        // Transfer `_amount` from `validatorSetContract` to `_staker`
+        require(_amount <= balances[validatorSetContract]);
+        balances[validatorSetContract] = balances[validatorSetContract].sub(_amount);
+        balances[_staker] = balances[_staker].add(_amount);
+        emit Transfer(validatorSetContract, _staker, _amount);
+    }
 
 }


varasev:
@akolotov Ok, done: https://github.com/poanetwork/poa-bridge-contracts/pull/129/commits/60644ade3f378d00e12b8167ca6ee5c7e6f685c4

contracts/ERC677BridgeTokenRewardable.sol
```diff
@@ -0,0 +1,66 @@
...
+        _;
+    }
+
+    function mintReward(address[] _receivers, uint256[] _rewards) external onlyBlockRewardContract {
+        for (uint256 i = 0; i < _receivers.length; i++) {
+            address to = _receivers[i];
+            uint256 amount = _rewards[i];
+
+            // Mint `amount` for `to`
+            totalSupply_ = totalSupply_.add(amount);
```
akolotov: may we use `super.mint(to, amount)` here?

contracts/ERC677BridgeTokenRewardable.sol
```diff
@@ -0,0 +1,66 @@
...
+            balances[to] = balances[to].add(amount);
+            emit Mint(to, amount);
+            emit Transfer(address(0), to, amount);
+        }
+    }
+
+    function stake(address _staker, uint256 _amount) external onlyValidatorSetContract {
+        // Transfer `_amount` from `_staker` to `validatorSetContract`
+        require(_amount <= balances[_staker]);
+        balances[_staker] = balances[_staker].sub(_amount);
```
akolotov: may we use `super.transferFrom(_staker, validatorSetContract, _amount)` here?

contracts/ERC677BridgeTokenRewardable.sol
```diff
@@ -0,0 +1,66 @@
...
+        balances[validatorSetContract] = balances[validatorSetContract].add(_amount);
+        emit Transfer(_staker, validatorSetContract, _amount);
+    }
+
+    function withdraw(address _staker, uint256 _amount) external onlyValidatorSetContract {
+        // Transfer `_amount` from `validatorSetContract` to `_staker`
+        require(_amount <= balances[validatorSetContract]);
+        balances[validatorSetContract] = balances[validatorSetContract].sub(_amount);
```
akolotov: may we use `super.transferFrom(_staker, validatorSetContract, _amount)` here?

contracts/ERC677BridgeTokenRewardable.sol
varasev: No, we can't use this with the current version of `openzeppelin-solidity` package because `super.mint` has the next modifiers:

image

(see `flats/ERC677BridgeToken_flat.sol`)

However, we could use [`_mint` internal function](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/7361ffd26b6354499d0b68d6d2d9da1f8cb21c8b/contracts/token/ERC20/ERC20.sol#L151) which is available in the last version of `openzeppelin-solidity`, but npm package update is required in this case.

contracts/ERC677BridgeTokenRewardable.sol
varasev: We can't use `transferFrom` because it requires allowance and uses `msg.sender`:

image

We could use [`_transfer` internal function](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/7361ffd26b6354499d0b68d6d2d9da1f8cb21c8b/contracts/token/ERC20/ERC20.sol#L136) which is available in the last version of `openzeppelin-solidity` (npm package update is required for that).

contracts/ERC677BridgeTokenRewardable.sol
varasev: The same as above :point_up_2:

akolotov:
Thanks for quick response.
I think it makes sense to modify deployment script for the `native-to-erc` mode to support a new parameter `DEPLOY_REWARDABLE_TOKEN` that could have value `true` or `false`. If it is `true`, `ERC677BridgeTokenRewardable` will be deployed, if it equals `false` (default value), the script must deal with `ERC677BridgeToken`.

Another moment - should we extend tests to cover new functionality?

varasev:
> I think it makes sense to modify deployment script for the `native-to-erc` mode to support a new parameter `DEPLOY_REWARDABLE_TOKEN` ...

Done: https://github.com/poanetwork/poa-bridge-contracts/pull/129/commits/bc289bd2dbc074e8106a303043c96d1f015e44f1

> Another moment - should we extend tests to cover new functionality?

Yes, it makes sense. We need to extend `test/poa20_test.js`, am I right?

akolotov:
> Yes, it makes sense. We need to extend test/poa20_test.js, am I right?

Yes, it is point to start. Then we need to look at scenarios where the new functionality is being used,

deploy/README.md
```diff
@@ -125,6 +125,16 @@ REQUIRED_NUMBER_OF_VALIDATORS=1
 # the Foreign network to confirm that the finalized agreement was transferred
 # correctly to the Foreign network.
 VALIDATORS="0x 0x 0x"
+
+# The flag defining whether to use ERC677BridgeTokenRewardable contract instead of
+# ERC677BridgeToken.
+DEPLOY_REWARDABLE_TOKEN=false
+# The address of ValidatorSet contract used by ERC677BridgeTokenRewardable contract.
+# Makes sense only when DEPLOY_REWARDABLE_TOKEN=true
+VALIDATOR_SET_ADDRESS=0x
```
akolotov: hmm... wait a minute: is this bridge validators or network validators? If they are just for network let's rename the parameter to `DPOS_VALIDATOR_SET_ADDRESS`.

deploy/src/native_to_erc/foreign.js
```diff
@@ -25,7 +26,10 @@ const {
   FOREIGN_REQUIRED_BLOCK_CONFIRMATIONS,
   BRIDGEABLE_TOKEN_NAME,
   BRIDGEABLE_TOKEN_SYMBOL,
-  BRIDGEABLE_TOKEN_DECIMALS
+  BRIDGEABLE_TOKEN_DECIMALS,
+  DEPLOY_REWARDABLE_TOKEN,
```
akolotov: we need to modify `src/loadEnv.js` in order to validate all defined parameters there.

deploy/src/native_to_erc/foreign.js
```diff
@@ -195,6 +199,36 @@ async function deployForeign() {
   assert.equal(Web3Utils.hexToNumber(setBridgeContract.status), 1, 'Transaction Failed')
   foreignNonce++
 
+  if (DEPLOY_REWARDABLE_TOKEN) {
+    console.log('\nset BlockReward contract on ERC677BridgeTokenRewardable')
```
akolotov: A quick question...
Will we consider the rewardable token deployed correctly if it has neither the block reward contract nor the validator set contract defined? If it is critical does it make sense to modify the token contract as so to block all operation until the contracts are defined (or initialize all needed contracts addresses in the token constructor)?

deploy/README.md
varasev: This is the address of [`ValidatorSet` contract](https://github.com/poanetwork/pos-contracts/blob/master/contracts/ValidatorSetAuRa.sol) which contains network validators. Ok, I've renamed the param: https://github.com/poanetwork/poa-bridge-contracts/pull/129/commits/47f766b6c90759281abffee38e0c0834417ed727

deploy/src/native_to_erc/foreign.js
varasev: Done: This is the address of [`ValidatorSet` contract](https://github.com/poanetwork/pos-contracts/blob/master/contracts/ValidatorSetAuRa.sol) which contains network validators. Ok, I've renamed the param: https://github.com/poanetwork/poa-bridge-contracts/pull/129/commits/47f766b6c90759281abffee38e0c0834417ed727.

deploy/src/native_to_erc/foreign.js
varasev: No, it doesn't make sense because `ERC677BridgeTokenRewardable` contract works just like `ERC677BridgeToken` if `setBlockRewardContract` and/or `setValidatorSetContract` functions are nevel called.

Nothing will be broken if we deploy `ERC677BridgeTokenRewardable`, use it for some time and then call `setBlockRewardContract` and/or `setValidatorSetContract` and continue using.

varasev:
> Yes, it is point to start. Then we need to look at scenarios where the new functionality is being used,

OK, I'll add the tests a bit later.

varasev:
The poa20 tests have been extended: https://github.com/poanetwork/poa-bridge-contracts/pull/129/commits/45a094f5dd48a81186315ffd857394ec603f8cf5

deploy/src/native_to_erc/foreign.js
akolotov: thanks for clarification

varasev:
The conflicts have been resolved.

akolotov:
@varasev sorry. just noticed. please change the target branch to `develop`.

varasev:
> please change the target branch to develop.

Done.

PR #130, opened 2018-12-26
Update contracts to properly traverse validators reward accounts tests

### LINKED ISSUE #123: Create utility to properly traverse validators/reward accounts + tests
To properly handle validators and reward addresses, to distribute `fee` among them, a double linked list will be helpful.

See https://github.com/poanetwork/poa-bridge-contracts/issues/119#issuecomment-448246916 for further details.

fernandomg:
Closes #123

Validators now are stored as a circular linked-list, whose sentinel node is the address `0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF`. It's stored under `"validatorsList"` in the EternalStorage's `addressStorage`.

Reward addresses are stored under `"validatorsRewards"` map, with the validator address as key.

`BridgeValidators` contract `initialize` method signature was modified, as it now requires `_initialRewards` addresses.

```solidity
 function initialize(
    uint256 _requiredSignatures,
    address[] _initialValidators,
    address[] _initialRewards,
    address _owner
 ) public returns (bool);
```

---
**Methods added**
```solidity
function getValidatorRewardAddress(address _validator) public view returns (address);
function getNextValidator(address _address) internal view returns (address);
function setValidatorRewardAddress(address _validator, address _reward) internal;
function deleteItemFromAddressStorage(string _mapName, address _address) private;
```

### DIFF
--- a/contracts/IBridgeValidators.sol
+++ b/contracts/IBridgeValidators.sol
@@ -5,4 +5,6 @@ interface IBridgeValidators {
     function isValidator(address _validator) public view returns(bool);
     function requiredSignatures() public view returns(uint256);
     function owner() public view returns(address);
+    function getNextValidator(address _validator) public view returns(address);
+    function getValidatorRewardAddress(address _validator) public view returns(address);
 }
--- a/contracts/IBridgeValidators.sol
+++ b/contracts/IBridgeValidators.sol
@@ -5,5 +5,4 @@ interface IBridgeValidators {
     function isValidator(address _validator) public view returns(bool);
     function requiredSignatures() public view returns(uint256);
     function owner() public view returns(address);
-    function validatorList() public view returns (address[]);
 }
--- a/contracts/IBridgeValidators.sol
+++ b/contracts/IBridgeValidators.sol
@@ -5,6 +5,6 @@ interface IBridgeValidators {
     function isValidator(address _validator) public view returns(bool);
     function requiredSignatures() public view returns(uint256);
     function owner() public view returns(address);
-    function getNextValidator(address _validator) public view returns(address);
     function getValidatorRewardAddress(address _validator) public view returns(address);
+    function validatorList() public view returns (address[]);
 }
--- a/contracts/upgradeable_contracts/BaseBridgeValidators.sol
+++ b/contracts/upgradeable_contracts/BaseBridgeValidators.sol
@@ -30,7 +30,7 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
         return (2, 1, 0);
     }
 
-    function _addValidator(address _validator) internal onlyOwner {
+    function _addValidator(address _validator) internal {
         require(_validator != address(0) && _validator != F_ADDR);
         require(!isValidator(_validator));
 
@@ -40,7 +40,7 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
         setValidatorCount(validatorCount().add(1));
     }
 
-    function _removeValidator(address _validator) internal onlyOwner {
+    function _removeValidator(address _validator) internal {
         require(validatorCount() > requiredSignatures());
         require(isValidator(_validator));
         address validatorsNext = getNextValidator(_validator);
--- a/contracts/upgradeable_contracts/BaseBridgeValidators.sol
+++ b/contracts/upgradeable_contracts/BaseBridgeValidators.sol
@@ -30,6 +30,37 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
         return (2, 1, 0);
     }
 
+    function _addValidator(address _validator) internal onlyOwner {
+        require(_validator != address(0) && _validator != F_ADDR);
+        require(!isValidator(_validator));
+
+        address firstValidator = getNextValidator(F_ADDR);
+        setValidator(_validator, firstValidator);
+        setValidator(F_ADDR, _validator);
+        setValidatorCount(validatorCount().add(1));
+    }
+
+    function _removeValidator(address _validator) internal onlyOwner {
+        require(validatorCount() > requiredSignatures());
+        require(isValidator(_validator));
+        address validatorsNext = getNextValidator(_validator);
+        address index = F_ADDR;
+        address next = getNextValidator(index);
+
+        while (next != _validator) {
+            index = next;
+            next = getNextValidator(index);
+
+            if (next == F_ADDR) {
+                revert();
+            }
+        }
+
+        setValidator(index, validatorsNext);
+        deleteItemFromAddressStorage("validatorsList", _validator);
+        setValidatorCount(validatorCount().sub(1));
+    }
+
     function requiredSignatures() public view returns (uint256) {
         return uintStorage[keccak256(abi.encodePacked("requiredSignatures"))];
     }
--- a/contracts/upgradeable_contracts/BaseBridgeValidators.sol
+++ b/contracts/upgradeable_contracts/BaseBridgeValidators.sol
@@ -1,0 +1,86 @@
+pragma solidity 0.4.24;
+
+import "./Ownable.sol";
+import "../libraries/SafeMath.sol";
+import "../upgradeability/EternalStorage.sol";
+
+
+contract BaseBridgeValidators is EternalStorage, Ownable {
+    using SafeMath for uint256;
+
+    address public constant F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+
+    event RequiredSignaturesChanged (uint256 requiredSignatures);
+
+    function setRequiredSignatures(uint256 _requiredSignatures)
+    external
+    onlyOwner
+    {
+        require(validatorCount() >= _requiredSignatures);
+        require(_requiredSignatures != 0);
+        uintStorage[keccak256(abi.encodePacked("requiredSignatures"))] = _requiredSignatures;
+        emit RequiredSignaturesChanged(_requiredSignatures);
+    }
+
+    function getBridgeValidatorsInterfacesVersion()
+    public
+    pure
+    returns (uint64 major, uint64 minor, uint64 patch)
+    {
+        return (2, 1, 0);
+    }
+
+    function requiredSignatures() public view returns (uint256) {
+        return uintStorage[keccak256(abi.encodePacked("requiredSignatures"))];
+    }
+
+    function validatorCount() public view returns (uint256) {
+        return uintStorage[keccak256(abi.encodePacked("validatorCount"))];
+    }
+
+    function isValidator(address _validator) public view returns (bool) {
+        return _validator != F_ADDR && getNextValidator(_validator) != address(0);
+    }
+
+    function isInitialized() public view returns (bool) {
+        return boolStorage[keccak256(abi.encodePacked("isInitialized"))];
+    }
+
+    function deployedAtBlock() public view returns (uint256) {
+        return uintStorage[keccak256("deployedAtBlock")];
+    }
+
+    function getNextValidator(address _address) public view returns (address) {
+        return addressStorage[keccak256(abi.encodePacked("validatorsList", _address))];
+    }
+
+    function validatorList() public view returns (address[]) {
+        address [] memory list = new address[](validatorCount());
+        uint256 counter = 0;
+        address nextValidator = getNextValidator(F_ADDR);
+
+        while (nextValidator != F_ADDR) {
+            list[counter] = nextValidator;
+            nextValidator = getNextValidator(nextValidator);
+            counter++;
+        }
+
+        return list;
+    }
+
+    function deleteItemFromAddressStorage(string _mapName, address _address) internal {
+        delete addressStorage[keccak256(abi.encodePacked(_mapName, _address))];
+    }
+
+    function setValidatorCount(uint256 _validatorCount) internal {
+        uintStorage[keccak256(abi.encodePacked("validatorCount"))] = _validatorCount;
+    }
+
+    function setValidator(address _prevValidator, address _validator) internal {
+        addressStorage[keccak256(abi.encodePacked("validatorsList", _prevValidator))] = _validator;
+    }
+
+    function setInitialize(bool _status) internal {
+        boolStorage[keccak256(abi.encodePacked("isInitialized"))] = _status;
+    }
+}
--- a/contracts/upgradeable_contracts/BaseBridgeValidators.sol
+++ b/contracts/upgradeable_contracts/BaseBridgeValidators.sol
@@ -35,8 +35,8 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
         require(!isValidator(_validator));
 
         address firstValidator = getNextValidator(F_ADDR);
-        setValidator(_validator, firstValidator);
-        setValidator(F_ADDR, _validator);
+        setNextValidator(_validator, firstValidator);
+        setNextValidator(F_ADDR, _validator);
         setValidatorCount(validatorCount().add(1));
     }
 
@@ -56,7 +56,7 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
             }
         }
 
-        setValidator(index, validatorsNext);
+        setNextValidator(index, validatorsNext);
         deleteItemFromAddressStorage("validatorsList", _validator);
         setValidatorCount(validatorCount().sub(1));
     }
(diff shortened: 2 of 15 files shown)


fernandomg:
@akolotov, @varasev 

After a code review with @patitonar, we've found that I need to make the following changes:

- validate reward addresses
- `addValidator` is missing the `_reward` address parameter
- +tests

I'll be adding those ASAP

patitonar:
@akolotov I'm not planning more commits under this PR. You can start with the review. Thanks

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
...
         emit RequiredSignaturesChanged(_requiredSignatures);
     }
 
-    function getBridgeValidatorsInterfacesVersion() public pure returns(uint64 major, uint64 minor, uint64 patch) {
+    function getBridgeValidatorsInterfacesVersion()
+        public
+        pure
+        returns (uint64 major, uint64 minor, uint64 patch)
+    {
         return (2, 0, 0);
```
akolotov: We need to increase the version of the contract.

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
...
+        address next = getNextValidator(index);
+
+        // find the element in the list pointing to _validator
+        while (next != _validator) {
+            index = next;
+            next = getNextValidator(index);
+        }
+
+        setValidator(index, validatorsNext);
+        deleteItemFromAddressStorage("validatorsList", _validator);
```
akolotov: could you confirm that after removing the mapping element is zeroed? Do we have the corresponding unit test for this?

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
...
         return (2, 0, 0);
     }
 
-    function requiredSignatures() public view returns(uint256) {
+    function requiredSignatures() public view returns (uint256) {
         return uintStorage[keccak256(abi.encodePacked("requiredSignatures"))];
     }
 
-    function validatorCount() public view returns(uint256) {
+    function validatorCount() public view returns (uint256) {
```
akolotov: In which scenarios is it being used? Could we remove `validatorCount` at all and re-work the corresponding pieces of code to use `getNextValidator` instead similar to the `while`-loop in the `validatorList` method.

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
...
-    function deployedAtBlock() public view returns(uint256) {
+    function deployedAtBlock() public view returns (uint256) {
         return uintStorage[keccak256("deployedAtBlock")];
     }
 
+    function getValidatorRewardAddress(address _validator) public view returns (address) {
+        return addressStorage[keccak256(abi.encodePacked("validatorsRewards", _validator))];
+    }
+
+    function getNextValidator(address _address) public view returns (address) {
```
akolotov: Should we revert if the corresponding element points to `adress(0)`?

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
...
     function removeValidator(address _validator) external onlyOwner {
         require(validatorCount() > requiredSignatures());
         require(isValidator(_validator));
-        setValidator(_validator, false);
+        address validatorsNext = getNextValidator(_validator);
+        address index = F_ADDR;
+        address next = getNextValidator(index);
+
+        // find the element in the list pointing to _validator
+        while (next != _validator) {
```
akolotov: we need to check that the loop will not eat all the gas (the end of the list is reached)

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
 contract BridgeValidators is IBridgeValidators, EternalStorage, Ownable {
     using SafeMath for uint256;
 
+    address public constant F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+
     event ValidatorAdded (address indexed validator);
```
akolotov: should we extend the event to report reward address as well?

contracts/upgradeable_contracts/BridgeValidators.sol
```diff
@@ -9,89 +9,167 @@ import "../upgradeability/EternalStorage.sol";
...
+        while (nextValidator != F_ADDR) {
+            list[counter] = nextValidator;
+            nextValidator = getNextValidator(nextValidator);
+            counter++;
+        }
+
+        return list;
+    }
+
+    function setValidatorRewardAddress(address _validator, address _reward) internal {
```
akolotov: Should we have a base validator contract without any code related to the reward and another contract like `RewardableValidators` which will inherit functionality of the base validator contract. It will allow to use the base contract in the cases when the subsidized bridge is needed.

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Done on https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/349bbbb355d9ee155eefc53886780a4e6f3f7f4c

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: I added a unit test for this case. https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/349bbbb355d9ee155eefc53886780a4e6f3f7f4c

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Done https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/5c553e06eb323e77c3628549c38f04bf224e07a0

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Yes, that makes sense. Here is the refactor https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/4e9db7b626ef73a847b019247865d9f0ae1d53d4

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Right now there is a use case if element points to `address(0)`, it help us know if the address is a validator.
https://github.com/poanetwork/poa-bridge-contracts/blob/4e9db7b626ef73a847b019247865d9f0ae1d53d4/contracts/upgradeable_contracts/BaseBridgeValidators.sol#L41-L43

I think it's a good idea to let this method just get the value, and call `revert()` on the method that called `getNextValidator` if the scenario of the method requires that. what do you think?

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: I don't think `validatorCount` is being used externally (only on unit tests). Here are the uses inside the contract:

https://github.com/poanetwork/poa-bridge-contracts/blob/5c553e06eb323e77c3628549c38f04bf224e07a0/contracts/upgradeable_contracts/BaseBridgeValidators.sol#L15-L19

https://github.com/poanetwork/poa-bridge-contracts/blob/5c553e06eb323e77c3628549c38f04bf224e07a0/contracts/upgradeable_contracts/BridgeValidators.sol#L65-L66

https://github.com/poanetwork/poa-bridge-contracts/blob/5c553e06eb323e77c3628549c38f04bf224e07a0/contracts/upgradeable_contracts/BaseBridgeValidators.sol#L57-L58

For `setRequiredSignatures` and `removeValidator`, `validatorCount` could be refactored to  use `getNextValidator` with the `while` loop and so avoid storing the value.

The only case that worries me is on `validatorList` method. To create the Array to return the list of validators we need to know the size of the Array. Performing the refactor, will cause that by invoking `validatorList` we will be iterating two times on the linked-list, one to get the length and the other to store the validator addresses on the Array to be returned. Also, `validatorList()` method will be used every time a transaction is processed by `fee contract` to distribute the fees.
Perhaps there is a way to avoid iterating the two times on the linked-list, any thoughts on this?

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: We have the following check: ` require(isValidator(_validator));` to ensure that the validator to be removed is in the list. But if that is not enough then I can add something like this:
```solidity
while (next != _validator) {
    index = next;
    next = getNextValidator(index);

    if (next == F_ADDR) {
        revert();
    }
}
```

contracts/upgradeable_contracts/BridgeValidators.sol
akolotov: Usage a variable in the `validatorList` makes sense since this method will be used intensively. So, let leave `validatorCount` as it is now. Thanks for clarification!!!

contracts/upgradeable_contracts/BridgeValidators.sol
akolotov: OK. Then let's go through all places where `getNextValidator` is invoked and check whether the scenario when it returns `address(0)` is valid or not. If it is not valid, we will `revert`.

contracts/upgradeable_contracts/BridgeValidators.sol
akolotov: yes. I think it makes sense to have this check just to be sure.

contracts/upgradeable_contracts/BaseBridgeValidators.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    function deleteItemFromAddressStorage(string _mapName, address _address) internal {
+        delete addressStorage[keccak256(abi.encodePacked(_mapName, _address))];
+    }
+
+    function setValidatorCount(uint256 _validatorCount) internal {
+        uintStorage[keccak256(abi.encodePacked("validatorCount"))] = _validatorCount;
+    }
+
+    function setValidator(address _prevValidator, address _validator) internal {
```
varasev: Maybe it's better to rename `setValidator` to `setNextValidator`?

contracts/IBridgeValidators.sol
```diff
@@ -5,4 +5,5 @@ interface IBridgeValidators {
     function isValidator(address _validator) public view returns(bool);
     function requiredSignatures() public view returns(uint256);
     function owner() public view returns(address);
+    function validatorList() public view returns (address[]);
```
akolotov: Is it valid only for rewardable validators?

contracts/upgradeable_contracts/BaseBridgeValidators.sol
```diff
@@ -0,0 +1,86 @@
...
+    external
+    onlyOwner
+    {
+        require(validatorCount() >= _requiredSignatures);
+        require(_requiredSignatures != 0);
+        uintStorage[keccak256(abi.encodePacked("requiredSignatures"))] = _requiredSignatures;
+        emit RequiredSignaturesChanged(_requiredSignatures);
+    }
+
+    function getBridgeValidatorsInterfacesVersion()
```
akolotov: does it make sense to have separate versions for the rewardable validator contract and for the generic validator contract?

contracts/upgradeable_contracts/RewardableValidators.sol
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    function addValidator(address _validator, address _reward) external onlyOwner {
+        require(_validator != address(0) && _validator != F_ADDR);
+        require(_reward != address(0));
+        require(!isValidator(_validator));
+
+        address firstValidator = getNextValidator(F_ADDR);
+        setValidator(_validator, firstValidator);
+        setValidatorRewardAddress(_validator, _reward);
```
akolotov: So, the difference in this method between two contracts is actually: 

```solidity
setValidatorRewardAddress(_validator, _reward);
emit ValidatorAdded(_validator, _reward);
```

Could we have the common functionality in the basic contract and invoke it in the child contract?
For example, `BridgeValidators` will have
```solidity
function addValidator(address _validator) external onlyOwner {
    _addValidator(_validator);
    emit ValidatorAdded(_validator);
}
```
and  `RewardableValidators` will have
```solidity
function addValidator(address _validator) external onlyOwner {
    _addValidator(_validator);
    setValidatorRewardAddress(_validator, _reward);
    emit ValidatorAdded(_validator, _reward);
}
```

Similar is for the `removeValidator` functionality.

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Done on https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/1d7456cf778b36653b9569602b15a9ee552bb2c3

contracts/upgradeable_contracts/BaseBridgeValidators.sol
patitonar: Thanks for suggestion! Done on https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/6912fdcad9b8a1c669edd3d0867245984d824e17

contracts/upgradeable_contracts/RewardableValidators.sol
patitonar: Thanks for suggestion! Fixed on https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/b74360256415b3fe2f5aca4a9e0a486655643caf

contracts/upgradeable_contracts/BridgeValidators.sol
patitonar: Done. Please check https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/212d7964122cd26838dfeb74f4dc584bb89796b8

contracts/upgradeable_contracts/BaseBridgeValidators.sol
patitonar: The only downside I see for having separate versions is that if we change something on `BaseBridgeValidators` that will affect both validator contracts so we will have to update both versions. On the bridge contracts we have only one version that is used for all bridge modes that is defined on `BasicBridge`, so I think it's a good idea to follow that approach here too. Does it make sense?

contracts/IBridgeValidators.sol
patitonar: I'm not sure if I understood the question correctly but `validatorList` is defined on `BaseBridgeValidators` so it is valid for rewardable validator and generic validator contracts

akolotov: I meant that `validatorList` is not needed for generic validator contract since there is no scenarios where it can be used.

contracts/upgradeable_contracts/BaseBridgeValidators.sol
akolotov: Ok. If it is required in the future we can address it later.

contracts/upgradeable_contracts/BaseBridgeValidators.sol
```diff
@@ -30,6 +30,37 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
         return (2, 1, 0);
     }
 
+    function _addValidator(address _validator) internal onlyOwner {
```
akolotov: we don't need the modifier `onlyOwner` here since it is already checked by the caller.

contracts/upgradeable_contracts/BaseBridgeValidators.sol
```diff
@@ -51,7 +51,7 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
             index = next;
             next = getNextValidator(index);
 
-            if (next == F_ADDR) {
+            if (next == F_ADDR || next == address(0) ) {
```
akolotov: I would say that we need to check the return value after the lines 37, 48 and 91.

contracts/upgradeable_contracts/BaseBridgeValidators.sol
patitonar: Thanks. Done on https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/a3b947a50721accd4cf7844992ca1b444a6f08c8

contracts/upgradeable_contracts/BaseBridgeValidators.sol
patitonar: Done, added those checks here https://github.com/poanetwork/poa-bridge-contracts/pull/130/commits/5bf54405c98eba14c525e42d7627595f170189fe

PR #131, opened 2019-01-03
Merge the changes related to security updates to the master branch

akolotov:
This set of changes addresses two issues:
  * Meaning of security roles for accounts that could perform management operations over the bridge contracts was changed
  * Security limits are introduced to reduce the impact of the validator's malicious behavior


akolotov:
As soon as we the PR is merged, I will release new version (2.2.0) of the contracts.

PR #132, opened 2019-01-10
Add Fee Manager implementation for ERC20-to-Native mode

### LINKED ISSUE #122: Create FeeManager basic contract and its interface
At least we're going to need:
- `IFeeManager`
- `BaseFeeManager`
- `FeeManagerHelpers` (or a different name, homologous to `Validatable`)

### LINKED ISSUE #125: Implement `FeeManager` contract for `erc20_to_native` mode + tests
The following approaches to assign reward for the bridge validators will be implemented here:

**Foreign to Home transfer**. As soon as the tokens are transferred to the Home side, the validators will get reward in the native coins:
1. User call `transfer(bridgeAddress, X)` on Foreign:
    * `X` tokens are on the account of the bridge contract 
2. Validators executeAffirmation on Home:
    * Calculate `fee`
    * Mint(`X-fee`) native coins for user
    * Mint(`fee/N`) to every validator

**Home to Foreign transfer**. As soon as the native coins are transferred to the Foreign side, the validators will get the reward in the native coins:
1. User transfer `X` native coins to the account of the bridge contract
    * Calculate `fee`
    * Burn(`X-fee`)
    * UserRequestForSignature(`X-fee`)
2. Validator submitSignatures on Home
   * Send(`fee/N`) coins from the bridge contract account for every validator
   * CollectedSignatures event
3. Validator executeSignatures to unlock (`X-fee`) tokens on Foreign

patitonar:
closes #122, closes #125

I'm not planning more changes on implementation of contracts, unless issues are found while adding unit tests, so they can be reviewed.

Pending:
- [x]  Unit tests


contracts/upgradeable_contracts/erc20_to_native/FeeManagerErcToNative.sol
```diff
@@ -0,0 +1,20 @@
...
+        return IBlockReward(addressStorage[keccak256(abi.encodePacked("blockRewardContract"))]);
+    }
+
+    function onAffirmationFeeDistribution(address _rewardAddress, uint256 _fee) internal {
+        IBlockReward blockReward = blockRewardContract();
+        blockReward.addExtraReceiver(_fee, _rewardAddress);
+    }
+
+    function onSignatureFeeDistribution(address _rewardAddress, uint256 _fee) internal {
+        _rewardAddress.send(_fee);
```
patitonar: Should we use `send` or `transfer` in this case?

contracts/upgradeable_contracts/BaseFeeManager.sol
```diff
@@ -0,0 +1,65 @@
...
+import "../libraries/SafeMath.sol";
+import "../IRewardableValidators.sol";
+
+
+contract BaseFeeManager is EternalStorage {
+    using SafeMath for uint256;
+
+    event FeeUpdated(uint256 fee);
+
+    function calculateFee(uint256 _value, bool _recover) external payable returns(uint256) {
```
patitonar: I had to set this method as `payable` otherwise it will revert when trying to call it using `delegatecall` from fallback method on `HomeBridgeErcToNative` https://github.com/poanetwork/poa-bridge-contracts/blob/c7d62b43e1097a2211c808a5e1430c436756813c/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol#L28

contracts/upgradeable_contracts/BasicHomeBridge.sol
```diff
@@ -155,4 +173,41 @@ contract BasicHomeBridge is EternalStorage, Validatable {
 
     function onFailedAffirmation(address, uint256, bytes32) internal {
     }
+
+    function feeManagerContract() public view returns(address) {
```
akolotov: the suggestion is to move all methods related to fee management to a separate contract like `RewardableBridge.sol` and inherit this contract in `HomeBridgeErcToNative`.

contracts/upgradeable_contracts/BasicHomeBridge.sol
```diff
@@ -74,9 +77,24 @@ contract BasicHomeBridge is EternalStorage, Validatable {
         if (signed >= reqSigs) {
             setNumMessagesSigned(hashMsg, markAsProcessed(signed));
             emit CollectedSignatures(msg.sender, hashMsg, reqSigs);
+
+            address feeManager = feeManagerContract();
```
akolotov: consider to add the call of `onSignaturesCollected` here and move these four lines to the implementation of this method that will be located in the `HomeBridgeErcToNative` contract. It will be similar we have for the `onExecuteAffirmation` method.

contracts/upgradeable_contracts/BasicHomeBridge.sol
```diff
@@ -155,4 +173,41 @@ contract BasicHomeBridge is EternalStorage, Validatable {
...
+    function setFeeManagerContract(address _feeManager) public onlyOwner {
+        require(_feeManager == address(0) || isContract(_feeManager));
+        addressStorage[keccak256(abi.encodePacked("feeManagerContract"))] = _feeManager;
+    }
+
+    function setFee(uint256 _fee) external onlyIfOwnerOfProxy {
+        require(feeManagerContract().delegatecall(abi.encodeWithSignature("setFee(uint256)", _fee)));
+    }
+
+    function getFee() public view returns(uint256) {
```
akolotov: it makes sense to call the corresponding method of the fee manager contract here because it is responsible for storing the fee value (imagine if the value should be calculated by somehow)

contracts/upgradeable_contracts/BasicHomeBridge.sol
```diff
@@ -74,9 +77,24 @@ contract BasicHomeBridge is EternalStorage, Validatable {
...
     }
 
+    function handleSignatureFeeDistribution(address feeManager, bytes message) internal {
+        address recipient;
+        uint256 amount;
+        bytes32 txHash;
+        address contractAddress;
+        (recipient, amount, txHash, contractAddress) = Message.parseMessage(message);
+        uint256 fee = calculateFee(amount, true, feeManager);
+        require(feeManager.delegatecall(abi.encodeWithSignature("distributeFeeFromSignatures(uint256)", fee)));
```
akolotov: Consider to have `distributeFeeFromSignatures` and `distributeFeeFromAffirmation` similarly you implemented `calculateFee` below in this contract. It will make the code more readable.

contracts/upgradeable_contracts/erc20_to_native/FeeManagerErcToNative.sol
patitonar: Thanks! Applied same approach here https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/51708637ac1a46d1657f5fd588a1814685c60e89

contracts/upgradeable_contracts/BaseFeeManager.sol
patitonar: Thanks for suggestion! `callcode` worked https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/98dfc8becfaf7aa4fbd2ae2b9440c0032cc3293c

contracts/upgradeable_contracts/BasicHomeBridge.sol
patitonar: Done https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/faf9712db27d7f3a22eda384f2131368791dbfe1

contracts/upgradeable_contracts/BasicHomeBridge.sol
patitonar: Done as part of this commit https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/faf9712db27d7f3a22eda384f2131368791dbfe1

contracts/upgradeable_contracts/BasicHomeBridge.sol
patitonar: Thanks, fixed https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/8876e15fffecea4a5dcb6fc62d20c6057512a6cf

contracts/upgradeable_contracts/BasicHomeBridge.sol
patitonar: Thanks, added `onSignaturesCollected` https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/1993efcf4c6831fd5cafbfe15c8788a426da9d6b

contracts/upgradeable_contracts/BaseFeeManager.sol
```diff
@@ -0,0 +1,65 @@
...
+        }
+        return _value.mul(fee).div(eth.sub(fee));
+    }
+
+    function setFee(uint256 _fee) external {
+        uintStorage[keccak256(abi.encodePacked("fee"))] = _fee;
+        emit FeeUpdated(_fee);
+    }
+
+    function getFee() public view returns(uint256) {
```
akolotov: is it really public? my understanding is that we always will call it by `callcode`

contracts/upgradeable_contracts/RewardableBridge.sol
```diff
@@ -0,0 +1,63 @@
...
+    function setFee(uint256 _fee) external onlyIfOwnerOfProxy {
+        require(feeManagerContract().delegatecall(abi.encodeWithSignature("setFee(uint256)", _fee)));
+    }
+
+    function getFee() public view returns(uint256) {
+        uint256 fee;
+        bytes memory callData = abi.encodeWithSignature("getFee()");
+        address feeManager = feeManagerContract();
+        assembly {
+            let result := delegatecall(gas, feeManager, add(callData, 0x20), mload(callData), 0, 32)
```
akolotov: will it be safer to call `callcode` here?

contracts/upgradeable_contracts/RewardableBridge.sol
```diff
@@ -0,0 +1,63 @@
+pragma solidity 0.4.24;
+
+import "./OwnedUpgradeability.sol";
+import "./Ownable.sol";
+
+
+contract RewardableBridge is Ownable, OwnedUpgradeability {
+
+    function setFee(uint256 _fee) external onlyIfOwnerOfProxy {
```
akolotov: please change the modifier to `onlyOwner` since whether the reward is distributed among the validators is the bridge parameter.

contracts/upgradeable_contracts/BaseFeeManager.sol
```diff
@@ -0,0 +1,65 @@
...
+        return uintStorage[keccak256(abi.encodePacked("fee"))];
+    }
+
+    function distributeFeeFromAffirmation(uint256 _fee) external {
+        distributeFeeProportionally(_fee, true);
```
akolotov: consider to have two constants REWARD_FOR_TRANSFERRING_FROM_HOME and REWARD_FOR_TRANSFERRING_FROM_FOREIGN. They will be used instead of `true` and `false`. In this case `_isAffirmation` bellow could be replaced by `_direction`.
So, the content of `onFeeDistribution` could look like
```solidity
    function onFeeDistribution(address _rewardAddress, uint256 _fee, bool _direction) internal {
        if (_direction == REWARD_FOR_TRANSFERRING_FROM_FOREIGN) {
            onAffirmationFeeDistribution(_rewardAddress, _fee);
        } else {
            onSignatureFeeDistribution(_rewardAddress, _fee);
        }
    }
```

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
```diff
@@ -84,10 +91,31 @@ contract HomeBridgeErcToNative is EternalStorage, BasicBridge, BasicHomeBridge,
         setTotalExecutedPerDay(getCurrentDay(), totalExecutedPerDay(getCurrentDay()).add(_value));
         IBlockReward blockReward = blockRewardContract();
         require(blockReward != address(0));
-        blockReward.addExtraReceiver(_value, _recipient);
+        uint256 valueToMint = _value;
+
+        address feeManager = feeManagerContract();
+        if (feeManager != address(0)) {
+            uint256 fee = calculateFee(valueToMint, false, feeManager);
+            distributeFeeFromAffirmation(fee, feeManager);
```
akolotov: Seems I found a potential bug: what if we have 3 validators and 1 gwei to distribute as a fee. In this case every validator will get 333333333 wei. The sum these values is 999999999 wei which is less than 1 gwei. Does it mean that the rest (1 wei) should be returned back to the user? It is applicable to transfer in both directions.
We will discuss this internally and I provide a suggestion. If you have any ideas please let me know.

akolotov: OK. We discussed this issue and the decision is to add the rest of fee to a validator chosen on pseudo-random manner. So, if a fee is 1 gwei, two validators will get per 333333333 and one will get 333333334.
In order to determine the validator-recipient for this value I suggest to use the block hash:
```solidity
    function random() public view returns(uint256) {
        return uint256(blockhash(block.number-1)) % validatorCount();
    }
```

patitonar: Thanks for suggestions. Fixed here https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/71870e0076fb801699c5c337bf52f221a214c83a

contracts/upgradeable_contracts/BaseFeeManager.sol
patitonar: Done https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/4b8f53a01459a6c31768acad7a8acbe8f079f4ad

contracts/upgradeable_contracts/RewardableBridge.sol
patitonar: Fixed https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/ee666a15f304de8187c7de8e058ce4314bf38932

contracts/upgradeable_contracts/RewardableBridge.sol
patitonar: Updated to use `callcode` https://github.com/poanetwork/poa-bridge-contracts/pull/132/commits/6b9ce95bccb7dbfa584935c9d198248d2a7b00a0

contracts/upgradeable_contracts/BaseFeeManager.sol
patitonar: It is called by `callcode` on `RewardableBridge.getFee` and also called by `BaseFeeManager` on `calculateFee`.
https://github.com/poanetwork/poa-bridge-contracts/blob/d5b9bb5dcae9b793e31f5a1ad16d82a56cdcd63c/contracts/upgradeable_contracts/BaseFeeManager.sol#L13-L14

`public` modifier is the only one that would work for both cases

akolotov: ok. thanks!

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
akolotov: great!

PR #140, opened 2019-01-23
140-erc-to-erc-multiple

bejavu:
Added a new bridge mode: `ERC_TO_ERC_MULTIPLE` that implement a many to many erc20 bridge between two networks.
To do so we created:
* A bridge factory contracts, both on home and foreign (`HomeBridgeFactory` and `ForeignBridgeFactory`) for creating mulitple ERC-to-ERC bridge contracts.
* A `BridgeMapper` contracts on home that map all of the bridge contracts on home and foreign so that one Oracle can process multiple bridges at the same time.
* Deploy scripts for home and foreign network of the new mode.


LiorRabin:
Created `token-bridge` pull requested for the new bridge mode: https://github.com/poanetwork/token-bridge/pull/131

deploy/src/deploymentUtils.js
```diff
@@ -109,6 +109,7 @@ async function sendNodeRequest(url, method, signedData) {
     })
   })
   const json = await request.json()
+  // console.log('json', json)
```
patitonar:
```suggestion
```

deploy/.env.example
```diff
@@ -11,9 +11,16 @@ BRIDGEABLE_TOKEN_SYMBOL=TEST
 BRIDGEABLE_TOKEN_DECIMALS=18
 
 HOME_RPC_URL=https://sokol.poa.network
-HOME_BRIDGE_OWNER=0x
-HOME_VALIDATORS_OWNER=0x
-HOME_UPGRADEABLE_ADMIN=0x
+# HOME_OWNER_MULTISIG=0x
+HOME_BRIDGE_OWNER=0x ##
+HOME_VALIDATORS_OWNER=0x ##
```
patitonar:
```suggestion
HOME_VALIDATORS_OWNER=0x
```

deploy/.env.example
```diff
@@ -11,9 +11,16 @@ BRIDGEABLE_TOKEN_SYMBOL=TEST
...
+HOME_FACTORY_OWNER=0x
+HOME_MAPPER_OWNER=0x
+HOME_UPGRADEABLE_ADMIN=0x ##
```
patitonar:
```suggestion
HOME_UPGRADEABLE_ADMIN=0x
```

deploy/.env.example
```diff
@@ -24,9 +31,14 @@ HOME_GAS_PRICE=1000000000
 BLOCK_REWARD_ADDRESS=
 
 FOREIGN_RPC_URL=https://sokol.poa.network
-FOREIGN_BRIDGE_OWNER=0x
-FOREIGN_VALIDATORS_OWNER=0x
-FOREIGN_UPGRADEABLE_ADMIN=0x
+FOREIGN_BRIDGE_OWNER=0x ##
```
patitonar:
```suggestion
FOREIGN_BRIDGE_OWNER=0x
```

deploy/.env.example
```diff
@@ -24,9 +31,14 @@ HOME_GAS_PRICE=1000000000
...
+FOREIGN_VALIDATORS_OWNER=0x ##
```
patitonar:
```suggestion
FOREIGN_VALIDATORS_OWNER=0x
```

deploy/.env.example
```diff
@@ -24,9 +31,14 @@ HOME_GAS_PRICE=1000000000
...
+# FOREIGN_OWNER_MULTISIG=0x
+FOREIGN_FACTORY_OWNER=0x
+FOREIGN_UPGRADEABLE_ADMIN=0x ##
```
patitonar:
```suggestion
FOREIGN_UPGRADEABLE_ADMIN=0x
```

deploy/.env.example
```diff
@@ -11,9 +11,16 @@ BRIDGEABLE_TOKEN_SYMBOL=TEST
...
-HOME_BRIDGE_OWNER=0x
-HOME_VALIDATORS_OWNER=0x
-HOME_UPGRADEABLE_ADMIN=0x
+# HOME_OWNER_MULTISIG=0x
+HOME_BRIDGE_OWNER=0x ##
+HOME_VALIDATORS_OWNER=0x ##
+HOME_FACTORY_OWNER=0x
+HOME_MAPPER_OWNER=0x
+HOME_UPGRADEABLE_ADMIN=0x ##
+# HOME_UPGRADEABLE_ADMIN_VALIDATORS=0x
```
patitonar: Please correct me if I'm wrong, but I think the following variables are not being used and can be removed:
```
# HOME_OWNER_MULTISIG=0x
# FOREIGN_OWNER_MULTISIG=0x
# HOME_UPGRADEABLE_ADMIN_VALIDATORS=0x
# HOME_UPGRADEABLE_ADMIN_BRIDGE=0x
# HOME_UPGRADEABLE_ADMIN_FACTORY=0x
# HOME_UPGRADEABLE_ADMIN_MAPPER=0x
# FOREIGN_UPGRADEABLE_ADMIN_VALIDATORS=0x
# FOREIGN_UPGRADEABLE_ADMIN_BRIDGE=0x
# FOREIGN_UPGRADEABLE_ADMIN_FACTORY=0x
```

Also I think it would be good if we include the following variables that are used on deploy scripts:
```
HOME_BRIDGE_VALIDATORS_IMPLEMENTATION_ADDRESS=
HOME_BRIDGE_IMPLEMENTATION_ADDRESS=
FOREIGN_BRIDGE_VALIDATORS_IMPLEMENTATION_ADDRESS=
FOREIGN_BRIDGE_IMPLEMENTATION_ADDRESS=
```

contracts/upgradeability/EternalStorage.sol
```diff
@@ -14,4 +14,12 @@ contract EternalStorage {
     mapping(bytes32 => bool) internal boolStorage;
     mapping(bytes32 => int256) internal intStorage;
 
+
+    mapping(bytes32 => uint256[]) internal uintArrayStorage;
```
patitonar: Is there any use case for these new Array mappings? If not, I think we should remove them

contracts/upgradeable_contracts/factories/HomeBridgeFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+
+contract HomeBridgeFactory is EternalStorage, EternalOwnable {
+
+  event HomeBridgeDeployed(address indexed _homeBridge, address indexed _homeValidators, address indexed _token, uint256 _blockNumber);
+
+  function getBridgeFactoryVersion() public pure returns(uint64 major, uint64 minor, uint64 patch) {
+    return (2, 2, 0);
+  }
+
+  function initialize(address _owner,
```
patitonar: We should check that the contract wasn't previously initialized, otherwise someone could override the parameters later.

We can define the following methods similar to the used on the other bridge contracts:
https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/BasicBridge.sol#L75-L77

https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/BasicBridge.sol#L79-L81

And then on `initiliaze` add 
```solidity
require(!isInitialized());
...
...
setInitialize(true);
```

The same should be implemented on `initialize` methods from `ForeignBridgeFactory` and `BridgeMapper`

contracts/upgradeable_contracts/factories/HomeBridgeFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+      address _bridgeValidatorsImplementation,
+      uint256 _requiredSignatures,
+      address[] _initialValidators,
```
patitonar: On `initialize` method of `BridgeValidators`, it checks the following:
```solidity
require(_initialValidators.length >= _requiredSignatures);
```

I think it would be good to have the same check here to avoid any issues when invoking the method `deployHomeBridge`.
This check should be also added on `ForeignBridgeFactory`

contracts/upgradeable_contracts/factories/HomeBridgeFactory.sol
```diff
@@ -0,0 +1,216 @@
...
+    require(_homeBridgeErcToErcImplementation != address(0));
+    require(_gasPrice > 0);
+    require(_requiredBlockConfirmations > 0);
+    require(_minPerTx > 0 && _maxPerTx > _minPerTx && _dailyLimit > _maxPerTx);
+    require(_foreignMaxPerTx < _foreignDailyLimit);
+    require(_homeBridgeOwner != address(0));
+    require(_homeProxyOwner != address(0));
+
+    setOwner(_owner);
+    setBridgeValidatorsImplementation(_bridgeValidatorsImplementation);
```
patitonar: While testing, I found an issue on how variables are set on `initialize` method. Calling methods that has the `onlyOwner` modifier will only work if the deployment account(`DEPLOYMENT_ACCOUNT_PRIVATE_KEY`) is the same as `HOME_FACTORY_OWNER`. This may not always be the case, setting multisignature wallets as owner of contracts is a common pattern.

To avoid this issue, on `initialize` method of other bridge modes we set the values directly:
https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol#L38-L46

Another option could be to define one internal method for every parameter like the following:
```solidity
  function _setBridgeValidatorsImplementation(address _bridgeValidatorsImplementation) internal {
    addressStorage[keccak256(abi.encodePacked("bridgeValidatorsImplementation"))] = _bridgeValidatorsImplementation;
  }
```
And call it from `setBridgeValidatorsImplementation` and `initialize`.

The same issue should be addressed on `ForeignBridgeFactory`

contracts/upgradeable_contracts/factories/ForeignBridgeFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+
+  function setRequiredBlockConfirmations(uint256 _requiredBlockConfirmations) public onlyOwner {
+    uintStorage[keccak256(abi.encodePacked("requiredBlockConfirmations"))] = _requiredBlockConfirmations;
+  }
+
+  function gasPrice() public view returns(uint256) {
+    return uintStorage[keccak256(abi.encodePacked("gasPrice"))];
+  }
+
+  function setGasPrice(uint256 _gasPrice) public onlyOwner {
```
patitonar: It looks like there are some methods that are repeated between `HomeBridgeFactory` and `ForeignBridgeFactory`. Does it makes sense to define a common contract where this shared logic could be stored and then extend from that contract? You can take a look at `BasicBridge` as an example https://github.com/poanetwork/poa-bridge-contracts/blob/master/contracts/upgradeable_contracts/BasicBridge.sol

contracts/upgradeable_contracts/factories/ForeignBridgeFactory.sol
```diff
@@ -0,0 +1,189 @@
+pragma solidity 0.4.24;
+
+import "../../IBridgeValidators.sol";
+import "../../IForeignBridge.sol";
+import "../../upgradeability/EternalStorageProxy.sol";
+import "../../upgradeability/EternalStorage.sol";
+import "../EternalOwnable.sol";
+
+contract ForeignBridgeFactory is EternalStorage, EternalOwnable {
```
patitonar: I think it could be a good idea to include some reference to the type of bridge the factory is about on the name of the contracts. Some options could be `HomeErcToErcBridgeFactory`, `ForeignErcToErcBridgeFactory` and `ErcToErcBridgeMapper`. Does it makes sense @akolotov ?

akolotov:
@bejavu @LiorRabin probably I read the code too quickly but how about the contracts upgrade scenario? If we found some security vulnerability in the bridge contracts how can we upgrade all the contracts born by the factory? Could you clarify the procedure a bit?

deploy/.env.example
bejavu: You are correct, this variable are obsolete, I will remove them
Also, I will add the optional variables to the `.env` templates as you suggested.

contracts/upgradeability/EternalStorage.sol
bejavu: I use the mapping `addressArrayStorage` for the `initialValidators` in the factories.

contracts/upgradeable_contracts/factories/HomeBridgeFactory.sol
bejavu: sure, will be updated

contracts/upgradeable_contracts/factories/HomeBridgeFactory.sol
bejavu: Although the `deployHomeBridge` calls the `BridgeValidators.initialize` and will fail if the wrong initial validator number and signatures requirements are set, I will add the require check to the `initialize` method and to the set `setInitialValidators` and `setRequiredSignatures` in both factories.

contracts/upgradeable_contracts/factories/HomeBridgeFactory.sol
bejavu: I understand the problem you mentioned, I will suggest (and implement) a third solution:
First set the owner as the `msg.sender` so that I could call all the setters from within the `initialize` method and at the end of the method, I will set the owner to be the `_owner` parameter.

contracts/upgradeable_contracts/factories/ForeignBridgeFactory.sol
bejavu: good idea, will do

contracts/upgradeable_contracts/factories/ForeignBridgeFactory.sol
bejavu: As I don't really see a real value for a native (from\to) erc factory I am not sure if this change is a must have one.

bejavu:
> @bejavu @LiorRabin probably I read the code too quickly but how about the contracts upgrade scenario? If we found some security vulnerability in the bridge contracts how can we upgrade all the contracts born by the factory? Could you clarify the procedure a bit?

The same as every other contract in your repo:
Every bridge and bridge validators contract is deployed with proxy.
The default owner of the proxies are set on the factory initialize method and can be changed by the factory setters in the future.
The owner of the proxy can upgrade the deployed bridge \ validators contracts.

Another method is to remove the default owner set by the factory and let it be chosen in the deploy method as a parameter. But since this method is an `ownerOnly` method and not public one, the owner of the factory can reassign the bridge proxy owner later.

bejavu:
@akolotov @patitonar,
Thank you both for all the comments.
I updated the PR to support all the changed that advised.

akolotov:
> The same as every other contract in your repo:
Every bridge and bridge validators contract is deployed with proxy.

Imagine that you have 20 bridge created by factory. So, do you need to perform 20 upgrades manually? It could nightmare especially if the upgradability admin is a multisig wallet.

bejavu:
> Imagine that you have 20 bridge created by factory. So, do you need to perform 20 upgrades manually? It could nightmare especially if the upgradability admin is a multisig wallet.
You can make a short script for that. The migration should be to the same implementation. 
Also, we can do a double proxy to upgrade all at once. 
To create a function for that in the factory is bad approach IMHO, since we need to save the deployed bridges, paginate it and skip failed once due to owner change.

akolotov:
I am just thinking we can get rid of the bridge factory at all and implement this on one bridge contract.

In this case it could be as following:
For the Foreign side:
1. Every `Transfer` watcher looks for the specific ERC20 token.
2. As soon as the watcher catches the `Transfer` event to the bridge address it confirms the relay operation by sending approval to the `HomeBridge` contract. The confirmation message contains: the value, the sender and the ERC20 token contract address (on the Foreign side). 
3. As soon as enough number of confirmations received, the HomeBridge contract **checks that the ERC20 token address corresponds to an ERC677 token and if the relationship is found**, mints new corresponding ERC677 tokens.

For the Home side:
1. A user calls `transferAndCall` on an ERC677 token
2. The `HomeBridge` contracts checks if the token is registered in the mapper and raise the event to requests validators to confirm the relay. **In the request it sends the foreign ERC20 token contract address.**
3. The validator confirms that `N` tokens are sent by a `sender` to be relayed **as the exact `foreign ERC20 token`**.
4. As soon as enough number of signatures collected, one of the validators relays the message.
5. The `ForeignBridge` contract checks signatures and releases `N` ERC20 tokens related to the exact `token contract` for the `sender`. 

This approach still uses the mapper, it allows to simplify the upgrade procedure for the bridge contracts (actually it remains the same), it reduces complexity of the code by getting rid of the bridge factory.
It requires changes on the token-bridge side to accept modified events but such behavior could be default for the one-to-one `erc-to-erc` bridge (and even can be considered to be OK for other bridge modes). Another part on the token bridge side is to manage watchers when new pair of tokens are registered or unregistered but you are implementing it anyway. But new watcher is needed only to watch `Transfers` events. The same watcher can be used to look at `UserRequestForSignature` events.

It still in line with the paradigm that user now still be able to relay signed confirmation from Home to Foreign if the validator does not perform this by some reason (e.g. not enough funds). Since the ERC20 token address will be part of the signed message, it means that the correct tokens will be released. My understanding is that for your implementation the user needs to know exact pair of the bridge contracts to perform the same.

The nice part with this approach also that contractAddress field is already is part of the message that received by the `ForeignBridge` contract:
https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/BasicForeignBridge.sol#L19

bejavu:
> I am just thinking we can get rid of the bridge factory at all and implement this in one bridge contract...

This is interesting, I will look into it tomorrow.

> It requires changes on the token-bridge side to accept modified events
I will also look at this one, since we already updated the token-bridge to listen to the mapper events to updates on new tokens bridge mapping. I think there won't be any big change then the one already done.

> My understanding is that for your implementation the user needs to know exact pair of the bridge contracts to perform the same.

This is the main issue with that approach. A foreign token can have only one bridge and one mirrored home token. If so, then the user don't need to know anything but the home token address, all other attributes can be taken from the mapper where the foreign token is the key,

akolotov:
> You can make a short script for that. The migration should be to the same implementation.
Also, we can do a double proxy to upgrade all at once.
To create a function for that in the factory is bad approach IMHO, since we need to save the deployed bridges, paginate it and skip failed once due to owner change.

The mutlisig wallet owners could have no opportunity to run the script. Now the upgrade may happen with usage of a wallet software only.

Here is how I see the upgrade could happen with your approach:
1. Deploy new implementation of the contract
2. Find all events where the bridges were deployed to find the bridge addresses. Try to filter out the bridges that are not used (your current code do not provide the way to remove the bridges)
3. Create a transaction to call `upgradeTo` for every bridge to be sent in the multisig. If a bridge requires some specific migration procedure you need to call `upgradeToAndCall`.
4. Ask other owners of this multisig wallet to make the number of transactions correspondingly number of bridges to confirm the implementation upgrade.

If some bridges are owned by different multisigs you need to communicate to the multisig owners to perform the step 3 and 4 by their own.

akolotov:
> If you want to make an easy to use DApp where a user (or an admin) wants to deploy its own bridge for its own token using only one transaction in a web browser, you will need a factory to encapsulated all of the deployment transactions into one transaction to be signed.

Correct me if I wrong: there should be two transactions: one is to deploy the bridge, another is to register new bridge in the mapper.

akolotov:
> A foreign token can have only one bridge and one mirrored home token. If so, then the user don't need to know anything but the home token address, all other attributes can be taken from the mapper where the foreign token is the key,

From this perspective if we have the mapper it does not matter how many bridges we have - one, two or dozens:
- if we have the pair of the bridge contracts for every pair of the tokens, the bridge contracts can be got from the mapper.
- if we have one pair of the bridge contracts that serves all pairs of the tokens, the bridge contract addresses are public and widely known.

bejavu:
To sums up:

There are two approaches for erc-to-erc-multi bridge.

1. By updating the bridges to support multiple tokens.
2. By deploying multiple bridges to each token

Both approaches needs some kind of a mapper to:
1. Map between the bridges and tokens on both networks.
2. Fire events for new token-bridge support or a removal of one.

A change to the oracle (https://github.com/poanetwork/token-bridge) is needed in both approaches.

The first approach needs a better understanding of the bridge parameters and limitation while on the second approach a bridge can be considered as a black box.
I chose the second approach because of it.

@akolotov @patitonar Can you please provide me a list of bridge parameters \ limitation names and purpose so I can make a new PR for the first approach as well?

The new PR will use a mapper to map between the foreign token as a key to all of its related attributes (bridges contracts, home token, bridge limitations). The mapper will validate the token transfers for the bridge after the signatures have processed.
Based in that suggestion: https://github.com/poanetwork/poa-bridge-contracts/pull/140#issuecomment-460067194

patitonar:
@bejavu Here is a list of parameters set on initialization of bridges:
- Used by validators on `token-bridge`: 
    - `gasPrice`: Default gas price for the network.
    - `requiredBlockConfirmations`: The number of block validators should wait before processing the events

- Security limits to user's behavior. Used on methods called by users when they initiates a transfer:  
    - `dailyLimit`: Daily transaction limit, as soon as this limit is exceeded, any transaction which requests to relay assets will fail.
    - `maxPerTx`: The maximum limit for one transaction.
    - `minPerTx`: The minimum limit for one transaction.
Example of how they are used on `erc-to-erc` bridge. On home network when a user call `transferAndCall` on the ERC677, `onTokenTransfer` from bridge contract is called:
https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/ERC677Bridge.sol#L17-L24
`whitinLimit` checks that the value is inside the limits:
https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/BasicBridge.sol#L120-L123

- Security limits to validator's behavior. Used on methods invoked by validators to execute the relay of assets:
    - `executionDailyLimit`: Daily transaction execution limit. Should be the same as `dailyLimit` from the other network. 
    - `executionMaxPerTx`: The maximum execution limit per transaction. Should be the same as `maxPerTx` from the other network.
Following the example mentioned before, after signatures were collected on Home side. A validator will call `executeSignatures` . In that method `withinExecutionLimit` is called to check that the value is inside the limits allowed to unlock and transfer the tokens.
https://github.com/poanetwork/poa-bridge-contracts/blob/46151629adb3dd20e6fb47d4ba91b88c6d388c4f/contracts/upgradeable_contracts/BasicBridge.sol#L125-L128

Let me know if you need any other explanation or details on a bridge parameter or any method.

bejavu:
Thank you @patitonar

bejavu:
@akolotov I added the option to double proxy the implementation during the deployment process as discussed.
For now, this approach is finished.

deploy/.env.example
```diff
@@ -35,6 +38,13 @@ FOREIGN_GAS_PRICE=10000000000
 #for bridge erc_to_erc and erc_to_native mode
 ERC20_TOKEN_ADDRESS=
 
+#Optional Implemetation addresses for ERC_TO_ERC_MULTIPLE mode
+HOME_BRIDGE_VALIDATORS_IMPLEMENTATION_ADDRESS=0x
+HOME_BRIDGE_IMPLEMENTATION_ADDRESS=0x
+FOREIGN_BRIDGE_VALIDATORS_IMPLEMENTATION_ADDRESS=0x
+FOREIGN_BRIDGE_IMPLEMENTATION_ADDRESS=0x
+DOUBLE_PROXY_IMPLEMENTATIONS=
```
akolotov: Does it make sense to provide opportunity to disable this? Would it be worth to use double proxy always if ERC_TO_ERC_MULTIPLE configuration is used?

bejavu: > Does it make sense to provide opportunity to disable this?
In my opinion, yes.
Different bridges (implementation version) can be applied here.

LiorRabin:
Hello @patitonar,
I've just added unit tests and updates to the gas consumption doc.

bejavu:
@akolotov
Thanks to @LiorRabin We updated the `Proxy` contract to support multiple nested proxies capabilities in this commit:  [06a396bd](https://github.com/poanetwork/poa-bridge-contracts/pull/140/commits/06a396bd7e52b590437b66bed3c23f3bcaf97781).

In a nutshell, you can't nest a proxy because the state persists and the `implementation` of the inner proxy is irrelevant since the outer proxy `implementation` is being used again, making the code go it to an infinite loop.
To solve it, we first need to see if the implementation is an inner proxy, defined by if it has an `implementation()` function.
```if (_impl.call(0x5c60da1b)) { // bytes(keccak256("implementation()"))```
then change the state variable `_implementation` of the outer proxy to be the implementation of the inner proxy, before the delegate call, thus making all the state remain *except* the `_implementation` variable.
```this.setImplementation(_innerImpl);```
After the delegate call is finished, revert the state variable `_implementation` before returning the delegate call result.
```
let x := mload(0x40)
mstore(x, sig)
mstore(add(x, 0x04), _impl)
let success := call(gas, thisAddress, 0, x, 0x24, x, 0x0)
```

File: contracts/upgradeability/Proxy.sol [hunk 15]
```diff
@@ -13,13 +13,25 @@ contract Proxy {
...
+
   /**
   * @dev Fallback function allowing to perform a delegatecall to the given implementation.
   * This function will return whatever the implementation call returns
   */
     function () payable public {
         address _impl = implementation();
         require(_impl != address(0));
+
+        address _innerImpl;
```
akolotov: the same proxy contract is used for validators contract. Could you clarify how these changes will affect it?

bejavu: Sorry, I didn't understand.
This is just to solve the nested proxy infinite loop problem.
If there is no nested proxy, meaning the implementation has no `implementation()` function, then the old proxy `delegatecall` will take place. I don't see where it could break stuff, but maybe I missed something.

akolotov: You are changing the contract that is used in several for the validator contract as well. How your changes will affect this contract?

bejavu: IMHO, no affect other functionality in other contracts, as long as you don't use an implementation with `implementation()` method in it, unless it is the nested proxy, as intended here.
But a good comprehensive audit should take place in ant case.

akolotov: is my understanding correct and we will increase gas usage of every call by two SSTORE command?

bejavu: Only if you use the double proxy. If not, no new sstore are added

akolotov: OK. Understood.

PR #156, opened 2019-02-28
Support transaction null status on deploy script

patitonar:
This PR takes #155 as base to include those changes on the deployment script. So it should be merged after that PR.

For every contract deployment if `status` is not success it will also check for `tx.contractAddress` in case `status` is not present.
https://github.com/poanetwork/poa-bridge-contracts/blob/b419821b4b38580d5a090ed4bda1867f9b62dc80/deploy/src/deploymentUtils.js#L50-L52

For transactions that changes the state if `status` is not present it will check if the state was changed. Most of these checks can be found on `deploymentUtils.js`
For example when `upgradeTo` is called:
https://github.com/poanetwork/poa-bridge-contracts/blob/b419821b4b38580d5a090ed4bda1867f9b62dc80/deploy/src/deploymentUtils.js#L159-L163

Also a retry functionality was implemented for when we check the new state so it can avoid any race conditions. If after all retries, state didn't change, it will fail and stop the script. It uses default configuration for retry that detailed [here](https://github.com/IndigoUnited/node-promise-retry#promiseretryfn-options)


deploy/src/erc_to_erc/foreign.js
```diff
@@ -40,154 +87,102 @@ async function deployForeign() {
...
   console.log('\nhooking up eternal storage to BridgeValidators')
-  const upgradeToBridgeVForeignData = await storageValidatorsForeign.methods
-    .upgradeTo('1', bridgeValidatorsForeign.options.address)
-    .encodeABI({ from: DEPLOYMENT_ACCOUNT_ADDRESS })
-  const txUpgradeToBridgeVForeign = await sendRawTxForeign({
-    data: upgradeToBridgeVForeignData,
-    nonce: foreignNonce,
-    to: storageValidatorsForeign.options.address,
-    privateKey: deploymentPrivateKey,
+  await upgradeProxy({
```
akolotov: `upgradeProxy` in `src/deploymentUtils.js` calls `sendRawTxHome` and it means that `HOME_DEPLOYMENT_GAS_PRICE` will be used for the transaction in the Foreign chain. Don't think that it is correct. The same is applicable for other new function.

deploy/src/deploymentUtils.js
```diff
@@ -147,13 +147,128 @@ function logValidatorsAndRewardAccounts(validators, rewards) {
...
+  validators,
+  rewardAccounts,
+  owner,
+  nonce,
+  url
+}) {
+  let data
+
+  if (isRewardableBridge) {
+    console.log(`REQUIRED_NUMBER_OF_VALIDATORS: ${requiredNumber}, HOME_VALIDATORS_OWNER: ${owner}`)
```
akolotov: Since this functionality could be used to initialize validators in the Foreign chain too it makes sense to remove the prefix `HOME_`. The same is applicable to the similar log message bellow.

patitonar: Thanks! Fixed here https://github.com/poanetwork/poa-bridge-contracts/pull/156/commits/d92dd655faae5068c355551a7b5893ba90afb54d

deploy/src/erc_to_erc/foreign.js
patitonar: Good catch! Thank you. Fixed here https://github.com/poanetwork/poa-bridge-contracts/pull/156/commits/f022bed9d19ac9126a56c8f4c5e272dfed66bdc5

PR #161, opened 2019-03-14
Migrate contracts to solidity version 0.4.19

patitonar:
Base branch was created based on `support-null-status-deploy-script`

### DIFF
--- a/contracts/upgradeability/ClassicEternalStorageProxy.sol
+++ b/contracts/upgradeability/ClassicEternalStorageProxy.sol
@@ -1,0 +1,41 @@
+pragma solidity 0.4.19;
+
+import "./EternalStorage.sol";
+import "./OwnedUpgradeabilityProxy.sol";
+
+
+/**
+ * @title EternalStorageProxy
+ * @dev This proxy holds the storage of the token contract and delegates every call to the current implementation set.
+ * Besides, it allows to upgrade the token's behaviour towards further implementations, and provides basic
+ * authorization control functionalities
+ */
+contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage {
+
+    function () payable public {
+        address _impl = implementation();
+        require(_impl != address(0));
+        bytes4 sig;
+        assembly { sig := calldataload(0) }
+        uint256 len = getSize(sig);
+        assembly {
+            calldatacopy(0x0, 0x0, calldatasize)
+            let result := delegatecall(gas, _impl, 0x0, calldatasize, 0, len)
+
+            switch result
+            case 0 { revert(0, len) }
+            default { return(0, len) }
+        }
+    }
+
+    function getSize(bytes4 sig) public pure returns(uint256) {
+        if (bytes4(sha3("signature(bytes32,uint256)")) == sig) {
+            return 132;
+        }
+        if (bytes4(sha3("message(bytes32)")) == sig) {
+            return 210;
+        }
+
+        return 32;
+    }
+}
--- a/contracts/upgradeability/ClassicEternalStorageProxy.sol
+++ b/contracts/upgradeability/ClassicEternalStorageProxy.sol
@@ -4,17 +4,16 @@ import "./EternalStorage.sol";
 import "./OwnedUpgradeabilityProxy.sol";
 
 
-contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage {
+contract ClassicEternalStorageProxy is EternalStorage, OwnedUpgradeabilityProxy {
 
     function () payable public {
         address _impl = implementation();
         require(_impl != address(0));
-        bytes4 sig;
-        assembly { sig := calldataload(0) }
-        uint256 len = getSize(sig);
+        uint256 len = getSize(msg.sig);
         assembly {
-            calldatacopy(0x0, 0x0, calldatasize)
-            let result := delegatecall(gas, _impl, 0x0, calldatasize, 0, len)
+            let ptr := mload(0x40)
+            calldatacopy(ptr, 0, calldatasize)
+            let result := delegatecall(gas, _impl, ptr, calldatasize, 0, len)
 
             switch result
             case 0 { revert(0, len) }
@@ -22,14 +21,11 @@ contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage 
         }
     }
 
-    function getSize(bytes4 sig) public pure returns(uint256) {
-        if (bytes4(sha3("signature(bytes32,uint256)")) == sig) {
-            return 132;
+    function getSize(bytes4 sig) public view returns(uint256) {
+        uint256 ret = uintStorage[keccak256("dataSizes", sig)];
+        if (ret == 0) {
+            ret = 32;
         }
-        if (bytes4(sha3("message(bytes32)")) == sig) {
-            return 210;
-        }
-
-        return 32;
+        return ret;
     }
 }
--- a/contracts/upgradeability/ClassicEternalStorageProxy.sol
+++ b/contracts/upgradeability/ClassicEternalStorageProxy.sol
@@ -4,12 +4,6 @@ import "./EternalStorage.sol";
 import "./OwnedUpgradeabilityProxy.sol";
 
 
-/**
- * @title EternalStorageProxy
- * @dev This proxy holds the storage of the token contract and delegates every call to the current implementation set.
- * Besides, it allows to upgrade the token's behaviour towards further implementations, and provides basic
- * authorization control functionalities
- */
 contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage {
 
     function () payable public {
--- a/contracts/upgradeable_contracts/BaseFeeManager.sol
+++ b/contracts/upgradeable_contracts/BaseFeeManager.sol
@@ -59,7 +59,11 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
 
     function distributeFeeProportionally(uint256 _fee, bytes32 _direction) internal {
         IRewardableValidators validators = rewardableValidatorContract();
-        address[] memory validatorList = validators.validatorList();
+        address[] memory validatorList;
+        validators.validatorList();
+        assembly {
+            returndatacopy(validatorList, 0, returndatasize)
+        }
         uint256 feePerValidator = _fee.div(validatorList.length);
 
         uint256 randomValidatorIndex;
--- a/contracts/upgradeable_contracts/BaseFeeManager.sol
+++ b/contracts/upgradeable_contracts/BaseFeeManager.sol
@@ -60,35 +60,32 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
     function distributeFeeProportionally(uint256 _fee, bytes32 _direction) internal {
         IRewardableValidators validators = rewardableValidatorContract();
         address F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
-        address [] memory validatorList = new address[](validators.validatorCount());
-        uint256 counter = 0;
-        address nextValidator = validators.getNextValidator(F_ADDR);
-        require(nextValidator != address(0));
+        uint256 numOfValidators = validators.validatorCount();
 
-        while (nextValidator != F_ADDR) {
-            validatorList[counter] = nextValidator;
-            nextValidator = validators.getNextValidator(nextValidator);
-            counter++;
-
-            if (nextValidator == address(0) ) {
-                revert();
-            }
-        }
-        uint256 feePerValidator = _fee.div(validatorList.length);
+        uint256 feePerValidator = _fee.div(numOfValidators);
 
         uint256 randomValidatorIndex;
-        uint256 diff = _fee.sub(feePerValidator.mul(validatorList.length));
+        uint256 diff = _fee.sub(feePerValidator.mul(numOfValidators));
         if (diff > 0) {
-            randomValidatorIndex = random(validatorList.length);
+            randomValidatorIndex = random(numOfValidators);
         }
 
-        for (uint256 i = 0; i < validatorList.length; i++) {
+        address nextValidator = validators.getNextValidator(F_ADDR);
+        require((nextValidator != F_ADDR) && (nextValidator != address(0)));
+
+        uint256 i = 0;
+        while (nextValidator != F_ADDR) {
             uint256 feeToDistribute = feePerValidator;
             if (diff > 0 && randomValidatorIndex == i) {
                 feeToDistribute = feeToDistribute.add(diff);
             }
-            address rewardAddress = validators.getValidatorRewardAddress(validatorList[i]);
+
+            address rewardAddress = validators.getValidatorRewardAddress(nextValidator);
             onFeeDistribution(rewardAddress, feeToDistribute, _direction);
+
+            nextValidator = validators.getNextValidator(nextValidator);
+            require(nextValidator != address(0));
+            i = i + 1;
         }
     }
 
--- a/contracts/upgradeable_contracts/BaseFeeManager.sol
+++ b/contracts/upgradeable_contracts/BaseFeeManager.sol
@@ -59,10 +59,20 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
 
     function distributeFeeProportionally(uint256 _fee, bytes32 _direction) internal {
         IRewardableValidators validators = rewardableValidatorContract();
-        address[] memory validatorList;
-        validators.validatorList();
-        assembly {
-            returndatacopy(validatorList, 0, returndatasize)
+        address F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+        address [] memory validatorList = new address[](validators.validatorCount());
+        uint256 counter = 0;
+        address nextValidator = validators.getNextValidator(F_ADDR);
+        require(nextValidator != address(0));
+
+        while (nextValidator != F_ADDR) {
+            validatorList[counter] = nextValidator;
+            nextValidator = validators.getNextValidator(nextValidator);
+            counter++;
+
+            if (nextValidator == address(0) ) {
+                revert();
+            }
         }
         uint256 feePerValidator = _fee.div(validatorList.length);
 
--- a/contracts/upgradeable_contracts/BaseFeeManager.sol
+++ b/contracts/upgradeable_contracts/BaseFeeManager.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.4.24;
+pragma solidity 0.4.19;
 
 import "../upgradeability/EternalStorage.sol";
 import "../libraries/SafeMath.sol";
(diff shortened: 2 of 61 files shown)


contracts/upgradeable_contracts/BaseFeeManager.sol
```diff
@@ -54,12 +54,26 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
     function getFeeManagerMode() public pure returns(bytes4);
 
     function random(uint256 _count) public view returns(uint256) {
-        return uint256(blockhash(block.number.sub(1))) % _count;
+        return uint256(block.blockhash(block.number.sub(1))) % _count;
     }
 
     function distributeFeeProportionally(uint256 _fee, bytes32 _direction) internal {
         IRewardableValidators validators = rewardableValidatorContract();
-        address[] memory validatorList = validators.validatorList();
```
patitonar: This line throws the following error:
```
TypeError: Type inaccessible dynamic type is not implicitly convertible to expected type address[] memory.
```
It seems that it is possible to return variably-sized data from a method, but it just cannot be read from within the EVM before the Byzantium hardfork.
https://ethereum.stackexchange.com/questions/36637/returning-static-array-inaccessible-dynamic-type/36652

So I had to duplicate `validatorList()` logic to get the list inside `distributeFeeProportionally`. I didn't like this solution but I couldn't find any other option to make it work.

contracts/upgradeable_contracts/BaseFeeManager.sol
```diff
@@ -54,12 +54,26 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
...
+        address F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+        address [] memory validatorList = new address[](validators.validatorCount());
```
akolotov: We could reduce the number of iterations. Now it is 2*N, where N is the number of validators.
Please consider the following enhancement:
```solidity
IRewardableValidators validators = rewardableValidatorContract();
address F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
numOfValidators = validators.validatorCount()

uint256 feePerValidator = _fee.div(numOfValidators);

uint256 randomValidatorIndex;
uint256 diff = _fee.sub(feePerValidator.mul(numOfValidators));
if (diff > 0) {
    randomValidatorIndex = random(numOfValidators);
}

address nextValidator = validators.getNextValidator(F_ADDR);
require((nextValidator != F_ADDR) && (nextValidator != address(0)));

uint256 = 0;
while (nextValidator != F_ADDR) {
    uint256 feeToDistribute = feePerValidator;
    if (diff > 0 && randomValidatorIndex == i) {
        feeToDistribute = feeToDistribute.add(diff);
    }

    address rewardAddress = validators.getValidatorRewardAddress(nextValidator);
    onFeeDistribution(rewardAddress, feeToDistribute, _direction);

    nextValidator = validators.getNextValidator(nextValidator);
    require(nextValidator != address(0))
    i = i + 1;
}
```

package.json
```diff
@@ -14,8 +14,8 @@
   "license": "GPLv3",
   "dependencies": {
     "ganache-cli": "^6.1.0",
-    "openzeppelin-solidity": "^1.10.0",
-    "truffle": "^4.1.11",
+    "openzeppelin-solidity": "^1.8.0",
+    "truffle": "^4.0.6",
```
phahulin: You might want to specify exact versions here
```
    "openzeppelin-solidity": "1.8.0",
    "truffle": "4.0.6",
```

phahulin:
I'm having troubles making this compile, there seems to be something wrong with dependencies. Please try to clone into a new folder and check if it compiles. There might also be something on my side...

Errors I was having:
1. `npm install` didn't work:
```bash
➜  npm install
npm ERR! code ENOLOCAL
npm ERR! Could not install from "node_modules/solidity-coverage/node_modules/web3/bignumber.js@git+https:/github.com/debris/bignumber.js.git#94d7146671b9719e00a09c29b01a691bc85048c2" as it does not contain a package.json file.
```
Removed `package-lock.json` and `deploy/package-lock.json` files and did npm install again - fixed

2. during deployment (used docker-compose) I was getting solidity compilation errors suggesting it was still using version 0.4.24 - fixed it by specifying exact versions of truffle and openzeppelin-solidity packages in package.json (see above).

3. then this error:
```bash
/contracts/deploy/src/loadEnv.js:7
const { isAddress, toBN } = require('web3').utils
        ^

TypeError: Cannot destructure property `isAddress` of 'undefined' or 'null'.
```
which seems to indicate that web3 is of wrong version, I tried to replace it with `new Web3(...)`

4. but still got the error later
```
Bridge mode: NATIVE_TO_ERC
Error: TypeError: Cannot read property 'constructor' of undefined
    at ProviderResolver.isWeb3Provider (/contracts/deploy/node_modules/web3-providers/dist/web3-providers.cjs.js:67:24)
    at ProviderResolver.resolve (/contracts/deploy/node_modules/web3-providers/dist/web3-providers.cjs.js:50:16)
    at Web3.AbstractWeb3Module (/contracts/deploy/node_modules/web3-core/dist/web3-core.cjs.js:26:51)
    at new Web3 (/contracts/deploy/node_modules/web3/dist/web3.cjs.js:31:68)
    at privateKeyToAddress (/contracts/deploy/src/deploymentUtils.js:140:10)
    at Object.<anonymous> (/contracts/deploy/src/native_to_erc/home.js:44:36)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at deployNativeToErc (/contracts/deploy/deploy.js:10:22)
    at main (/contracts/deploy/deploy.js:117:13)
    at Object.<anonymous> (/contracts/deploy/deploy.js:131:1)
```

contracts/upgradeable_contracts/BaseFeeManager.sol
patitonar: Thanks for the suggestion! I applied the changes here https://github.com/poanetwork/poa-bridge-contracts/pull/161/commits/fb1a221fef7cec25c043270a2cc2fec8a47ed54e

package.json
patitonar: Thanks. I specified exact version of `truffle` and `openzeppelin-solidity`. Also exact version of `web3` on `deploy/package.json` and commited the updated `package-lock.json` files. This should fix the errors you mentioned

phahulin:
Thanks, now it started to deploy :tada: but failed with VM error :disappointed: 
Here's the log :spiral_notepad: 
```Bridge contract deployment started
Bridge mode: NATIVE_TO_ERC
deploying storage for home validators
pending txHash 0xd298c242798a32619b971136363749ff26277d41b0c56824acc5d80970125836
[Home] BridgeValidators Storage:  0x0cB781EE62F815bdD9CD4c2210aE8600d43e7040

deploying implementation for home validators
pending txHash 0xdc3ed778030138784271e1ecf7b94c198f87db251fc7db4cfe562252fb9db072
[Home] BridgeValidators Implementation:  0xEc7c0AB3C2f165e790da152624DAb84633bE14F2

hooking up eternal storage to BridgeValidators
pending txHash 0xb7555e4177e46668aabd44410a5ee99b56d589136e15c999d2e3ef876f258dcd

initializing Home Bridge Validators with following parameters:

REQUIRED_NUMBER_OF_VALIDATORS: 1, HOME_VALIDATORS_OWNER: 0x131c608ee32e18430edb439296d1d40819b509e9
VALIDATORS
==========
1: 0x131c608ee32e18430edb439296d1d40819b509e9, reward address 0x131c608ee32e18430edb439296d1d40819b509e9
pending txHash 0x928b1a91cbe5437b38356c2368bf9f6158b640dbf0c9df712b3d24d3b50f1e7c
Error: Error: Returned error: VM execution error.
    at Object.ErrorResponse (/contracts/deploy/node_modules/web3-core-helpers/src/errors.js:29:16)
    at /contracts/deploy/node_modules/web3-core-requestmanager/src/index.js:140:36
    at XMLHttpRequest.request.onreadystatechange (/contracts/deploy/node_modules/web3-providers-http/src/index.js:77:13)
    at XMLHttpRequestEventTarget.dispatchEvent (/contracts/deploy/node_modules/xhr2/lib/xhr2.js:64:18)
    at XMLHttpRequest._setReadyState (/contracts/deploy/node_modules/xhr2/lib/xhr2.js:354:12)
    at XMLHttpRequest._onHttpResponseEnd (/contracts/deploy/node_modules/xhr2/lib/xhr2.js:509:12)
    at IncomingMessage.<anonymous> (/contracts/deploy/node_modules/xhr2/lib/xhr2.js:469:24)
    at emitNone (events.js:111:20)
    at IncomingMessage.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1064:12)
    at _combinedTickCallback (internal/process/next_tick.js:139:11)
    at process._tickCallback (internal/process/next_tick.js:181:9)
```

patitonar:
@phahulin Is this error present when you try to deploy to any chain? or only when deploying to ETC?

phahulin:
I was able to deploy on sokol+kovan without errors, so I suppose there's something missing in ETC.

akolotov:
@patitonar could we try to deploy contracts through kotti? or deploy your own chain by using the same genesis that is used in Parity for the classic network: https://github.com/paritytech/parity-ethereum/blob/a9a278a6e146895cf03387f8db89c25cc031c3c2/ethcore/res/ethereum/classic.json

contracts/upgradeability/ClassicEternalStorageProxy.sol
```diff
@@ -0,0 +1,35 @@
...
+contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage {
+
+    function () payable public {
+        address _impl = implementation();
+        require(_impl != address(0));
+        bytes4 sig;
+        assembly { sig := calldataload(0) }
+        uint256 len = getSize(sig);
+        assembly {
+            calldatacopy(0x0, 0x0, calldatasize)
```
akolotov: consider to use the way implemented in the original `Proxy`:
```solidity
let ptr := mload(0x40)
calldatacopy(ptr, 0, calldatasize)
let result := delegatecall(gas, _impl, ptr, calldatasize, 0, len)
```

contracts/upgradeability/ClassicEternalStorageProxy.sol
```diff
@@ -0,0 +1,35 @@
...
+            let result := delegatecall(gas, _impl, 0x0, calldatasize, 0, len)
+
+            switch result
+            case 0 { revert(0, len) }
+            default { return(0, len) }
+        }
+    }
+
+    function getSize(bytes4 sig) public pure returns(uint256) {
```
akolotov: Consider to create a separate file `ReturnDataSizes.sol` with the following content:
```solidity
contract ReturnDataSizesRegistrar {
    mapping(bytes4 => uint256) internal returnDataSizes;
}
```

so, the `ClassicEternalStorageProxy` contract and its `getSize()` method could look like:
```solidity
import "./OwnedUpgradeabilityProxy.sol";
import "./ReturnDataSizes.sol";

contract ClassicEternalStorageProxy is ReturnDataSizesRegistrar, OwnedUpgradeabilityProxy, EternalStorage {
    . . .
    . . .
    function getSize(bytes4 sig) public pure returns(uint256) {
        uint256 ret = returnDataSizes[sig];
        if (ret == 0) {
            ret = 32;
        }
        return ret;
    }
```

In this case `BasicHomeBridge.sol` will be changed like this:
```solidity
import "../libraries/Message.sol
import "../upgradeability/ReturnDataSizesRegistrar.sol";

contract BasicHomeBridge is ReturnDataSizesRegistrar, EternalStorage, Validatable {
    using SafeMath for uint256;
    returnDataSizes[bytes4(sha3("signature(bytes32,uint256)")] = 132;
    returnDataSizes[bytes4(sha3("message(bytes32)")] = 210;
```

akolotov: This will provide more logical way to define returned data sizes in the modules where the corresponding method is defined. And will reduce a bit the size of other contracts using the proxy template (e.g. the Validator contract).

deploy/src/erc_to_erc/home.js
```diff
@@ -15,7 +15,7 @@ const {
 } = require('../deploymentUtils')
 const { web3Home, deploymentPrivateKey, HOME_RPC_URL } = require('../web3')
 
-const EternalStorageProxy = require('../../../build/contracts/EternalStorageProxy.json')
+const EternalStorageProxy = require('../../../build/contracts/ClassicEternalStorageProxy.json')
```
akolotov: I think this is from your local testing environment and is not needed so far for the `native-to-erc` bridge

contracts/upgradeable_contracts/BaseFeeManager.sol
akolotov: I think implementing the Fee Manager in such manner will reduce the gas consumption and the code of the validator contract even for ordinary version of the bridge contract.
I mean we can implement this for the `119-Epic-rewards-for-bridge-validators` branch as well.
What do you think?

contracts/upgradeability/ClassicEternalStorageProxy.sol
```diff
@@ -0,0 +1,35 @@
...
+import "./OwnedUpgradeabilityProxy.sol";
+
+
+contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage {
+
+    function () payable public {
+        address _impl = implementation();
+        require(_impl != address(0));
+        bytes4 sig;
+        assembly { sig := calldataload(0) }
```
akolotov: We can remove this and use:
```solidity
address _impl = implementation();
require(_impl != address(0));
uint256 len = getSize(msg.sig);
```

contracts/upgradeable_contracts/BaseFeeManager.sol
patitonar: I think it's a good idea to implement this on `119-Epic-rewards-for-bridge-validators`. We should create a new issue for this

contracts/upgradeability/ClassicEternalStorageProxy.sol
akolotov: And the question is if it is safe enough to use `returnDataSizes`? Does it make sense to use `uintStorage`?

patitonar: Perhaps using the existing EternalStorage, in this case `uintStorage` could be a safer idea

contracts/upgradeability/ClassicEternalStorageProxy.sol
```diff
@@ -0,0 +1,35 @@
...
+        uint256 len = getSize(sig);
+        assembly {
+            calldatacopy(0x0, 0x0, calldatasize)
+            let result := delegatecall(gas, _impl, 0x0, calldatasize, 0, len)
```
akolotov: I have noticed a mention of a bug related to `delegatecall` here: https://ropsten.etherscan.io/solcbuginfo (DelegateCallReturnValue). I also found the following post: https://github.com/paritytech/parity-ethereum/issues/6201#issuecomment-319294744.

So, since we are using zero pointer to get the return value I would make a test when we return some zero value from the method that is called by `delegatecall` to make sure that `result` will not be zero.
If it happens, probably we need to pass a free memory pointer instead of zero as the 5th parameter. So, I will leave the post https://ethereum.stackexchange.com/a/28457 here as reference that could help.

contracts/upgradeability/ClassicEternalStorageProxy.sol
patitonar: I like the rest of the approach, the only thing is that I think we should initialize the variables on `initialize` method of `HomeBridge`
```
returnDataSizes[bytes4(sha3("signature(bytes32,uint256)"))] = 132;
returnDataSizes[bytes4(sha3("message(bytes32)"))] = 210;
```
Otherwise they won't be available as part of the state on the proxy contract

akolotov: Correct. Thanks for pointing this out!

contracts/upgradeability/ClassicEternalStorageProxy.sol
patitonar: Done

deploy/src/erc_to_erc/home.js
patitonar: Thank you! Removed

contracts/upgradeability/ClassicEternalStorageProxy.sol
patitonar: Done

contracts/upgradeability/ClassicEternalStorageProxy.sol
patitonar: Implemented the changes here: https://github.com/poanetwork/poa-bridge-contracts/pull/161/commits/a072da6e89a00979c052c1d31aab40d9d1657e45

contracts/upgradeability/ClassicEternalStorageProxy.sol
```diff
@@ -4,32 +4,28 @@ import "./EternalStorage.sol";
 import "./OwnedUpgradeabilityProxy.sol";
 
 
-contract ClassicEternalStorageProxy is OwnedUpgradeabilityProxy, EternalStorage {
+contract ClassicEternalStorageProxy is EternalStorage, OwnedUpgradeabilityProxy {
```
patitonar: I have to re-order the inheritance of the contracts because we weren't able to read the correct EternalStorage values from the proxy contract. For example on this line:
```
uint256 ret = uintStorage[keccak256("dataSizes", sig)];
``` 
The order is important because of how slot of memory are assigned. Before, proxy storage was assigning:
[0] owner of proxy
[1] version
[2] implementation
[3...] all eternal storage mappings

While on implementation contract it was assigning all eternal storage mappings on firsts slots. So when methods calls were delegated to implementation contracts there was no issues on reading and writing the values, but when trying to read o write values from proxy contracts it will be reading/writing on differents slots.

Here is an example on how EternalStorageProxy is defined on zeppelin repo: https://github.com/zeppelinos/labs/blob/master/upgradeability_using_eternal_storage/contracts/EternalStorageProxy.sol

Here is detailed explanation of this https://github.com/zeppelinos/labs/tree/master/upgradeability_using_eternal_storage

Does this makes sense to you?

contracts/upgradeable_contracts/BaseFeeManager.sol
patitonar: Created https://github.com/poanetwork/poa-bridge-contracts/issues/163

contracts/upgradeability/ClassicEternalStorageProxy.sol
patitonar: I just tested adding two methods that returns `false` and `0`. Called them through `delegatecall`, on both cases the transaction call wasn't reverted and values were returned correctly

akolotov:
Let's rename the branch to `develop-for-classic` and close the PR without merging to `support-null-status-deploy-script-0.4.19`.

patitonar:
Created and pushed `develop-for-classic` based on `contracts-version-0.4.19`

PR #162, opened 2019-03-18
[WIP]Vdc adrian

JuliusBrain:


akolotov:
@JuliusBrain thanks for PR. Could you share your thoughts why we need to remove the block reward functionality from the official version of the bridge contracts used for existing bridges like DAI-xDAI?

JuliusBrain:
I was just testing something and I thought I was on my private repo, didn't
mean to make a PR against an original repository of poa-bridge :)

I'm looking at ways of making ERC-TO-NATIVE work on sidechains which do not
support BlockRewardContract functionality.

Best Regards,
Adrian Hetman

akolotov:
@JuliusBrain thanks for clarification! Great to see other projects adapting the bridge for their needs! If you need any help you can always ask it through Gitter.

PR #165, opened 2019-04-01
Avoid calling blockReward if zero value on foreign transfer on erc-to-native

patitonar:


contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
```diff
@@ -81,10 +81,12 @@ contract HomeBridgeErcToNative is EternalStorage, BasicBridge, BasicHomeBridge,
     }
 
     function onExecuteAffirmation(address _recipient, uint256 _value) internal returns(bool) {
-        setTotalExecutedPerDay(getCurrentDay(), totalExecutedPerDay(getCurrentDay()).add(_value));
-        IBlockReward blockReward = blockRewardContract();
-        require(blockReward != address(0));
-        blockReward.addExtraReceiver(_value, _recipient);
+        if (_value > 0) {
```
akolotov: consider to move this check in the `executeAffirmation` method of the `BasicHomeBridge` contract. For me it is more logical to locate it there since it will not reveal the case why `executeAffirmation` is not called. Moreover I think it is good to have this logic common for all bridges rather than only `native-to-erc20` mode.

patitonar: Makes sense. Done

patitonar:
Gist with changes: https://gist.github.com/patitonar/88e6d2a6b2253056ace9ec662b7465f6

PR #167, opened 2019-04-11
Add validatorList on BaseBridgeValidators

patitonar:
See this comment https://github.com/poanetwork/bridge-ui/pull/184#discussion_r274128161


contracts/upgradeable_contracts/BaseBridgeValidators.sol
```diff
@@ -30,6 +30,25 @@ contract BaseBridgeValidators is EternalStorage, Ownable {
         return (2, 1, 0);
```
akolotov: Should we increase the version here?

patitonar: Yes, Thanks! Updated here https://github.com/poanetwork/poa-bridge-contracts/pull/167/commits/7581f1241a8152fb03c6a8be3569d93005b2b664

PR #169, opened 2019-04-15
Update RewardableValidators to be compatible with BridgeValidators  

### LINKED ISSUE #168: RewardableValidators event definitions causes additional complexity
See this thread https://github.com/poanetwork/bridge-ui/pull/184#discussion_r274034802

patitonar:
Closes #168

### DIFF
--- a/contracts/upgradeable_contracts/RewardableValidators.sol
+++ b/contracts/upgradeable_contracts/RewardableValidators.sol
@@ -5,9 +5,6 @@ import "./BaseBridgeValidators.sol";
 
 contract RewardableValidators is BaseBridgeValidators {
 
-    event ValidatorAdded (address indexed validator);
-    event ValidatorRemoved (address indexed validator);
-
     function initialize(
         uint256 _requiredSignatures,
         address[] _initialValidators,
--- a/contracts/upgradeable_contracts/RewardableValidators.sol
+++ b/contracts/upgradeable_contracts/RewardableValidators.sol
@@ -5,7 +5,7 @@ import "./BaseBridgeValidators.sol";
 
 contract RewardableValidators is BaseBridgeValidators {
 
-    event ValidatorAdded (address indexed validator, address reward);
+    event ValidatorAdded (address indexed validator);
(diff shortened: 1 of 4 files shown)


contracts/upgradeable_contracts/RewardableValidators.sol
```diff
@@ -54,11 +54,11 @@ contract RewardableValidators is BaseBridgeValidators {
         return isInitialized();
     }
 
-    function addValidator(address _validator, address _reward) external onlyOwner {
+    function addRewardableValidator(address _validator, address _reward) external onlyOwner {
```
akolotov: We need to redefine `addValidator` in this contract as so it cause `revert()` if it is called. Such redefinition will keep data consistency if it is called occasionally.

patitonar: I'm not sure I understood this correctly. `addValidator(address _validator)` method is defined on `BridgeValidators` contract, not `BaseBridgeValidators`.  So the only method to add a validator on `RewardableValidators` contract is `addRewardableValidator(address _validator, address _reward)`

akolotov: Yes, you are correct! sorry for misunderstanding.

contracts/upgradeable_contracts/RewardableValidators.sol
```diff
@@ -5,7 +5,7 @@ import "./BaseBridgeValidators.sol";
 
 contract RewardableValidators is BaseBridgeValidators {
 
-    event ValidatorAdded (address indexed validator, address reward);
+    event ValidatorAdded (address indexed validator);
```
akolotov: Do you think it make sense to move declarations of both events to the BaseBridgeValidators contract?

patitonar: Good catch! Yes, I think it makes sense. Done https://github.com/poanetwork/poa-bridge-contracts/pull/169/commits/9be3f38bdf2fa584a0b1367154abf33a6757ad99

PR #177, opened 2019-04-29
Add ERC677 support on foreign bridge in erc20-to-erc20 bridge mode

### LINKED ISSUE #175: ERC677 support on the Foreign side in the erc20-to-erc20 bridge mode
Current implementation of the `erc20-to-erc20` bridge mode assumes that pure ERC20 token is used on the Foreign side. So, when the tokens are relayed to the Home side, the method `transfer` is invoked by the user and the relay request is handled by the TokenBridge oracle without involvement the bridge contract on the Foreign side. It is inconvenient due to following:
- there is no way to revert the request if it exceeds the limits.
- there is no way to halt the bridge operations

Historically, such kind of functionality was suggested for already deployed ERC20 tokens to enhance them through the bridging ability and described shortcomings were considered as insignificant in comparison with the value such solution could bring. But for there are the cases for the bridge when it is going to be used with just deployed tokens. In this case the token could be extended by ERC677 standard in order to address issues described above. 

That's why the bridge contracts working in `erc20-to-erc20` mode could be improved to operate with ERC677 tokens. In order to achieve this the corresponding foreign bridge contract should be extended with support of `onTokenTransfer` method which will raise the `UserRequestForAffirmation` event.

patitonar:
Closes #175


contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
```diff
@@ -22,6 +19,29 @@ contract ForeignBridgeErcToErc is BasicBridge, BasicForeignBridge {
...
+        );
+        return isInitialized();
+    }
+
+    function _initialize(
+        address _validatorContract,
+        address _erc20token,
+        uint256 _requiredBlockConfirmations,
+        uint256 _gasPrice,
+        uint256 _maxPerTx,
```
akolotov: Do you have any idea why we left `_maxPerTx` here? If we look at `validations` in `loadEnv.js`, there is no check for this parameter in case of `erc20-to-erc20` bridge that's why I think it was left here by mistake.

deploy/src/erc_to_erc/preDeploy.js
```diff
@@ -0,0 +1,18 @@
+const { web3Foreign } = require('../web3')
+const { ERC20_TOKEN_ADDRESS, ERC20_EXTENDED_BY_ERC677 } = require('../loadEnv')
+const { abi } = require('../../../build/contracts/ERC677BridgeToken.json')
+
+async function preDeploy() {
```
akolotov: Is it possible to have this check in `validations` of the `src/loadEnv.js` file?

contracts/upgradeable_contracts/ERC20ExtendedBridge.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity 0.4.24;
+
+
+import "./BasicBridge.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/ERC20Basic.sol";
```
akolotov: We have already very a similar interface `ERC677Bridge`. But in the new logic it makes sense to call it `ERC677BridgeForBurnableMintableToken`, doesn't it?

Below is my suggestion how to re-organize the hierarchy of the contracts in this case.

*ERC677Bridge.sol*
```solidity
pragma solidity 0.4.24;


import "./BasicBridge.sol";
import "../ERC677.sol";

contract ERC677Bridge is BasicBridge {
    function erc677token() public view returns(ERC677) {
        return ERC677(addressStorage[keccak256(abi.encodePacked("erc677token"))]);
    }

    function setErc677token(address _token) internal {
        require(_token != address(0) && isContract(_token));
        addressStorage[keccak256(abi.encodePacked("erc677token"))] = _token;
    }

    function onTokenTransfer(address _from, uint256 _value, bytes /*_data*/) external returns(bool) {
    	ERC677 token = erc677token();
        require(msg.sender == address(token));
        require(withinLimit(_value));
        setTotalSpentPerDay(getCurrentDay(), totalSpentPerDay(getCurrentDay()).add(_value));
        bridgeSpecificActionsOnTokenTransfer(token, _from, _value);
        return true;
    }

    function bridgeSpecificActionsOnTokenTransfer(ERC677 _token, address _from, uint256 _value) internal {
        fireEventOnTokenTransfer(_from, _value);
    }

    function fireEventOnTokenTransfer(address /*_from */, uint256 /* _value */) internal;

}
```

*ERC677BridgeForBurnableMintableToken.sol*
```solidity
pragma solidity 0.4.24;


import "./BasicBridge.sol";
import "./ERC677Bridge.sol";

contract ERC677BridgeForBurnableMintableToken is ERC677Bridge {
    function bridgeSpecificActionsOnTokenTransfer(IBurnableMintableERC677Token _token, address _from, uint256 _value) internal {
        erc677token().burn(_value);
        fireEventOnTokenTransfer(_from, _value);
    }

    function fireEventOnTokenTransfer(address /*_from */, uint256 /* _value */) internal;

}
```

* *erc20_to_erc20/HomeBridgeErcToErc.sol* will inherit *ERC677BridgeForBurnableMintableToken* rather than *ERC677Bridge*
* *native_to_erc20/ForeignBridgeNativeToErc.sol* will inherit *ERC677BridgeForBurnableMintableToken* rather than *ERC677Bridge*
* *erc20_to_erc20/ForeignBridgeErc677ToErc677.sol*
```solidity
pragma solidity 0.4.24;

import "./ForeignBridgeErcToErc.sol";
import "../ERC677Bridge.sol";

contract ForeignBridgeErc677ToErc677 is ERC677Bridge, ForeignBridgeErcToErc {

    event UserRequestForAffirmation(address recipient, uint256 value);

    function initialize(
        address _validatorContract,
        address _erc20token,
        uint256 _gasPrice,
        uint256 _dailyLimit,
        uint256 _maxPerTx,
        uint256 _minPerTx,
        uint256 _homeDailyLimit,
        uint256 _homeMaxPerTx,
        address _owner
    ) public returns(bool) {
        require(_minPerTx > 0 && _maxPerTx > _minPerTx && _dailyLimit > _maxPerTx);

        _initialize(
            _validatorContract,
            _erc20token,
            _requiredBlockConfirmations,
            _gasPrice,
            _homeDailyLimit,
            _homeMaxPerTx,
            _owner
        );

        uintStorage[keccak256(abi.encodePacked("dailyLimit"))] = _dailyLimit;
        uintStorage[keccak256(abi.encodePacked("maxPerTx"))] = _maxPerTx;
        uintStorage[keccak256(abi.encodePacked("minPerTx"))] = _minPerTx;

        return isInitialized();
    }

    function erc20token() public view returns(ERC20Basic) {
        return erc677token();
    }

    function setErc20token(address _token) private {
        setErc677token(_token);
    }

    function fireEventOnTokenTransfer(address _from, uint256 _value) internal {
        emit UserRequestForAffirmation(_from, _value);
    }
}
```

contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeExtendedErcToErc.sol
```diff
@@ -0,0 +1,57 @@
...
+        address /* _validatorContract */,
+        address /* _erc20token */,
+        uint256 /* _requiredBlockConfirmations */,
+        uint256 /* _gasPrice */,
+        uint256 /* _maxPerTx */,
+        uint256 /* _homeDailyLimit */,
+        uint256 /* _homeMaxPerTx */,
+        address /* _owner */
+    ) public returns(bool) {
+        revert();
```
akolotov: there is no need to make this if the method `extendedInitialize` is called `initialize` instead. What do you think?

contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
patitonar: I think I found the reason. It was introduced as part of this issue for `erc-to-native` and `erc-to-erc` modes https://github.com/poanetwork/poa-bridge-contracts/issues/110. So if it is correct that `_maxPerTx` should be there, then we need to add the validation on `loadEnv.js`

akolotov: Oh! Thanks for finding this

patitonar:
> What do you think is better:
make this PR against the 119-Epic-rewards-for-bridge-validators branch?
make the PR against develop and merge it later into 119-Epic-rewards-for-bridge-validators since this feature will be needed in POSDAO chain together with other changes related to introduction of the validators reward?

My bad, I should have asked about the base branch for this. I used the `develop` branch as base because there wasn't any needed changes from `119-Epic-rewards-for-bridge-validators` for this feature. I think we can merge this into `develop` and then merge `develop` into `119-Epic-rewards-for-bridge-validators` branch

akolotov:
> we can merge this into `develop` and then merge `develop` into `119-Epic-rewards-for-bridge-validators` branch

OK

deploy/src/erc_to_erc/preDeploy.js
patitonar: It is possible, but it will increase complexity on the usage of `loadEnv.js`. Right now this file reads the env variables and perform the validations synchronously. If we perform requests as part of these validations, the file will export a promise that will return the env variables instead of just returning the env variables. So on every file it is used we will have to handle this and on some utils files such us `src/web3.js` the exported values will also become a Promise because it need the env variables to generate and export objects as `web3Home` and `web3Foreign`. I think it would be a good idea to keep `loadEnv.js` as it is. If you want to avoid adding this new `preDeploy` file, maybe we can perform the check inside the foreign deploy and deploy first foreign instead of home on this bridge mode

contracts/upgradeable_contracts/erc20_to_erc20/ForeignBridgeErcToErc.sol
patitonar: Added validation for `maxPerTx` for all bridge modes https://github.com/poanetwork/poa-bridge-contracts/pull/177/commits/17b907a821c4a4f6f3f9589597f1b5c581f577d9

contracts/upgradeable_contracts/ERC20ExtendedBridge.sol
patitonar: Thanks for the detailed explanation! Here I applied the changes https://github.com/poanetwork/poa-bridge-contracts/pull/177/commits/5d54ffb20bda80cf64a59529569bec4f1ef60ba5

deploy/src/erc_to_erc/preDeploy.js
akolotov: Thanks for explanation. Let's leave `preDeploy` as you implemented it already.

contracts/upgradeable_contracts/ERC677Bridge.sol
```diff
@@ -15,16 +16,17 @@ contract ERC677Bridge is BasicBridge {
     }
 
     function onTokenTransfer(address _from, uint256 _value, bytes /*_data*/) external returns(bool) {
-        require(msg.sender == address(erc677token()));
+        IBurnableMintableERC677Token token = erc677token();
```
akolotov: Since this contract is assumed to be common for other contracts covering the functionality to bridge erc677 tokens, it should work with the `ERC677` instance rather than `BurnableMintableERC677Token`.

patitonar: For method `erc677token()` wee need to return `IBurnableMintableERC677Token` because `HomeBridgeErcToErc` and `ForeignBridgeNativeToErc` calls `erc677token().mint()` which is not defined on `ERC677`

If we use `ERC677` on `onTokenTransfer` method, also `ERC677` should be used on `bridgeSpecificActionsOnTokenTransfer(ERC677 _token, address _from, uint256 _value)`

But then on `ERC677BridgeForBurnableMintableToken` contract, method `bridgeSpecificActionsOnTokenTransfer(IBurnableMintableERC677Token _token, address _from, uint256 _value)`
is not overriding the method defined on `ERC677Bridge` but interpreted as a different method which cause incorrect behaviour.

Perhaps an option could be to instead of returning/receiving `ERC677` or `IBurnableMintableERC677Token` we can return and receive parameters as `address` and then convert to the proper contract to call the needed methods

akolotov: Could you consider such kind of changes as a suggestion to address your comments?
```diff
diff --git a/contracts/upgradeable_contracts/ERC677Bridge.sol b/contracts/upgradeable_contracts/ERC677Bridge.sol
index 3d99026..37777b3 100644
--- a/contracts/upgradeable_contracts/ERC677Bridge.sol
+++ b/contracts/upgradeable_contracts/ERC677Bridge.sol
@@ -5,7 +5,6 @@ import "./BasicBridge.sol";
 import "../ERC677.sol";
-import "../IBurnableMintableERC677Token.sol";
 
 contract ERC677Bridge is BasicBridge {
-    function erc677token() public view returns(IBurnableMintableERC677Token) {
-        return IBurnableMintableERC677Token(addressStorage[keccak256(abi.encodePacked("erc677token"))]);
+    function erc677token() public view returns(ERC677) {
+        return ERC677(addressStorage[keccak256(abi.encodePacked("erc677token"))]);
     }
@@ -18,3 +17,3 @@ contract ERC677Bridge is BasicBridge {
     function onTokenTransfer(address _from, uint256 _value, bytes /*_data*/) external returns(bool) {
-        IBurnableMintableERC677Token token = erc677token();
+        ERC677 token = erc677token();
         require(msg.sender == address(token));
@@ -26,3 +25,3 @@ contract ERC677Bridge is BasicBridge {
 
-    function bridgeSpecificActionsOnTokenTransfer(IBurnableMintableERC677Token _token, address _from, uint256 _value) internal {
+    function bridgeSpecificActionsOnTokenTransfer(ERC677 _token, address _from, uint256 _value) internal {
         fireEventOnTokenTransfer(_from, _value);
diff --git a/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol b/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
index f93e841..3483626 100644
--- a/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
+++ b/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
@@ -8,4 +8,4 @@ import "../IBurnableMintableERC677Token.sol";
 contract ERC677BridgeForBurnableMintableToken is ERC677Bridge {
-    function bridgeSpecificActionsOnTokenTransfer(IBurnableMintableERC677Token _token, address _from, uint256 _value) internal {
-        _token.burn(_value);
+    function bridgeSpecificActionsOnTokenTransfer(ERC677 _token, address _from, uint256 _value) internal {
+        IBurnableMintableERC677Token(_token).burn(_value);
         fireEventOnTokenTransfer(_from, _value);
diff --git a/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol b/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
index 6fb9992..af2b3db 100644
--- a/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
+++ b/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
@@ -63,3 +63,3 @@ contract HomeBridgeErcToErc is ERC677Receiver, EternalStorage, BasicBridge, Basi
         setTotalExecutedPerDay(getCurrentDay(), totalExecutedPerDay(getCurrentDay()).add(_value));
-        return erc677token().mint(_recipient, _value);
+        return IBurnableMintableERC677Token(erc677token()).mint(_recipient, _value);
     }
diff --git a/contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol b/contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
index 7bdd35e..70f1d0c 100644
--- a/contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
+++ b/contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
@@ -53,3 +53,3 @@ contract ForeignBridgeNativeToErc is ERC677Receiver, BasicBridge, BasicForeignBr
     function claimTokensFromErc677(address _token, address _to) external onlyIfOwnerOfProxy {
-        erc677token().claimTokens(_token, _to);
+        IBurnableMintableERC677Token(erc677token()).claimTokens(_token, _to);
     }
@@ -58,3 +58,3 @@ contract ForeignBridgeNativeToErc is ERC677Receiver, BasicBridge, BasicForeignBr
         setTotalExecutedPerDay(getCurrentDay(), totalExecutedPerDay(getCurrentDay()).add(_amount));
-        return erc677token().mint(_recipient, _amount);
+        return IBurnableMintableERC677Token(erc677token()).mint(_recipient, _amount);
     }
```
If they are OK I could commit them to the branch `175-ERC677-support-foreign-erc20-to-erc20` and you can double check after that that everything is ready for merge

patitonar: Yes, looks good to me! Thanks!

akolotov: I pushed the changes. Please double check and test them.

patitonar: Changes looks good and all tests are passing so I think we can merge this

PR #181, opened 2019-05-09
Support multiple evm versions

### LINKED ISSUE #163: Refactor distributeFeeProportionally method on BaseFeeManager
We can reduce the number of iterations to get the list of validators and distribute the fees. See this [comment](https://github.com/poanetwork/poa-bridge-contracts/pull/161#discussion_r266012183)

### LINKED ISSUE #180: Support of different EVM version to compile contracts
As per work to adapt the bridge contracts to work in the Ethereum Classic environment the separate branch `develop-for-classic` was created. The following were done there:
- to downgrade of the compiler version in the pragma to eliminate usage of opcodes introduced as part of  the byzantium fork;
- to fix of the issues related to the compiler version downgrade;
- get rid of usage `returndatacopy` and `returndatasize` opcodes in cases where `delegatecall` could call a method returning data.

Improvements and bug fixing in both branches `develop` and `develop-for-classic` require double work since these branches have lots of common code so it is necessary to develop an approach to have sources for different version of EVMs in one branch.

patitonar:
This PR includes the following changes:
- [X] Propagate the changes from the develop-for-classic (Proxy contract and this fix #163)
- [X] Upgrade Truffle to version 5
- [X] Build for differents evm-versions on `npm run compile`
- [X] Update deploy scripts to support using the correct contract builds using env variables

Also:
- [X] Removed migrations files from `migrations` folder since there was no use case. We have to keep the empty folder otherwise truffle throws an error when trying to run the unit tests
- [X] Added a specific script to run the unit test with the gas report which is much slower so it doesn't affect the development process. 
- [X] Fix unit tests. All unit tests need to be fixed because of the new truffle version, which changed the web3 version, the big number library and the api to interact with the contracts on the tests. Here is more information about the breaking changes https://truffleframework.com/blog/truffle-v5-has-arrived
- [X] Added linter for unit tests

Closes #180 
Closes #163


patitonar:
I run a test in a environment equal to etc-wetc bridge
- Deployed contracts
- Used bridge-ui to make transfers on both sides. All information was displayed correctly.
- Fee distribution worked correctly
- Fee events emitted correctly
- Changed fee manager contract address
- Upgraded Bridge Contract to a new implementation
- Updated fee values

> Will it be possible to get the list of validators from the contracts deployed in the claissic environment with these changes?

The method `validatorList()` still returns an empty array on the classic environment, so in bridge-ui the list of validators is build using the events

akolotov:
Thanks! I will continue the review.

PR #191, opened 2019-06-14
Add 0x/sol-coverage code coverage

patitonar:
The previous library we were using for code coverage had some issues, I fixed some of them but there is one that does not support using Truffle 5, there are some forks of the projects that solves this but none of them works with 0.4.x contracts. (see related issue https://github.com/sc-forks/solidity-coverage/issues/309)

Added `0x/sol-coverage` which seems to be working with our current configuration, also added `coveralls` to display coverage information on each PR.


package.json
```diff
@@ -10,19 +10,26 @@
     "flatten": "bash flatten.sh",
     "lint:js": "eslint .",
     "lint:js:fix": "eslint . --fix",
-    "watch-tests": "./node_modules/.bin/nodemon ./node_modules/.bin/truffle test --network test"
+    "watch-tests": "./node_modules/.bin/nodemon ./node_modules/.bin/truffle test --network test",
+    "coverage": "SOLIDITY_COVERAGE=true scripts/test.sh"
```
akolotov: tried to run the coverage testing in Docker environment and got lots of strange messages and an error at the end.
Please check: https://gist.github.com/akolotov/b63d983ccc10fd597521d2f4bb6abc9d

patitonar: Fixed the errors https://github.com/poanetwork/poa-bridge-contracts/pull/191/commits/ff8d177f2b16c4801c4b97cd6a1974796073b56c. Also coverage build is working now https://travis-ci.org/poanetwork/poa-bridge-contracts/jobs/547244075. You can see the coveralls result on this PR on below  the comments on check displayed with travis `coverage/coveralls — First build on add-code-coverage at 89.185%`.

I was able to run the coverage with docker too. Please try re-building and running the script.

There are some warnings `Unable to find matching bytecode for ..` those seems to be triggered when using mock contracts `ERC677BridgeTokenRewardableMock`, `RevertFallback`, etc, those contracts are ignored for the coverage report so I think it is OK to ignore them

akolotov: Thanks! I have run the test in the docker and got the error: https://gist.github.com/akolotov/eeaefe718cff3b6f0fe4578003f4ceab. Then I ran it again and the issue did not happen. I will open a separate issue for this.

PR #197, opened 2019-06-19
Update setBlockRewardContract method validations

### LINKED ISSUE #194: Remove use of `bridgesAllowedLength` from `HomeBridgeErcToNative.setBlockRewardContract` function
Since the `BlockReward` contract in the [`posdao-contracts`](https://github.com/poanetwork/posdao-contracts) repo doesn't have `bridgesAllowed` and `bridgesAllowedLength` getters anymore, we should remove the use of `bridgesAllowedLength` from the [`HomeBridgeErcToNative.setBlockRewardContract`](https://github.com/poanetwork/poa-bridge-contracts/blob/bbb97a63c900f03a902d0e82358abac3b294e4d9/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol#L119) function.

patitonar:
Closes #194

### DIFF
--- a/contracts/upgradeable_contracts/erc20_to_erc20/FeeManagerErcToErcPOSDAO.sol
+++ b/contracts/upgradeable_contracts/erc20_to_erc20/FeeManagerErcToErcPOSDAO.sol
@@ -13,7 +13,14 @@ contract FeeManagerErcToErcPOSDAO is BlockRewardFeeManager {
     }
 
     function setBlockRewardContract(address _blockReward) external {
-        require(_blockReward != address(0) && isContract(_blockReward) && (IBlockReward(_blockReward).bridgesAllowedLength() != 0));
+        require(_blockReward != address(0) && isContract(_blockReward));
+        bool isBlockRewardContract = false;
+        if (_blockReward.call(abi.encodeWithSignature("blockRewardContractId()"))) {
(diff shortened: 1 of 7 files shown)


contracts/upgradeable_contracts/erc20_to_erc20/FeeManagerErcToErcPOSDAO.sol
```diff
@@ -13,7 +13,14 @@ contract FeeManagerErcToErcPOSDAO is BlockRewardFeeManager {
     }
 
     function setBlockRewardContract(address _blockReward) external {
-        require(_blockReward != address(0) && isContract(_blockReward) && (IBlockReward(_blockReward).bridgesAllowedLength() != 0));
+        require(_blockReward != address(0) && isContract(_blockReward));
+        bool isBlockRewardContract = false;
```
akolotov: Since it could be a questionable place in the future let's add a commit like "Before store the contract we need to make sure that it is the block reward contract in actual fact, call a specific method from the contract that should return a specific value"

patitonar: Added comments https://github.com/poanetwork/poa-bridge-contracts/pull/197/commits/f0382fb915c5122b9dd15962cf8364d6789fda8f

akolotov: thanks

PR #198, opened 2019-06-19
Remove proxyOwner method from proxy contract

### LINKED ISSUE #195: Simplify the logic of proxy contract by removing odd role proxyOwner
The current version of the contracts assumes that the roles `proxyOwner` and `upgradeabilityOwner` responsible for the same functional:

https://github.com/poanetwork/poa-bridge-contracts/blob/bbb97a63c900f03a902d0e82358abac3b294e4d9/contracts/upgradeability/OwnedUpgradeabilityProxy.sol#L38-L50

The recommendation we got during the security audit is to simplify the contracts and remove this duplication. 

It is necessary to look at other token bridge components to see where this functionality is being used and if it is safe to remove the `proxyOwner` role as so `upgradeabilityOwner` remain.

patitonar:
Closes #195


contracts/upgradeability/OwnedUpgradeabilityProxy.sol
```diff
@@ -27,25 +27,17 @@ contract OwnedUpgradeabilityProxy is UpgradeabilityOwnerStorage, UpgradeabilityP
     * @dev Throws if called by any account other than the owner.
     */
     modifier onlyProxyOwner() {
```
akolotov: It makes sense to rename `onlyProxyOwner` as well

contracts/upgradeable_contracts/OwnedUpgradeability.sol
```diff
@@ -6,7 +6,7 @@ import "../IOwnedUpgradeabilityProxy.sol";
 contract OwnedUpgradeability {
 
     function upgradeabilityAdmin() public view returns (address) {
```
akolotov: Do we need `upgradeabilityAdmin` in this case at all? What impact for other components could be if we leave `upgradeabilityOwner` only (assuming that it is public)?

patitonar: `upgradeabilityAdmin` is only used on `onlyIfOwnerOfProxy` on the contracts and no other bridge components uses it. If it is needed, other components can call `upgradeabilityOwner` directly.
In this case, I think we can remove `upgradeabilityAdmin` and redefine `onlyIfOwnerOfProxy` as:
```solidity
    modifier onlyIfOwnerOfProxy() {
        require(msg.sender == IOwnedUpgradeabilityProxy(this).upgradeabilityOwner());
        _;
    }
```

patitonar: Applied changes https://github.com/poanetwork/poa-bridge-contracts/pull/198/commits/41f9ba8cf67e04817a7d59f7a8a865c83f30d1a8

contracts/upgradeability/OwnedUpgradeabilityProxy.sol
patitonar: Done https://github.com/poanetwork/poa-bridge-contracts/pull/198/commits/f3c94b7841be3d0a4a694b30932cf4abe305f6d1

patitonar:
>line 10 (address private _upgradeabilityOwner;), why do you use private visibility modifier instead of internal? (edited)
Do you think it make sense to make this one-line-change under this PR?

Sure, it is related. Added the change on this commit https://github.com/poanetwork/poa-bridge-contracts/pull/198/commits/e73d2aa4e01a6f1c7b3d6f57f798dff0643b7e38

contracts/upgradeable_contracts/OwnedUpgradeability.sol
```diff
@@ -4,14 +4,9 @@ import "../IOwnedUpgradeabilityProxy.sol";
 
 
 contract OwnedUpgradeability {
-
-    function upgradeabilityAdmin() public view returns (address) {
-        return IOwnedUpgradeabilityProxy(this).proxyOwner();
-    }
-
     // Avoid using onlyProxyOwner name to prevent issues with implementation from proxy contract
     modifier onlyIfOwnerOfProxy() {
```
akolotov: `onlyIfOwnerOfProxy` looks differently comparing to changes were made and does not reflect actual meaning.
Consider to rename it to `onlyIfUpgradeabilityOwner`

contracts/upgradeable_contracts/erc20_to_erc20/BasicForeignBridgeErcToErc.sol
```diff
@@ -39,7 +39,7 @@ contract BasicForeignBridgeErcToErc is BasicBridge, BasicForeignBridge {
         return bytes4(keccak256(abi.encodePacked("erc-to-erc-core")));
     }
 
-    function claimTokens(address _token, address _to) public onlyIfOwnerOfProxy {
+    function claimTokens(address _token, address _to) public onlyIfUpgradeabilityOwner {
```
akolotov: Do we still need this renaming even if we know that it will be removed under https://github.com/poanetwork/poa-bridge-contracts/pull/203?

patitonar: You are right, probably it just make sense to remove it for `BasicForeignBridgeErcToErc.sol` 
 and `ForeignBridgeErcToNative.sol`

PR #204, opened 2019-06-26
Improve signatures validation on Message library

patitonar:


contracts/libraries/Message.sol
```diff
@@ -98,6 +97,7 @@ library Message {
         bytes32[] _ss,
         IBridgeValidators _validatorContract) internal view {
         require(isMessageValid(_message));
+        require(_vs.length == _rs.length && _rs.length == _ss.length);
```
akolotov: Do you think it is really needed? If validator is not misbehaving it will always true (spend extra gas). Whereas if the validator did something incorrectly the transaction will be reverted in the line 107 anyway.

patitonar: I think you're right. The validator won't be able to misbehave anyway, so this will only hurt with extra gas on valid transactions. I think we can close this PR without merging.

akolotov: This is good that you prepared tests to check this functionality - it makes sense to have them. So, we will close the PR but rollback the changes in the discussed line of the code.

Probably it will be good idea also to add a comment prior to this line which says something like 'It is not necessary to check that arrays have the same length since it will be handled during attempt to access to the corresponding elements in the loop and the call will be reverted. It will save gas for the rational validators actions and still be be safe enough from security point of view'

patitonar: Thanks! Removed the check and added a comment on the method https://github.com/poanetwork/poa-bridge-contracts/pull/204/commits/151288af8c5d59afbc092e79155a5a3ffbdf43f0

contracts/libraries/Message.sol
```diff
@@ -90,14 +90,16 @@ library Message {
         return keccak256(abi.encodePacked(prefix, msgLength, message));
     }
 
+    // It is not necessary to check that arrays have the same length since it will be handled
```
akolotov: Consider to move the comment just before the line containing `require(_vs.length >= requiredSignatures);`

patitonar: Fixed https://github.com/poanetwork/poa-bridge-contracts/pull/204/commits/f1fc07bb5ca88eb137023c3b1446739b1f108bf5

PR #209, opened 2019-06-26
Add fee value check

### LINKED ISSUE #205: No check that the fee percentage is less than 100%
During the security audit it was found that operations to set the fee does not check that the fee is less than `1`:

https://github.com/poanetwork/poa-bridge-contracts/blob/bbb97a63c900f03a902d0e82358abac3b294e4d9/contracts/upgradeable_contracts/BaseFeeManager.sol#L24-L27

https://github.com/poanetwork/poa-bridge-contracts/blob/bbb97a63c900f03a902d0e82358abac3b294e4d9/contracts/upgradeable_contracts/BaseFeeManager.sol#L33-L36

patitonar:
Closes #205

### DIFF
--- a/contracts/upgradeable_contracts/BaseFeeManager.sol
+++ b/contracts/upgradeable_contracts/BaseFeeManager.sol
@@ -12,6 +12,7 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
     event HomeFeeUpdated(uint256 fee);
     event ForeignFeeUpdated(uint256 fee);
 
+    // This is not a real fee value but a relative value used to calculate the fee percentage
     uint256 internal constant MAX_FEE = 1 ether;
 
     function calculateFee(uint256 _value, bool _recover, bytes32 _feeType) public view returns(uint256) {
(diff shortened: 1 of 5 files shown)


contracts/upgradeable_contracts/BaseFeeManager.sol
```diff
@@ -12,16 +12,22 @@ contract BaseFeeManager is EternalStorage, FeeTypes {
     event HomeFeeUpdated(uint256 fee);
     event ForeignFeeUpdated(uint256 fee);
 
+    uint256 internal constant MAX_FEE = 1 ether;
```
akolotov: consider to add a description explaining that `1 ether` is not a fee the bridge can gather as a maximum but some relative value used to calculate the fee percentage.

patitonar: Thanks for suggestion, added a comment https://github.com/poanetwork/poa-bridge-contracts/pull/209/commits/53f2a1ceaaf2316573af12244c24e428d6d843d7

PR #213, opened 2019-06-27
Fix claimTokens to support all token transfers

### LINKED ISSUE #206: claimTokens fails if a token's method transfer does not return true
It was found during the security audit that `claimTokens()` function will not work with tokens, which do not return `true` on `transfer()` function calls.
https://github.com/poanetwork/poa-bridge-contracts/blob/bbb97a63c900f03a902d0e82358abac3b294e4d9/contracts/upgradeable_contracts/BasicBridge.sol#L139

Consider to use an approach suggested in the OpenZeppelin library (or even the proper version of the library itself) to address the issue:
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/c9f328ef66251db7fac7c704dd6c5523fc53b0ab/contracts/token/ERC20/SafeERC20.sol#L67-L73

patitonar:
Closes #206


contracts/libraries/Token.sol
```diff
@@ -0,0 +1,37 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-solidity/contracts/token/ERC20/ERC20Basic.sol";
+
+library Token {
+
+    function claimTokens(address _token, address _to) internal {
+        require(_to != address(0));
+        if (_token == address(0)) {
```
akolotov: this part is to claim native coins not tokens consider to have it in a separate internal method.

contracts/libraries/Token.sol
```diff
@@ -0,0 +1,37 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-solidity/contracts/token/ERC20/ERC20Basic.sol";
+
+library Token {
+
+    function claimTokens(address _token, address _to) internal {
+        require(_to != address(0));
```
akolotov: consider to have this as a modifier for original method `claimTokens`.

contracts/libraries/Token.sol
```diff
@@ -0,0 +1,37 @@
...
+        if (_token == address(0)) {
+            _to.transfer(address(this).balance);
```
akolotov: does it make sense to use here the approach with selfdestructor to secure the coins transfer?

patitonar:
@akolotov Thanks for the suggestions, applied the changes mentioned and also refactored a little the solution, please check https://github.com/poanetwork/poa-bridge-contracts/pull/213/commits/f87c1af1c3d814f43ccde54a46a1b639ed9f8921

PR #220, opened 2019-07-01
Update ERC677BridgeToken transferFrom to call contractFallback

### LINKED ISSUE #216: transfer and transferFrom behaves differently when bridge contract is receiver
The method `transfer` in `ERC677BridgeToken` has been extended to handle properly the case when it was used to transfer tokens to the bridge contract instead of calling `transferAndCall`.

https://github.com/poanetwork/poa-bridge-contracts/blob/e32ef6e1f2cc78a07fd96a321b426a64a2f87132/contracts/ERC677BridgeToken.sol#L57-L68

But the method `transferFrom` remains unchanged. So, behavior will be inconsistent for the similar functionality.

patitonar:
Closes #216

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -46,7 +46,7 @@ contract ERC677BridgeToken is
     }
 
     function getTokenInterfacesVersion() public pure returns(uint64 major, uint64 minor, uint64 patch) {
-        return (2, 0, 0);
+        return (2, 1, 0);
     }
 
     function superTransfer(address _to, uint256 _value) internal returns(bool)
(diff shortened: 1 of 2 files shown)


contracts/ERC677BridgeToken.sol
```diff
@@ -57,21 +57,28 @@ contract ERC677BridgeToken is
     function transfer(address _to, uint256 _value) public returns (bool)
     {
         require(superTransfer(_to, _value));
-        if (isContract(_to) && !contractFallback(_to, _value, new bytes(0))) {
+        callAfterTransfer(msg.sender, _to, _value);
+        return true;
+    }
+
+    function transferFrom(address _from, address _to, uint256 _value) public returns (bool) {
```
akolotov: Since we modify the core functionality of the contract consider to increase the version returned by `getTokenInterfacesVersion`.

patitonar: updated version https://github.com/poanetwork/poa-bridge-contracts/pull/220/commits/94c3e02364ea570d953054341e17707aef5230bd

PR #225, opened 2019-07-03
Remove duplicated checks of contracts

### LINKED ISSUE #217: duplicated checks of contracts in ERC677BridgeTokenRewardable
The method `isContract` is enough to check if operations are going to be made with an actual contract in the following methods:

https://github.com/poanetwork/poa-bridge-contracts/blob/e32ef6e1f2cc78a07fd96a321b426a64a2f87132/contracts/ERC677BridgeTokenRewardable.sol#L17-L25

The check `... != address(0)` is redundant here.

patitonar:
Closes #217 

Besides of duplicate checks on `ERC677BridgeTokenRewardable`, this pattern was also repeated on other contracts so I applied the changes to these contracts too.

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -24,7 +24,7 @@ contract ERC677BridgeToken is
     public DetailedERC20(_name, _symbol, _decimals) {}
 
     function setBridgeContract(address _bridgeContract) onlyOwner public {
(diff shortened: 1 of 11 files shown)


akolotov:
Asked auditors if it is OK (safe) from their side to optimize all places in the code as it is suggested in this PR.

PR #228, opened 2019-07-12
Improve checks of configuration parameters for contracts deployment

### LINKED ISSUE #185: check correctness of configuration parameters before the contracts deployment
At this moment a validation of the configuration parameters (`deploy/src/loadEnv.js`) allows to avoid very trivial mistakes in the configuration. But sometimes the bridge adopters meet the issues related with the contract internals (e.g. `FOREIGN_MAX_AMOUNT_PER_TX` cannot be equal to `FOREIGN_DAILY_LIMIT` in the `native-to-erc20` mode or necessity to point out a real contract as the ERC20 token on the foreign side in the `erc20-to-erc20` mode). In this case they get the initialization of the bridge contracts failed without detailed description of the problem and raise questions in the forum/gitter/slack etc. 

It is necessary to review the code of the bridge contracts and implement the corresponding checks on the deployment script side as so the adopters receive meaningful information about the contracts deployment/initialization failures.

patitonar:
Closes #185 

Refactored `loadEnv.js` so now:
- First we check that all required parameters are present and in the correct type
- Then we check that the values assigned to the parameters makes sense in order to comply with the requires of the contracts.

Added `preDeploy` checks on the deployment script, to verify that the contracts addresses provided on the parameters have code.

### DIFF
--- a/deploy/src/loadEnv.js
+++ b/deploy/src/loadEnv.js
@@ -154,7 +154,8 @@ if (BRIDGE_MODE === 'NATIVE_TO_ERC') {
   if (DEPLOY_REWARDABLE_TOKEN === 'true') {
     validations = {
       ...validations,
-      DPOS_STAKING_ADDRESS: addressValidator()
+      DPOS_STAKING_ADDRESS: addressValidator(),
+      BLOCK_REWARD_ADDRESS: addressValidator()
     }
   }
 }
--- a/deploy/src/loadEnv.js
+++ b/deploy/src/loadEnv.js
@@ -167,7 +167,6 @@ if (BRIDGE_MODE === 'ERC_TO_ERC') {
     BRIDGEABLE_TOKEN_SYMBOL: envalid.str(),
     BRIDGEABLE_TOKEN_DECIMALS: envalid.num(),
     DEPLOY_REWARDABLE_TOKEN: envalid.bool(),
-    BLOCK_REWARD_ADDRESS: addressValidator(),
     ERC20_EXTENDED_BY_ERC677: envalid.bool()
   }
 
@@ -181,7 +180,8 @@ if (BRIDGE_MODE === 'ERC_TO_ERC') {
   if (DEPLOY_REWARDABLE_TOKEN === 'true') {
     validations = {
       ...validations,
-      DPOS_STAKING_ADDRESS: addressValidator()
+      DPOS_STAKING_ADDRESS: addressValidator(),
+      BLOCK_REWARD_ADDRESS: addressValidator()
     }
   }
 }
--- a/deploy/src/loadEnv.js
+++ b/deploy/src/loadEnv.js
@@ -6,6 +6,9 @@ const { isAddress, toBN } = require('web3').utils
 const envalid = require('envalid')
 const { ZERO_ADDRESS, EVM_TYPES } = require('./constants')
 
+const homePrefix = 'HOME'
+const foreignPrefix = 'FOREIGN'
+
 // Validations and constants
 const evmVersions = [EVM_TYPES.BYZANTIUM, EVM_TYPES.SPURIOUSDRAGON]
 const validBridgeModes = ['NATIVE_TO_ERC', 'ERC_TO_ERC', 'ERC_TO_NATIVE']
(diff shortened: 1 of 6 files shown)


deploy/src/loadEnv.js
```diff
@@ -144,25 +167,25 @@ if (BRIDGE_MODE === 'ERC_TO_ERC') {
     BRIDGEABLE_TOKEN_SYMBOL: envalid.str(),
     BRIDGEABLE_TOKEN_DECIMALS: envalid.num(),
     DEPLOY_REWARDABLE_TOKEN: envalid.bool(),
-    DPOS_STAKING_ADDRESS: addressValidator(),
     BLOCK_REWARD_ADDRESS: addressValidator(),
```
akolotov: in this mode `BLOCK_REWARD_ADDRESS` is also applicable only if `DEPLOY_REWARDABLE_TOKEN` is `true`.

deploy/src/loadEnv.js
```diff
@@ -111,10 +147,8 @@ if (BRIDGE_MODE === 'NATIVE_TO_ERC') {
     BRIDGEABLE_TOKEN_NAME: envalid.str(),
     BRIDGEABLE_TOKEN_SYMBOL: envalid.str(),
     BRIDGEABLE_TOKEN_DECIMALS: envalid.num(),
-    FOREIGN_DAILY_LIMIT: bigNumValidator(),
     FOREIGN_MIN_AMOUNT_PER_TX: bigNumValidator(),
-    DEPLOY_REWARDABLE_TOKEN: envalid.bool(),
-    BLOCK_REWARD_ADDRESS: addressValidator()
+    DEPLOY_REWARDABLE_TOKEN: envalid.bool()
```
akolotov: So, we don't need `BLOCK_REWARD_ADDRESS` in this mode, do we?

deploy/src/loadEnv.js
```diff
@@ -34,6 +37,36 @@ const validateRewardableAddresses = (validators, rewards) => {
...
+      `The number of validators VALIDATORS = ${
+        validators.split(' ').length
+      } should be bigger or equal the number of required signatures REQUIRED_NUMBER_OF_VALIDATORS = ${requiredValidators} .`
+    )
+  }
+}
+
+function checkGasPrices(gasPrice, prefix) {
+  if (gasPrice.isZero()) {
+    throw new Error(`${prefix}_GAS_PRICE should be bigger than zero.`)
```
akolotov: in fact the gas price could be zero e.g. for `erc-to-native` bridge when the validators are whitelisted so can send transactions with zero gas price.

deploy/src/loadEnv.js
patitonar: Thanks! Fixed https://github.com/poanetwork/poa-bridge-contracts/pull/228/commits/418580b76c74ffe1640a469027b01d2468d1407d

deploy/src/loadEnv.js
patitonar: Sorry on the confusion, We actually need `BLOCK_REWARD_ADDRESS` for foreign side when `DEPLOY_REWARDABLE_TOKEN` is `true` on `native_to_erc`. Fixed here https://github.com/poanetwork/poa-bridge-contracts/pull/228/commits/857c860dcfbd3f3ab5d3181a108cd7fcadff0357

deploy/src/loadEnv.js
patitonar: Thanks. Now gas price check is applied for all cases except for home gas price on `erc_to_native` mode https://github.com/poanetwork/poa-bridge-contracts/pull/228/commits/8d35e1e6dd82456a4ce60917abe4b510a016bc9f

akolotov: I think that for future POSDAO chain (as home) it can be zero in the `erc-to-erc` mode as well. What if we just check that the gas price either `0` or more than `100` mWei for any side or any mode? Do you see any issue?

patitonar: The only issue I see is that all home and foreign contracts, except home `HomeBridgeErcToNative`, check that gasPrice value is > `0`. For example: 
https://github.com/poanetwork/poa-bridge-contracts/blob/df1fa675138a0d18f3a276cb89b166c77b9c2482/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol#L134

So a user will be able to pass the parameters check by setting home or foreign gas price with a value of `0` but then get the initialization transaction of the bridge reverted

akolotov: Got you! Let's keep the checks as you implemented so far. I will open an issue to allow configure 0 gas price for the `erc-to-erc` mode.

Issue #189, opened 2019-05-24
Change the name of the contracts package 


akolotov:
Currently the package name contains legacy name:
https://github.com/poanetwork/poa-bridge-contracts/blob/bbb97a63c900f03a902d0e82358abac3b294e4d9/package.json#L2
It is necessary to replace it by `token-bridge-contracts`.

As soon as the name is changed and pushed in the `master` branch the PR against [the tokenbridge repo](https://github.com/poanetwork/tokenbridge) must be created to change the name of used package in `package.json`.

akolotov:
Included in [TokenBridge contracts 2.3.3](https://github.com/poanetwork/poa-bridge-contracts/releases/tag/2.3.3)

PR #230, opened 2019-07-12
Update package name

patitonar:
First part of #189

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "poa-parity-bridge-contracts",
+  "name": "token-bridge-contracts",
   "version": "1.0.0",
   "lockfileVersion": 1,
   "requires": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "token-bridge-contracts",
-  "version": "1.0.0",
+  "version": "2.3.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "poa-parity-bridge-contracts",
+  "name": "token-bridge-contracts",
   "version": "1.0.0",
```
akolotov: does it make sense to update the version too? Is it possible to make it auto upgradable if we tag a new release?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "poa-parity-bridge-contracts",
+  "name": "token-bridge-contracts",
   "version": "1.0.0",
```
akolotov: Similar to the comment above.

patitonar: I'll update the version on `package` and `package-lock` files to the current version. There are some tools to automate all the publishing workflow but they seems overkill for our needs. I think we could use [npm version](https://docs.npmjs.com/cli/version) to generate the new releases. Here are some possible steps:

1. Run `npm version <newversion>`. This command will:
- Update `version` on `package.json` and `package-lock.json`
- Commit the changes
- Create a new git tag for the new version

2. Push the commit and tag `git push && git push --tags` (we can skip this manual step if we define a `postversion` script to do that)
3. Create the new GitHub Release using the new tag

akolotov: Thanks for clarification. Probably we will use this tool in the future.

patitonar: Updated version https://github.com/poanetwork/poa-bridge-contracts/pull/230/commits/617f62004ca4c404c5ca27f1c423fe625d26a4a4

PR #256, opened 2019-07-30
Check proxy upgrade contract code

patitonar:
The default behavior of `delegatecall` returns `true` if the contract does not exist. Because of that, it makes sense to check `extcodesize > 0` when setting the implementation contract.

We have repeated implementations of the method `isContract` in different contracts. I removed all the implementation and used the `openzeppelin-solidity` implementation of this method which is exactly the same.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0e65947efbffc592cffea8c2ae9d3b8e11659854/contracts/AddressUtils.sol#L16-L27

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -8,8 +8,6 @@ import "./interfaces/IBurnableMintableERC677Token.sol";
 import "./upgradeable_contracts/Claimable.sol";
 
 contract ERC677BridgeToken is IBurnableMintableERC677Token, DetailedERC20, BurnableToken, MintableToken, Claimable {
-    using AddressUtils for address;
-
     address public bridgeContract;
 
     event ContractFallbackCallFailed(address from, address to, uint256 value);
@@ -19,7 +17,7 @@ contract ERC677BridgeToken is IBurnableMintableERC677Token, DetailedERC20, Burna
     }
 
     function setBridgeContract(address _bridgeContract) external onlyOwner {
-        require(_bridgeContract.isContract());
+        require(AddressUtils.isContract(_bridgeContract));
         bridgeContract = _bridgeContract;
     }
 
@@ -32,7 +30,7 @@ contract ERC677BridgeToken is IBurnableMintableERC677Token, DetailedERC20, Burna
         require(superTransfer(_to, _value));
         emit Transfer(msg.sender, _to, _value, _data);
 
-        if (_to.isContract()) {
+        if (AddressUtils.isContract(_to)) {
             require(contractFallback(msg.sender, _to, _value, _data));
         }
         return true;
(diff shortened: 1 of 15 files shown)


patitonar:
Since `AddressUtils` library only has one method there shouldn't be many changes in gas consumption. However I found that the way in how the library is used there are significant gas differences.

The first approach was to use `using AddressUtils for address;` sintax so then we can do `addressVar.isContract()`. Here are some comparison with code from `develop` branch:

**List of some of the contracts refactored, no new features added**

| Deployments  | develop | Previous Changes | Diff |
|---|---|---|---|
ERC677BridgeToken | 1519320 | 1526562 | 7242
ERC677BridgeTokenRewardable | 1809154 | 1821228 | 12074
HomeBridgeNativeToErc | 4957573 | 4965009 | 7436
ForeignBridgeNativeToErc | 3996542 | 4006198 | 9656
HomeBridgeErcToNative | 6062092 | 6074233 | 12141
FeeManagerErcToErcPOSDAO | 891814 | 894232 | 2418
ERC677BridgeToken.setBridgeContract | 29263 | 29338 | 75

**Contract and method with new feature added for checking the address is a contract**

| Deployments  | develop | Previous Changes | Diff |
|---|---|---|---|
EternalStorageProxy | 332435 | 342230 | 9795
EternalStorageProxy.upgradeTo | 35814 | 36648 | 834

Given the high increase in gas consumption I tried a different approach with better results which is to use the library as `AddressUtils.isContract(addressVar)` (https://github.com/poanetwork/poa-bridge-contracts/pull/256/commits/fe66d3f6c08ad56e98a9dfd55d047b5ca23a459a). Here are some comparison with code from `develop` branch: 

**List of some of the contracts refactored, no new features added**

| Deployments  | develop | Previous Changes | Diff |
|---|---|---|---|
Deployments  | develop | current changes | Diff
ERC677BridgeToken | 1519320 | 1519320 | 0
ERC677BridgeTokenRewardable | 1809154 | 1809154 | 0
HomeBridgeNativeToErc | 4957573 | 4957637 | 64
ForeignBridgeNativeToErc | 3996542 | 3996542 | 0
HomeBridgeErcToNative | 6062092 | 6062092 | 0
FeeManagerErcToErcPOSDAO | 891814 | 891814 | 0
ERC677BridgeToken.setBridgeContract | 29263 | 29263 | 0

**Contract and method with new feature added for checking the address is a contract**

| Deployments  | develop | Previous Changes | Diff |
|---|---|---|---|
EternalStorageProxy | 332435 | 339811 | 7376
EternalStorageProxy.upgradeTo | 35814 | 36573 | 759

As we can see, using the library with the second approach has almost no impact in gas consumption compared to using our own defined `isContract` method.

akolotov:
Thanks for solid investigation!

PR #258, opened 2019-08-01
Add previous values on events

patitonar:
A common practice is to log both the old and new values on the events. This PR updates the events emitted when a configuration parameter is changed in the bridge contracts to log both values.


patitonar:
Right, I agree with you, I think that it makes sense to leave the events as they were defined and emit them in initialization methods so all the values are included in the event history correctly.

patitonar:
I'm closing this PR in favor of https://github.com/poanetwork/poa-bridge-contracts/pull/262

PR #261, opened 2019-08-02
Move mappings key to constants

patitonar:
Moved mappings key to constant. This helps prevent issues caused by typos or erroneous copy-pastes, this also improves readability. Also some repeated code related to the constant was extracted to base contracts. I think the easiest way to review this PR is by reviewing commit per commit.

### DIFF
--- a/contracts/upgradeable_contracts/BlockRewardBridge.sol
+++ b/contracts/upgradeable_contracts/BlockRewardBridge.sol
@@ -1,0 +1,29 @@
+pragma solidity 0.4.24;
+
+import "../interfaces/IBlockReward.sol";
+import "../upgradeability/EternalStorage.sol";
+import "openzeppelin-solidity/contracts/AddressUtils.sol";
+
+contract BlockRewardBridge is EternalStorage {
+    bytes32 internal constant BLOCK_REWARD_CONTRACT = keccak256(abi.encodePacked("blockRewardContract"));
+
+    function _blockRewardContract() internal view returns (IBlockReward) {
+        return IBlockReward(addressStorage[BLOCK_REWARD_CONTRACT]);
+    }
+
+    function _setBlockRewardContract(address _blockReward) internal {
+        require(AddressUtils.isContract(_blockReward));
+
+        // Before store the contract we need to make sure that it is the block reward contract in actual fact,
+        // call a specific method from the contract that should return a specific value
+        bool isBlockRewardContract = false;
+        if (_blockReward.call(abi.encodeWithSignature("blockRewardContractId()"))) {
+            isBlockRewardContract =
+                IBlockReward(_blockReward).blockRewardContractId() == bytes4(keccak256("blockReward"));
+        } else if (_blockReward.call(abi.encodeWithSignature("bridgesAllowedLength()"))) {
+            isBlockRewardContract = IBlockReward(_blockReward).bridgesAllowedLength() != 0;
+        }
+        require(isBlockRewardContract);
+        addressStorage[BLOCK_REWARD_CONTRACT] = _blockReward;
+    }
+}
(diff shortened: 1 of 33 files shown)


contracts/upgradeable_contracts/BlockRewardBridge.sol
```diff
@@ -0,0 +1,29 @@
...
+        return IBlockReward(addressStorage[BLOCK_REWARD_CONTRACT]);
+    }
+
+    function _setBlockRewardContract(address _blockReward) internal {
+        require(AddressUtils.isContract(_blockReward));
+
+        // Before store the contract we need to make sure that it is the block reward contract in actual fact,
+        // call a specific method from the contract that should return a specific value
+        bool isBlockRewardContract = false;
+        if (_blockReward.call(abi.encodeWithSignature("blockRewardContractId()"))) {
```
akolotov: The compiler could prepare the method descriptor on the compilation time and store just 4 bytes in the code now but I am not sure. Could you investigate if it is so or it is worth (from the gas consumption point of view) to store `abi.encodeWithSignature("blockRewardContractId()")` in a constant as well?

patitonar: From [docs](https://solidity.readthedocs.io/en/v0.4.24/contracts.html#constant-state-variables) about Constant State Variables:
`The compiler does not reserve a storage slot for these variables, and every occurrence is replaced by the respective constant expression (which might be computed to a single value by the optimizer).`

I tried 2 options:

### Test 1 
Defined:
```solidity
bytes internal constant BLOCK_REWARD_CONTRACT_ID = abi.encodeWithSignature("blockRewardContractId()");
```

And then using it as:
```solidity
if (_blockReward.call(BLOCK_REWARD_CONTRACT_ID)) {
```
Gas consumption results:

| Contract | Before | After | Diff|
|-----|-----|-----|-----|
HomeBridgeErcToNative | 6110869 | 6110869 | 0
FeeManagerErcToErcPOSDAO | 894232 | 894168 | -64
HomeBridgeErcToNative.setBlockRewardContract | 34362 | 34362 | 0

### Test 2
Defined the selector hash directly in the constant
```solidity
bytes4 internal constant BLOCK_REWARD_CONTRACT_ID = 0x2ee57f8d; // blockRewardContractId() selector
```

Gas consumption results:

| Contract | Before | After | Diff|
|-----|-----|-----|-----|
HomeBridgeErcToNative | 6110869 | 6082606 | -28263
FeeManagerErcToErcPOSDAO | 894232 | 860151 | -34081
HomeBridgeErcToNative.setBlockRewardContract | 34362 | 34160 | -202 

From test 1, it seems that it is not worth to extract the signature to a constant from gas perspective. However, it looks like if we define the constant as in test 2 we can save some considerable gas. We can apply this to similar other cases on the other contracts too. If we do this, I think we should have a comment on each constant to know the selector it is referring to

akolotov: Nice research!
Can the line
```
bytes internal constant BLOCK_REWARD_CONTRACT_ID = abi.encodeWithSignature("blockRewardContractId()");
```
consume more gas since there is no exact size of data?

I agree with putting just hexadecimal value of the method selector to optimize gas usage.

patitonar:
Added the selector hexadecimal values as constants. Here are the main gas consumption changes

| Contract | Develop | Current Branch | Diff|
|-----|-----|-----|-----|
FeeManagerErcToNative | 1175450 | 1175386 | -64
FeeManagerErcToErcPOSDAO | 891814 | 826663 | −65151
HomeBridgeErcToErc | 5788764 | 5713926 | −74838
HomeBridgeErcToErcPOSDAO | 5953366 | 5878208 | −75158
HomeBridgeErcToNative | 6105969 | 5979713 | −126256
HomeBridgeNativeToErc | 4973809 | 4898591 | −75218
ForeignBridgeNativeToErc | 4012906 | 3937875 | −75031

PR #272, opened 2019-08-14
Process validator's reward accounts only when used in deploy script

### LINKED ISSUE #271: VALIDATORS_REWARD_ACCOUNTS can be omitted for native-to-erc mode
Currently `VALIDATORS_REWARD_ACCOUNTS` should be configured for `native-to-erc` mode even if both `HOME_REWARDABLE` and `FOREIGN_REWARDABLE` are `false`. But this requirement is not correct. There is no need to configure accounts to get reward for bridge operations if fee manager is not used.

https://github.com/poanetwork/poa-bridge-contracts/blob/048f51e0e8e7ce4213b727c0982aadab6c994db9/deploy/src/native_to_erc/home.js#L27

patitonar:
Closes #271

### DIFF
--- a/deploy/src/erc_to_native/home.js
+++ b/deploy/src/erc_to_native/home.js
@@ -52,7 +52,7 @@ const isFeeManagerPOSDAO = HOME_FEE_MANAGER_TYPE === 'POSDAO_REWARD'
 
 let VALIDATORS_REWARD_ACCOUNTS = []
 
-if (isRewardableBridge) {
+if (isRewardableBridge && !isFeeManagerPOSDAO) {
   VALIDATORS_REWARD_ACCOUNTS = env.VALIDATORS_REWARD_ACCOUNTS.split(' ')
 }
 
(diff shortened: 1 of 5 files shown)


deploy/src/erc_to_native/home.js
```diff
@@ -51,6 +50,12 @@ const DEPLOYMENT_ACCOUNT_ADDRESS = privateKeyToAddress(DEPLOYMENT_ACCOUNT_PRIVAT
 const isRewardableBridge = HOME_REWARDABLE === 'BOTH_DIRECTIONS'
 const isFeeManagerPOSDAO = HOME_FEE_MANAGER_TYPE === 'POSDAO_REWARD'
 
+let VALIDATORS_REWARD_ACCOUNTS = []
+
+if (isRewardableBridge) {
```
akolotov: In case of POSDAO Fee Manager we also don't need to specify validators reward accounts since the reward will be distributed among the network validators and stakers. The same could be applicable for `erc-to-erc` mode.

patitonar: Added special checks for POSDAO Fee Manager https://github.com/poanetwork/poa-bridge-contracts/pull/272/commits/3d892ed69169ee8737d66116940ef896740b49e3

PR #276, opened 2019-09-06
Update project dependencies

### LINKED ISSUE #275: Security alerts on 06.09.2019
The following alerts raised by GitHub:
* CVE-2019-10747 (set-value)
  **Recommendation**: upgrade to 2.0.1 or later 
* CVE-2019-10746 (mixin-deep)
  **Recommendation**: upgrade to 1.3.2 or later 
* GHSA-3gx7-xhv7-5mx3 (eslint-utils)
  **Recommendation**: upgrade to 1.4.1 or later

patitonar:
Closes #275 

Updated main and `deploy` dependencies using `npm audit` tool.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,60 +5,184 @@
   "requires": true,
   "dependencies": {
     "@0x/assert": {
-      "version": "2.0.10",
-      "resolved": "https://registry.npmjs.org/@0x/assert/-/assert-2.0.10.tgz",
-      "integrity": "sha512-tM3j5aZnJGTMr+7OAF61gmflrgEpUkp1LHNsnIx5VnVG+X7JzwV87O44PU9bALVQI1ufYvAUwhIkmK8jbR5Agw==",
+      "version": "2.1.5",
+      "resolved": "https://registry.npmjs.org/@0x/assert/-/assert-2.1.5.tgz",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -9316,43 +10259,39 @@
...
           "requires": {
             "esrecurse": "^4.1.0",
             "estraverse": "^4.1.1"
           }
         },
         "eslint-utils": {
-          "version": "1.3.1",
-          "resolved": "https://registry.npmjs.org/eslint-utils/-/eslint-utils-1.3.1.tgz",
-          "integrity": "sha512-Z7YjnIldX+2XMcjr7ZkgEsOj/bREONV60qYeB/bjMAqqqZ4zxKyWX+BOUkdmRmA9riiIPVvo5x86m5elviOk0Q=="
+          "version": "1.4.0",
```
akolotov: Does it make sense to have `1.4.2` here as well?

patitonar: The version of `eslint-utils` installed for the project in `node_modules` will be `1.4.2` which is defined in line `4676`. In this case this other `1.4.0` version of `eslint-utils` is a dependency of `ganache-core`. So it will be located inside `node_modules/ganache-core/node_modules` and only used in the context of `ganache-core`. We are already using the last version of this library so we cannot upgrade it to a version that uses a newer version of `eslint-utils`. I think its OK  to use that version since it is a helper library for `eslint` which is used to lint the code. So in the only case where this version will be used is when linting the code of `ganache-core`, operation that we do not perform as part of our project. We only run the linter for our code and in that case `1.4.2` will be used

akolotov: Thanks Gerardo for explanation.

PR #282, opened 2019-09-09
Add result information for AMB message call

### LINKED ISSUE #281: Result of message call for AMB feature
This issue is opened as reflection corollary of the issue described here: https://github.com/poanetwork/poa-bridge-contracts/issues/267#issuecomment-529027876.

The main idea is to provide an ability to properly handle a situation when a message call was failed as part of an attempt to relay it from one chain to another. At this particular moment we just report about the invocation status by the event:

https://github.com/poanetwork/poa-bridge-contracts/blob/9c19ec51ac23d83d44d50a71f63fe01f749547c1/contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol#L40-L55

But it is not enough since there is no simple way to have a reliable confirmation that the call causes the fail or success.

It is very undesired to use a scheme when the bridge invokes a call back if the status of the message execution is `false`. The reason is possibility of an attack to dry out the validator balance or spam the bridge. An attacker will generate such relay request so the validators will spend lots of resources to execute the call back.

Below is a proposal to provide reliable information about the message call status:

1. The bridge contract will store the status of every message call by associating it with the originating message hash (or something like this). Eventually we need to have correspondence between an originating call on one side and the result on another side. This information will be available for other contracts.
2. A contract (that uses the bridge) could requests the status of the message call by its hash. After that the contract is able to send a message call on another side of the bridge reporting that the original call was failed. 
3. The contract-receiver gets the confirmation (relayed through the bridge) that the corresponding call was failed and performs actions to recover the unsuccessful relay request. E.g. for the `erc677-to-erc677` bridge the mediator contract could unlock/mint the corresponding number of tokens. And it assumes that the contract mediator (that will be the contract-receiver then) needs to store a correspondence between the relay request and his attributes (the sender and value).

patitonar:
Closes #281 

Implemented the approach mentioned in [this comment](https://github.com/poanetwork/poa-bridge-contracts/issues/281#issuecomment-529516782).

Also added `indexed` property to parameters of AMB events. This could be useful for filtering events from/to a specific mediator or by transaction hash.

### DIFF
--- a/contracts/interfaces/IAMB.sol
+++ b/contracts/interfaces/IAMB.sol
@@ -5,6 +5,8 @@ interface IAMB {
     function maxGasPerTx() external view returns (uint256);
     function transactionHash() external view returns (bytes32);
     function messageCallStatus(bytes32 _txHash) external view returns (address);
-    function messageDataHash(bytes32 _txHash) external view returns (bytes32);
+    function failedMessageDataHash(bytes32 _txHash) external view returns (bytes32);
+    function failedMessageReceiver(bytes32 _txHash) external view returns (address);
+    function failedMessageSender(bytes32 _txHash) external view returns (address);
     function requireToPassMessage(address _contract, bytes _data, uint256 _gas) external;
 }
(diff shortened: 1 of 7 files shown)


patitonar:
@akolotov Updated the implementation to store the receiver and sender for failed transactions on AMB. A separate PR will be provided for using the new methods on `erc677-to-erc677` mode

PR #283, opened 2019-09-09
Update project dependencies to latest versions

### LINKED ISSUE #278: Security alerts on 07.09.2019
The following alerts raised by GitHub:
* WS-2018-0590 (diff)
  **Recommendation**: upgrade to 3.5.0 or later 
* WS-2018-0236 (mem)
  **Recommendation**: upgrade to 4.0.0 or later

patitonar:
Closes #278 

Updated all dependencies to the latest versions. 

`prettier-plugin-solidity` updated some of the rules  that are not configurable so some minor format changes where made to the contracts related to the usage of `**` operator and the usage of parenthesis in some operations.

I wasn't able to update the dependency related to the alert:
```
WS-2018-0236 (mem)
Recommendation: upgrade to 4.0.0 or later
```
The `mem` library is a very deep dependency of some of the `0x` tools that we use for the code coverage report and there is no versions that uses a newer version of `mem`.


akolotov:
> I wasn't able to update the dependency related to the alert:
> `WS-2018-0236 (mem)`
> `Recommendation: upgrade to 4.0.0 or later`

Do you have ideas what is purpose of this tool? May you add a short description that there is no security risk even if we have older version of the library in our repo?

patitonar:
> Do you have ideas what is purpose of this tool? May you add a short description that there is no security risk even if we have older version of the library in our repo?

`mem` memoize functions, used to speed up consecutive function calls by caching the result of calls with identical input. Before version 4.0.0 has a memory leak due to old results not being removed from the cache despite reaching `maxAge`.

The `mem` package is used by our tools for code coverage report: `@0x/sol-compiler -> yargs -> os-locale -> mem`. So there is no risk in our contracts or deploy scripts.

Also, the `os-locale` package does not specify `maxAge` in [its uses](https://github.com/sindresorhus/os-locale/blob/966ff9d03ea8c7911fd71e57fb60f042ceaea73e/index.js#L4-L6), keeping the cached values in memory forever. So it seems this bug is not relevant to the use of `mem` in our dependencies.

akolotov:
thanks for explanation!

PR #302, opened 2019-10-08
Add alternative receiver for transfer in native to erc20 mode

### LINKED ISSUE #297: Alternative receiver in NATIVE-to-ERC20 mode
This issue is to implement suggestions from #295 for the contracts related to the `native-to-erc20` bridge mode

patitonar:
Closes #297

### DIFF
--- a/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
+++ b/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
@@ -2,10 +2,19 @@ pragma solidity 0.4.24;
 
 import "./ERC677Bridge.sol";
 import "../interfaces/IBurnableMintableERC677Token.sol";
+import "../libraries/Bytes.sol";
 
 contract ERC677BridgeForBurnableMintableToken is ERC677Bridge {
-    function bridgeSpecificActionsOnTokenTransfer(ERC677 _token, address _from, uint256 _value) internal {
+    function getReceiver(address _from, bytes _data) internal view returns (address recipient) {
+        recipient = _from;
+        if (_data.length == 20) {
+            recipient = Bytes.bytesToAddress(_data);
+        }
+        require(recipient != address(0));
+    }
+
+    function bridgeSpecificActionsOnTokenTransfer(ERC677 _token, address _from, uint256 _value, bytes _data) internal {
         IBurnableMintableERC677Token(_token).burn(_value);
-        fireEventOnTokenTransfer(_from, _value);
+        fireEventOnTokenTransfer(getReceiver(_from, _data), _value);
     }
 }
--- a/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
+++ b/contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
@@ -2,20 +2,8 @@ pragma solidity 0.4.24;
 
 import "./ERC677Bridge.sol";
 import "../interfaces/IBurnableMintableERC677Token.sol";
-import "../libraries/Bytes.sol";
-import "./OtherSideBridgeStorage.sol";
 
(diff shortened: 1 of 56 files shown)


contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
```diff
@@ -2,15 +2,22 @@ pragma solidity 0.4.24;
...
 
-contract ERC677BridgeForBurnableMintableToken is ERC677Bridge {
-    function bridgeSpecificActionsOnTokenTransfer(
-        ERC677 _token,
-        address _from,
-        uint256 _value,
-        bytes /*_data*/
-    ) internal {
+contract ERC677BridgeForBurnableMintableToken is ERC677Bridge, OtherSideBridgeStorage {
+    function chooseReceiver(address _from, bytes _data) internal view returns (address recipient) {
```
akolotov: seems that we have the same method in `amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol`. Does it make sense to have a separate module for it?

contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
```diff
@@ -22,7 +22,8 @@ contract ForeignBridgeNativeToErc is
         uint256 _requiredBlockConfirmations,
         uint256[] _homeDailyLimitHomeMaxPerTxArray, // [ 0 = _homeDailyLimit, 1 = _homeMaxPerTx ]
         address _owner,
-        uint256 _decimalShift
+        uint256 _decimalShift,
+        address _bridgeOnOtherSide
```
akolotov: is my understanding correct and the scenario `approve-then-relayTokens` is not supported in the bridge mode. Do you think it make sense to keep consistency and support it for all bridge modes where erc677 tokens are intended to be used? If yes, I will create a new issue.

patitonar: You are correct, `approve-then-relayTokens` was only added for the modes where `erc677` is not present in order to be able to specify an alternative receiver. I think that since `erc677` can specify a different receiver using `transferAndCall` method, `relayTokens` is not needed for those cases because it will be a duplicated functionality and from user perspective requires two transactions instead of one so I think `transferAndCall` will be always preferred

contracts/upgradeable_contracts/ERC677BridgeForBurnableMintableToken.sol
patitonar: Moved definition of the method https://github.com/poanetwork/poa-bridge-contracts/pull/302/commits/fd800242b38fcc04946b8b5cceddc3ed4970cf77

contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
akolotov: right. but consider a following situation: a user knows that on the bridge `A` he/she can transfer tokens by pairing `approve` and `relayTokens`. Next, the same user tries to transfer tokens through the bridge `B` but fails. In both cases he/she considers the token contract as erc20 compatible. But in one case it works and does not work in another. I assumed this consistency when raised the comment initially. I understand that's it is not mandatory to have but asking your opinion whether we can suggest the user this level of usability in the future.

patitonar: Thanks for the explanation, in that case keeping consistency and allowing the users to interact in the same way with all the bridges is a nice usability feature, let's create a new issue for this.

akolotov: Ok. I will do.

PR #303, opened 2019-10-09
Use hex-identifiers directly

maxaleks:


varasev:
@maxaleks Please fix the syntax warnings Travis reported about

varasev:
I think we should base this PR on the `develop` branch, not the `master`. Am I right, @akolotov ?

contracts/mocks/FeeManagerMock.sol
```diff
@@ -12,7 +12,7 @@ contract FeeManagerMock is BaseFeeManager {
     }
 
     function getFeeManagerMode() external pure returns (bytes4) {
-        return bytes4(keccak256(abi.encodePacked("manages-one-direction")));
+        return 0xf2aed8f7; // "manages-one-direction"
```
varasev: Please clarify in the comments how the hash was got:

// bytes4(keccak256(abi.encodePacked("manages-one-direction")));

instead of just

// "manages-one-direction"

PR #306, opened 2019-10-14
Relative daily limit

maxaleks:


deploy/src/loadEnv.js
```diff
@@ -111,7 +112,8 @@ let validations = {
   FOREIGN_RPC_URL: envalid.str(),
   FOREIGN_BRIDGE_OWNER: addressValidator(),
   FOREIGN_UPGRADEABLE_ADMIN: addressValidator(),
-  FOREIGN_MAX_AMOUNT_PER_TX: bigNumValidator()
+  FOREIGN_MAX_AMOUNT_PER_TX: bigNumValidator(),
+  RELATIVE_DAILY_LIMIT: envalid.bool(),
```
akolotov: this variable should not be checked for AMB since it does not make sense there.

deploy/src/amb_erc677_to_erc677/foreign.js
```diff
@@ -2,9 +2,16 @@ const Web3Utils = require('web3-utils')
 const { web3Foreign, FOREIGN_RPC_URL } = require('../web3')
 const { deployContract, privateKeyToAddress, upgradeProxy } = require('../deploymentUtils')
 const {
-  foreignContracts: { EternalStorageProxy, ForeignAMBErc677ToErc677: ForeignBridge }
+  foreignContracts: {
+    EternalStorageProxy,
+    ForeignAMBErc677ToErc677: ForeignBridge,
```
akolotov: Does it make sense to rename `ForeignBridge` to `ForeignBridgeAbsoluteDailyLimit`?

deploy/src/amb_erc677_to_erc677/home.js
```diff
@@ -17,13 +17,15 @@ const {
...
   BLOCK_REWARD_ADDRESS,
-  DPOS_STAKING_ADDRESS
+  DPOS_STAKING_ADDRESS,
+  RELATIVE_DAILY_LIMIT,
 } = require('../loadEnv')
 
 const {
   homeContracts: {
     EternalStorageProxy,
     HomeAMBErc677ToErc677: HomeBridge,
```
akolotov: Does it make sense to rename `HomeBridge` to `HomeBridgeAbsoluteDailyLimit`?

patitonar:
Based on the comment from https://github.com/poanetwork/tokenbridge-contracts/issues/287#issuecomment-544859675
>If we consider two operations deposit and withdrawal such as the balance of the bridge is increased and after the first one and reduced after the second one, so the relative daily limits must be applied only for the withdrawals. For the deposits absolute limits should be used still.

it looks like the implementation is **correctly** applying the daily limits, here are some notes from my analysis:

**Native - ERC**
- Native tokens deposit in Home: 
  * Absolute daily limit check in Home **(OK)**
  * Absolute execution daily limit check in Foreign **(OK)**
- Token withdrawal in Foreign:
  * Relative daily limit check in Foreign **(OK)**
  * Relative execution daily limit check in Home **(OK)**

**ERC - Native**
- Tokens deposit in Foreign:
  * No limits checks can be provided to commons Transfers **(OK)**
  * Absolute daily limit check in Foreign for transfer calling relayTokens **(OK)**
  * Absolute execution daily limit in Home **(OK)**
- Native token withdrawal in Home:
  * Relative daily limit check in Home **(OK)**
  * Relative execution daily limit check in Foreign **(OK)**

**ERC - ERC**
- Tokens deposit in Foreign:
  * No limits checks can be provided to commons Transfers **(OK)**
  * Absolute daily limit check in Foreign for transfer calling relayTokens **(OK)**
  * Absolute daily limit check in Foreign for transfer when using ERC677 **(OK)**
  * Absolute execution daily limit check in Home **(OK)**
- Tokens withdrawal in Home:
  * Relative daily limit check in Home **(OK)**
  * Relative execution daily limit check in Foreign **(OK)**

**AMB-ERC-ERC**
- Tokens deposit in Foreign:
  * Absolute daily limit check in Foreign **(OK)**
  * Absolute execution daily limit check in Home **(OK)**
- Tokens withdrawal in Home:
  * Relative daily limit check in Home **(OK)**
  * Relative execution daily limit check in Foreign **(OK)**

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignAMBErc677ToErc677RelativeDailyLimit.sol
```diff
@@ -0,0 +1,54 @@
...
+import "../RelativeExecutionDailyLimit.sol";
+
+contract ForeignAMBErc677ToErc677RelativeDailyLimit is ForeignAMBErc677ToErc677, RelativeExecutionDailyLimit {
+    function handleBridgedTokens(
+        address _recipient,
+        uint256 _value,
+        bytes32 _nonce
+    ) public {
+        _updateTodayLimit();
+        super.handleBridgedTokens(_recipient, _value, _nonce);
```
patitonar: Do we need to call `_updateTodayLimit`? It seems that this method calls `_calculateLimit()` the first time when the limit is not set.
Then inside `handleBridgedTokens`, the method `withinExecutionLimit` will be called, then `executionDailyLimit`,  and then `_getTodayLimit()` which will try to call `_calculateLimit()` if it wasn't calculated.

Unless I'm missing something I think we could remove `_updateTodayLimit` and its calls so only `_getTodayLimit` is called and update `_getTodayLimit` to something like this:
```
function _getTodayLimit() internal returns (uint256) {
    uint256 limit = _todayLimit();
    if (limit == 0) {
        // not set yet
        limit = _calculateLimit();
        _setTodayLimit(limit);
        emit TodayLimitSet(limit);
    }
    return limit;
}
```
This same behavior could be applied for all other bridge modes where relative daily limits are used too.
What do you think?

maxaleks: Good catch, @patitonar!
I thought about it. But if we call `_setTodayLimt` inside `_getTodayLimit` then `executionDailyLimit` and `dailyLimit` will be not `view` methods

patitonar: You're right, I missed that. Thanks

contracts/upgradeable_contracts/BaseRelativeDailyLimit.sol
```diff
@@ -0,0 +1,85 @@
...
+    function _calculateLimit() internal view returns (uint256) {
+        uint256 balance = _getTokenBalance();
+        uint256 unlimitedBalance = _minPerTx();
+        if (balance <= unlimitedBalance) {
+            return balance;
+        }
+        uint256 limit = targetLimit();
+        uint256 thresh = threshold();
+        uint256 multiplier = 1 ether ** 2;
+        if (balance > unlimitedBalance && balance < thresh) {
```
patitonar: I think the check `balance > unlimitedBalance` is not needed since it will always be `true` because of previous check 
```javascript
 if (balance <= unlimitedBalance) {
    return balance;
}
```

contracts/upgradeable_contracts/BaseRelativeDailyLimit.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    function targetLimit() public view returns (uint256) {
+        return uintStorage[TARGET_LIMIT];
+    }
+
+    function threshold() public view returns (uint256) {
+        return uintStorage[THRESHOLD];
+    }
+
+    function setTargetLimit(uint256 _targetLimit) external onlyOwner {
```
patitonar: I think we should a unit test to cover this method

contracts/upgradeable_contracts/BaseRelativeDailyLimit.sol
```diff
@@ -0,0 +1,85 @@
...
+        require(_targetLimit <= 1 ether);
+        uintStorage[TARGET_LIMIT] = _targetLimit;
+        emit TargetLimitChanged(_targetLimit);
+    }
+
+    function setThreshold(uint256 _threshold) external onlyOwner {
```
patitonar: I think we should a unit test to cover this method

contracts/upgradeable_contracts/BaseRelativeDailyLimit.sol
maxaleks: Yes, you are right. I saw it, but decided not to change change the proposed implementation https://github.com/poanetwork/tokenbridge-contracts/pull/306#discussion_r334659756 until the review

contracts/upgradeable_contracts/BaseRelativeDailyLimit.sol
maxaleks: In which file do you recommend adding this test?

patitonar: One option could be `AMBErc677ToErc677Behavior.test.js` since we are testing `targetLimit()` and `threshold()` there too

contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErcRelativeDailyLimit.sol
```diff
@@ -0,0 +1,78 @@
...
+        _updateTodayLimit();
+        super.executeAffirmation(recipient, value, transactionHash);
+    }
+
+    function initialize(
+        address _validatorContract,
+        uint256[] _requestLimitsArray, // [ 0 = _dailyLimit, 1 = _maxPerTx, 2 = _minPerTx ]
+        uint256 _homeGasPrice,
+        uint256 _requiredBlockConfirmations,
+        uint256[] _executionLimitsArray, // [ 0 = _targetLimit, 1 = _tthreshold, 2 = _foreignMaxPerTx, 3 = _foreignMinPerTx ]
```
patitonar:
```suggestion
        uint256[] _executionLimitsArray, // [ 0 = _targetLimit, 1 = _threshold, 2 = _foreignMaxPerTx, 3 = _foreignMinPerTx ]
```

contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErcRelativeDailyLimit.sol
```diff
@@ -0,0 +1,78 @@
...
+        setInitialize();
+        return isInitialized();
+    }
+
+    function rewardableInitialize(
+        address _validatorContract,
+        uint256[] _requestLimitsArray, // [ 0 = _dailyLimit, 1 = _maxPerTx, 2 = _minPerTx ]
+        uint256 _homeGasPrice,
+        uint256 _requiredBlockConfirmations,
+        uint256[] _executionLimitsArray, // [ 0 = _targetLimit, 1 = _tthreshold, 2 = _foreignMaxPerTx, 3 = _foreignMinPerTx ]
```
patitonar:
```suggestion
        uint256[] _executionLimitsArray, // [ 0 = _targetLimit, 1 = _threshold, 2 = _foreignMaxPerTx, 3 = _foreignMinPerTx ]
```

contracts/upgradeable_contracts/RelativeDailyLimit.sol
```diff
@@ -0,0 +1,23 @@
...
+
+    function setMinPerTx(uint256 _minPerTx) external onlyOwner {
+        require(_minPerTx < maxPerTx());
+        uintStorage[MIN_PER_TX] = _minPerTx;
+    }
+
+    function setMaxPerTx(uint256 _maxPerTx) external onlyOwner {
+        require(_maxPerTx > minPerTx());
+        uintStorage[MAX_PER_TX] = _maxPerTx;
+    }
```
patitonar: In the current implementation when we need to stop bridge operations (for example in case for xDai bridge upgrade) we use `setMaxPerTx` to set this parameter to `0`. In this method implementation, while the check `require(_maxPerTx > minPerTx());` does makes sense for the integrity of the parameters we will be loosing the ability to stop the bridge operations. @akolotov  does it makes sense to remove this check in order to be able to stop bridge operations?

patitonar:
@akolotov @maxaleks 
I'm adding this comment so we can debate on the best approach for the architecture of the contracts, so we can explore and analyse some alternatives before moving forward with this implementation.

From my perspective the implementation in this PR of adding a new `Home` and `Foreing` end contract with the relative daily limit feature (Ex in Native-to-Erc: `HomeBridgeNativeToErc` and `HomeBridgeNativeToErcRelativeDailyLimit`) have some advantages:
- The implementations are well separated so if someone wants to use absolute limits it avoids expending extra gas with the methods of relative limits.

and disadvantages:
- If after a first deployed version a user wants to change from using absolute limits to relative (or the other way around) it has to deploy and upgrade the implementation contract.
- We have to maintain, test and deploy different versions of Home and Foreign contracts for each mode. In the case a new feature is added and it behaves differently on each existing case we could end up duplicating our contracts (ex: `HomeBridgeNativeToErc`, `HomeBridgeNativeToErcFeatureX`, `HomeBridgeNativeToErcRelativeDailyLimit` and `HomeBridgeNativeToErcRelativeDailyLimitFeatureX`).

I tried to think on other ideas and I wasn't able to find an ideal solution but here are some other options as an starting point.

**Option 1**
Have one contract where depending on initialization absolute or relative will be used, based on the inputs. For example the initialize method could support all parameters to be set in the limits arrays, and if the absolute daily limit is not zero, it will be used. Relative daily limit logic could be integrated directly but when calling `_calculateLimit` the absolute daily limit could be also returned if it was configured instead of the calculated one for the relative limit case.

Advantages:
- Only one home and foreign contract, which will help scale on adding future features.
- Contract can move from absolute or relative from one way to another easily by calling a method to the set the parameter.

Disadvantages:
- The contract will have a some code that might not be used (specially if absolute is used). A similar situation that is present when a bridge is not using the fee manager, there is some logic to interact with the fee manager contract that is not used.

**Option 2**
Similar to Option 1, only one contract but we can consider to move some logic to a different deployed implementation contract and delegate the calls similar as we do with fee manager. 

Here are some options: 
- we can have for example the absolute limit as default, deploy and delegate the calls to the other contract if relative limit logic wants to be used.
- Absolute limit logic and relative limit logic are both in external contracts

Advantages:
- Only one home and foreign contract
- Contract can move from absolute and relative from one way to another easily (depending on the option used)

Disadvantages:
- One more external contract to be deployed
- Overhead of methods to interact with external logic contract and one extra parameter to store the address

I would like to know your opinions on the current approach, the presented options or add new ones. There are some advantages and disadvantages for each case, we need to think which are our pririorities about the solution in order to select the option that better fit our case.

contracts/upgradeable_contracts/RelativeDailyLimit.sol
maxaleks: Are you about `setDalyLimit` method? https://github.com/poanetwork/tokenbridge-contracts/blob/develop/contracts/upgradeable_contracts/BasicTokenBridge.sol#L75

We can make the same check for this implementation of `setMaxPerTx`:
```
require(_maxPerTx > minPerTx() || _maxPerTx == 0);
```

patitonar: Yes, I think having that check it would work

varasev:
Hi,

We also should remember of maximum contract's bytecode size of 24 Kb on Ethereum side. On xDai side, we could increase this limit, but we shouldn't increase that too much.

So, I believe we should take into account the contract's size and choose the approach which would let us keep the size as small as possible (but also consider contract-to-contract interactions overhead, of course).

The overhead is critical on Ethereum side because bridge validators pay for transactions in ETH. But it's not critical on xDai side (because transactions there are much cheaper).

maxaleks:
Hi,
I agree with both of you, @patitonar and @varasev. So I would like to hear the opinion of @akolotov

akolotov:
> So, I believe we should take into account the contract's size and choose the approach which would let us keep the size as small as possible (but also consider contract-to-contract interactions overhead, of course).
> 
> The overhead is critical on Ethereum side because bridge validators pay for transactions in ETH. But it's not critical on xDai side (because transactions there are much cheaper).

Sorry for delay with answer. Since we are near to limit of the contract size for the Ethereum Mainnet block we need to chose such approaches that will reduce the size of the code. So it is possible either by having _small_ contracts for every combination of possible deployment options or by moving all functionality to external contracts/libraries.

The comment about increase of gas usage when external contracts/libraries are used is valid. That's why I did a simple experiment:
  * there are two contracts `Test0` (library) and `Test1` (main contract)
  ```solidity
  pragma solidity 0.5.11;

  contract Storage {
    uint256 public result;
  }

  contract Test0 is Storage {
    function a(uint256 _v) public {
      result = _v * _v;
    }
  }

  contract Test1 is Storage {
    Test0 extcont;

    constructor (address _a) public {
      extcont = Test0(_a);
    }

    function b(uint256 _v) public {
      result = _v * _v;
    }    

    function ext_b() external {
      b(4);
    }

    function ext_b_this() external {
      Test1(this).b(3);
    }

    function ext_a() external {
      address(extcont).delegatecall(abi.encodeWithSelector(Test0(extcont).a.selector, 5)) ;
    }
  }
  ``` 
  * when the storage was warmed up (by initializing the variable `result`) the invocations of the `Test1` methods causes the following gas consumption:
    * `ext_a`: 7131
    * `ext_b_this`: 6985
    * `ext_b`: 5261

So, as you can see the difference is not large between calls `ext_b` and `ext_a`. Yes, it is almost 1k but I thinks it is reasonable trade-off to continue enhance the bridge functionality without significant increase of the complexity of the code maintenance. That's why I vote for implementing the mechanism of limits in form of the external contract/library (both absolute and relative limits) - changing the approach to operate with limits will be performing by deploying a new external implementation and changing the address in the bridge contract. As Gerardo said it is similar to what we have for the fee management now.

varasev:
@maxaleks could you please measure the size of the modified contracts in this `relative-daily-limit` branch? For `Home` and `Foreign` versions.

varasev:
I agree with the proposal if that would help to reduce the size of the code significantly. So I think we need to calculate the current code size in `relative-daily-limit` and delta between it and the size without limits logic, i.e., just compare the sizes of the contracts in `develop` and `relative-daily-limit` to know how many Kb approximately we could win.

maxaleks:
Approximate sizes:

|Contract  | Size (kb) |
| ------------- | :---: |
| ForeignAMBErc677ToErc677  | 14.5  |
| ForeignAMBErc677ToErc677RelativeDailyLimit  | 16  |
| HomeAMBErc677ToErc677  | 14.5  |
| HomeAMBErc677ToErc677RelativeDailyLimit  | 16  |
| ForeignBridgeErc677ToErc677  | 11  |
| ForeignBridgeErc677ToErc677RelativeDailyLimit  | 12.5  |
| HomeBridgeErcToErc  | 20  |
| HomeBridgeErcToErcRelativeDailyLimit  | 21.5  |
| HomeBridgeErcToErcPOSDAO  | 21  |
| HomeBridgeErcToErcPOSDAORelativeDailyLimit  | 22  |
| ForeignBridgeErcToNative  | 12.5  |
| ForeignBridgeErcToNativeRelativeDailyLimit  | 14  |
| HomeBridgeErcToNative  | 20.5  |
| HomeBridgeErcToNativeRelativeDailyLimit  | 24  |
| ForeignBridgeNativeToErc  | 14  |
| ForeignBridgeNativeToErcRelativeDailyLimit  | 15  |
| HomeBridgeNativeToErc  | 17  |
| HomeBridgeNativeToErcRelativeDailyLimit  | 18.5  |

varasev:
@maxaleks OK, let's implement @akolotov's proposal then. The size table would be helpful to compare sizes after the proposed approach is implemented.

maxaleks:
@varasev, ok, I will start working on it

maxaleks:
I moved all logic related to limits to external contract and removed all duplicate contracts with ending `...RelativeDailyLimit`. Also I optimized some `initialize` methods

Tests are not yet updated

Comparison with the old version:

|Contract  | New size (kb) | Old size (kb) | Old size (relative limit) (kb) |
| ------------- | :---: | :---: | :---: |
| ForeignAMBErc677ToErc677  | 13.5  | 14.5 | 16 |
| HomeAMBErc677ToErc677  | 13.5  | 14.5 | 16 |
| ForeignBridgeErc677ToErc677  | 10  | 11 | 12.5 |
| HomeBridgeErcToErc  | 19  | 20 | 21.5 |
| HomeBridgeErcToErcPOSDAO  | 19.5  | 21 | 22 |
| ForeignBridgeErcToNative  | 11.5  | 12.5 | 14 |
| HomeBridgeErcToNative  | 19.5  | 20.5 | 22 |
| ForeignBridgeNativeToErc  | 12.5  | 14 | 15 |
| HomeBridgeNativeToErc  | 15.7  | 17 | 18.5 |

I think the new implementation is simpler and more elegant, thanks @patitonar and @akolotov for the idea

varasev:
> Tests are not yet updated

@maxaleks OK, so currently everything is ready in this PR except unit tests, right?

maxaleks:
> @maxaleks OK, so currently everything is ready in this PR except unit tests, right?

Except unit tests and deployment scripts. Possibly some logic will be changed during the tests update

contracts/upgradeable_contracts/BasicRelativeDailyLimit.sol
```diff
@@ -0,0 +1,85 @@
+pragma solidity 0.4.24;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "./BasicLimits.sol";
+
+contract BasicRelativeDailyLimit is BasicLimits {
```
patitonar: Is it correct that here `BasicLimits` should be also be renamed to `AbsoluteDailyLimit`?

contracts/upgradeable_contracts/BasicHomeBridge.sol
```diff
@@ -22,7 +22,8 @@ contract BasicHomeBridge is EternalStorage, Validatable, BasicBridge, BasicToken
         uint256 NumberOfCollectedSignatures
     );
 
-    function executeAffirmation(address recipient, uint256 value, bytes32 transactionHash) external onlyValidator {
+    function executeAffirmation(address recipient, uint256 value, bytes32 transactionHash) public onlyValidator {
```
patitonar: Is there any reason to update the modifier of `executeAffirmation` from `external` to `public`?

contracts/upgradeable_contracts/BasicTokenBridge.sol
```diff
@@ -7,95 +7,180 @@ import "./Ownable.sol";
...
-        require(_maxPerTx < dailyLimit());
-        uintStorage[MAX_PER_TX] = _maxPerTx;
+    function _execute(bytes4 _method, uint256 _value) internal {
+        require(limitsContract().delegatecall(abi.encodeWithSelector(_method, _value)));
     }
 
-    function setMinPerTx(uint256 _minPerTx) external onlyOwner {
-        require(_minPerTx < dailyLimit() && _minPerTx < maxPerTx());
-        uintStorage[MIN_PER_TX] = _minPerTx;
+    function _getTokenBalance() internal view returns (uint256) {}
```
patitonar: Do you think we could declare `_getTokenBalance` as an abstract method instead of an empty one?
```suggestion
    function _getTokenBalance() internal view returns (uint256);
```

contracts/upgradeable_contracts/BasicRelativeDailyLimit.sol
```diff
@@ -0,0 +1,85 @@
...
+        emit TargetLimitChanged(_targetLimit);
+    }
+
+    function setThreshold(uint256 _threshold) external {
+        require(_threshold >= _minPerTx());
+        uintStorage[THRESHOLD] = _threshold;
+        emit ThresholdChanged(_threshold);
+    }
+
+    function updateTodayLimit(uint256 _balance) public {
```
patitonar: Does it makes sense to change the modifier to `external`?
```suggestion
    function updateTodayLimit(uint256 _balance) external {
```

contracts/upgradeable_contracts/AbsoluteDailyLimit.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    function minPerTx() public view returns (uint256) {
+        return uintStorage[MIN_PER_TX];
+    }
+
+    function executionMinPerTx() public view returns (uint256) {
+        return uintStorage[EXECUTION_MIN_PER_TX];
+    }
+
+    function withinLimit(uint256 _amount, uint256 _tokenBalance) public view returns (bool) {
```
patitonar: Does it makes sense to change the modifier to `external`?
```suggestion
    function withinLimit(uint256 _amount, uint256 _tokenBalance) external view returns (bool) {
```

contracts/upgradeable_contracts/AbsoluteDailyLimit.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint256 nextLimit = totalSpentPerDay(getCurrentDay()).add(_amount);
+        return dailyLimit(_tokenBalance) >= nextLimit && _amount <= maxPerTx() && _amount >= minPerTx();
+    }
+
+    function withinExecutionLimit(uint256 _amount, uint256 _tokenBalance) public view returns (bool) {
```
patitonar: Does it makes sense to change the modifier to `external`?
```suggestion
    function withinExecutionLimit(uint256 _amount, uint256 _tokenBalance) external view returns (bool) {
```

contracts/upgradeable_contracts/AbsoluteDailyLimit.sol
```diff
@@ -0,0 +1,150 @@
...
+        require(_maxPerTx < dailyLimit());
+        uintStorage[MAX_PER_TX] = _maxPerTx;
+    }
+
+    function setMinPerTx(uint256 _minPerTx) external {
+        require(_minPerTx < dailyLimit() && _minPerTx < maxPerTx());
+        uintStorage[MIN_PER_TX] = _minPerTx;
+    }
+
+    function updateTodayLimit(uint256) public {}
```
patitonar: Does it makes sense to change the modifier to `external`?
```suggestion
    function updateTodayLimit(uint256) external {}
```

contracts/upgradeable_contracts/BasicRelativeDailyLimit.sol
maxaleks: Sorry, I forgot to push the changes with contract fixes that I made

patitonar:
@maxaleks can you please resolve the merge conflicts?

maxaleks:
> @maxaleks can you please resolve the merge conflicts?

Done

deploy/src/native_to_erc/home.js
```diff
@@ -57,10 +63,19 @@ if (isRewardableBridge) {
...
   let initializeHomeBridgeData
 
+  const executionLimitsArray = RELATIVE_DAILY_LIMIT
+    ? [TARGET_LIMIT, THRESHOLD, FOREIGN_MAX_AMOUNT_PER_TX, FOREIGN_MIN_AMOUNT_PER_TX]
+    : [FOREIGN_DAILY_LIMIT, FOREIGN_MAX_AMOUNT_PER_TX, FOREIGN_MIN_AMOUNT_PER_TX]
+
+  const RELATIVE_DAILY_LIMIT_PARAMS = `TARGET_LIMIT: ${TARGET_LIMIT} which is ${
+    Web3Utils.fromWei(Web3Utils.toBN(TARGET_LIMIT).mul(Web3Utils.toBN(100)))
+  }%,
+  THRESHOLD: ${THRESHOLD} which is ${Web3Utils.fromWei(THRESHOLD)} in eth,`
```
patitonar: If `RELATIVE_DAILY_LIMIT=false` the variables `TARGET_LIMIT` and `THRESHOLD` could not be defined in `.env` file, in that case, `Web3Utils.fromWei(Web3Utils.toBN(TARGET_LIMIT)` and `Web3Utils.fromWei(THRESHOLD)` fails with an error. The same applies to the scripts of the other modes. I think a possible solution could be to only define `RELATIVE_DAILY_LIMIT_PARAMS` if `RELATIVE_DAILY_LIMIT=true`

patitonar:
@maxaleks FYI I updated the new script that verifies the contracts in the explorer as part of the deployment to include the new limit contracts https://github.com/poanetwork/tokenbridge-contracts/pull/306/commits/4aa2b668254f0e3b837210e6250bf5ba8c262f25

maxaleks:
@patitonar ok, thanks!

maxaleks:
> Can you check the merge one more time? For example, I have found that `_relayTokens` was removed from the `ForeignBridgeErcToNative.sol` by some reason.

I removed `_relayTokens` because it is duplicated in this contract ([here](https://github.com/poanetwork/tokenbridge-contracts/blob/develop/contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol#L95) and [here](https://github.com/poanetwork/tokenbridge-contracts/blob/develop/contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol#L183)). Or is this the correct logic?

maxaleks:
> Also I see that the tests were failed for the latest changes in the branch -- could you elaborate the root cause of fails?

Could you rerun the task or, even better, give me rights to do it by myself?

akolotov:
> > Can you check the merge one more time? For example, I have found that `_relayTokens` was removed from the `ForeignBridgeErcToNative.sol` by some reason.
> 
> I removed `_relayTokens` because it is duplicated in this contract ([here](https://github.com/poanetwork/tokenbridge-contracts/blob/develop/contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol#L95) and [here](https://github.com/poanetwork/tokenbridge-contracts/blob/develop/contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol#L183)). Or is this the correct logic?

They are different as far as I see:

```
95: function _relayTokens(address _sender, address _receiver, uint256 _amount) internal {
```
vs
```
183: function _relayTokens(address _sender, address _receiver, uint256 _amount, address _token) internal {
```

maxaleks:
> They are different as far as I see

Sorry, I missed it when fixing merge conflicts

PR #319, opened 2019-11-20
Add support for two tokens in erc to native

### LINKED ISSUE #318: two tokens (SCD and MCD) support
This issue is continuation of the activity started in https://github.com/poanetwork/tokenbridge-contracts/issues/310. 

As per the [migration plans](https://blog.makerdao.com/what-to-expect-with-the-launch-of-multi-collateral-dai/) provided by MakerDAO both SCD and MCD tokens will exist until the [Emergency Shutdown (ES)](https://github.com/makerdao/community/blob/master/faqs/emergency-shutdown.md) is performed for SCD. It means that during some period of time the price of both tokens will be equal and users will be able to operates with both token contracts. For the xDai bridge this period of time means that users would like to deposit either SCD or MCD to the xDai chain. **There is no any risk for such kind of deposits if the bridge will have opportunity to swap new deposited SCD tokens to MCD before SCD ES.**

The phase 2 of MCD token support introduction assumes that the xDai bridge functionality will be changed as so any deposit of **either SCD or MCD** will cause the mint action of the corresponding amount xDai coins in the xDai chain whereas withdrawal of tokens from the xDai chain will lead to unlocking the corresponding amount of **MCD tokens**.

The overall idea of supporting deposits for both tokens is the following:
  1. The TokenBridge oracle will watch for `Transfer` event emitted by SCD and MCD token contracts.
  2. If the event was raised by the MCD token contract the deposit confirmed in the same manner as now.
  3. If the event was raised by the SCD token contract, the oracle checks if SCD ES had happened to this moment. No deposit of SCD will be transferred to the xDai chain after ES. 
  4. If SCD ES did not take place, the oracle will relay the deposit to the xDai chain.
  5. Then the oracle invokes the method `swapTokens` from the bridge contract on the Ethereum Mainnet.
  6. The bridge contract will call `swapSaiToDai` of the MakerDAO migration contract to swap to MCD tokens entire SCD balance owned by the bridge contract.  

In case if tokens are being deposited by using the alternative transfer feature with calling `relayTokens` the tokens swap will be handled by this method rather than by the oracle.

In order to eliminate the possibility of drying out a validator account by deposit requests with small token values, a threshold for a minimum balance of SCD tokens will be introduced. The oracle will not call `swapTokens` if the current SCD balance of the bridge account is below this threshold. But in order to reduce the time between SCD deposit and its swap to MCD the call of `swapTokens` will be done every time when validators confirmations are executed for a withdrawal operation.

### Ethereum Mainnet side

In order to support scenarios described above the following changes are needed to be introduced in the TokenBridge contracts:
  * a new method to check the status of SCD ES
  * a new method to choose the validator to call `swapTokens` for an appeared deposit
  * a new method to store address of the SCD token contract 
  * new methods to set and check the minimal SCD balance threshold
  * the new `swapTokens` method to swap tokens as per request
  * in the `relayTokens` logic to swap tokens
  * in the `executeSignatures` logic to swap tokens
  * in the `claimTokens` logic to forbit claim the SCD tokens till SCD ES.

#### Checking status of SCD ES

MakerDAO provides a contract [`top`](https://developer.makerdao.com/dai/1/api/top) that manages the global settlement. The public integer [`caged`](https://github.com/makerdao/sai/blob/00ccfd780295e86cee54f57b6cd0673b036ae370/src/top.sol#L62) is set up to a time stamp of the block where SCD ES happens.

The address of the contract in the Ethereum Mainnet is [`0x9b0ccf7c8994e19f39b2b4cf708e0a7df65fa8a3`](https://etherscan.io/address/0x9b0ccf7c8994e19f39b2b4cf708e0a7df65fa8a3#readContract). The variable `caged` is zero before ES.

It means that the next method introduced in the bridge contract will return `true` if the SCD ES has not been applied to the moment of the time stamp provided as the method parameter.

```solidity
function isTokenSwapAllowed(uint256 _ts) public view returns(bool) {
    uint256 es_ts = ISaiTop(0x9b0ccf7C8994E19F39b2B4CF708e0A7DF65fA8a3).caged();
    if ((es_ts > 0) && (_ts > es_ts)) {
        return false;
    }
    return true;
}
```

This method is for `contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol`.

#### Validator selection

The idea to select the validator to trigger swap for the corresponding deposit of SCD is the following:
  * the validator's node calls the method `isValidatorDuty` defined below
  * this method uses the latest block number and the number of the validator in the validatators list to determine if this validator is on duty to perform some actions
  * if the node receives `true` from the method `isValidatorDuty` it proceeds with the actions to swap SCD tokens

```solidity
function isValidatorDuty(address _validator) external view returns(bool) {
    uint256 counter = 0;
    address next = getNextValidator(F_ADDR);
    require(next != address(0));

    while (next != F_ADDR) {
        if (next == _validator) {
            return (block.number % validatorCount() == counter);
        }
        
        next = getNextValidator(next);
        counter++;

        require(next != address(0));
    }

    return false;
}
```

The method should be incorporated into the validators contract.

Since it will be more than one way to swap SCD tokens during the bridge operations it is safe if the validator is not chosen accidentally by this approach. The balance of SCD tokens could be swapped later when another deposit request is handled or during execution of a withdrawal request. 

#### Storing SCD token address

The current version of the bridge contract on the foreign side stores the bridgeable token contract address and provides it through the public method `erc20token()`. This method is invoked by the oracle, UI and monitor as so these TokenBridge components could interact with the token contract without necessity to point out its address in the corresponding component configuration.

After the Phase 2 migration the ability to get both SCD and MCD tokens contract addresses from the bridge contract is required.

The new method `halfduplexErc20token()` will returns the address of SCD contract.

```solidity
function halfduplexErc20token() public pure returns(ERC20Basic) {
    return(ERC20Basic(0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359));
}
```

This method will be located in `contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol`.

#### Minimal SCD balance threshold

Since the new bridge parameter is introduced, new setter and getter methods must be provided in `contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol`:

```solidity
function setMinHDTokenBalance(uint256 _minBalance) external onlyOwner {
    uintStorage[MIN_HDTOKEN_BALANCE] = _minBalance;
}

function minHDTokenBalance() external view returns(uint256) {
    return uintStorage[MIN_HDTOKEN_BALANCE];
}
```

Also it makes sense to provide a method that will check if the SCD token balance is above the threshold as atomic operation:

```solidity
function isHDTokenBalanceAboveMinBalance() public view returns(bool) {
    if (halfduplexErc20token().balanceOf(address(this)) > uintStorage[MIN_HDTOKEN_BALANCE]) {
        return true;
    }
    return false;
}
```

#### Swap tokens operation

The main method that is introduced by the Phase 2 migration allows anyone to swap SCD tokens owned by the bridge contract to MCD. **The method will be available until SCD ES is executed.** As it was described above the bridge validators are responsible for calling this method as soon as new deposit request of SCD tokens detected.

```solidity
function swapTokens() public {
    require(isTokenSwapAllowed(now));

    address migrationContract = 0xc73e0383F3Aff3215E6f04B0331D58CeCf0Ab849; //as per https://changelog.makerdao.com/releases/mainnet/1.0.0/contracts.json

    uint256 curHDTokenBalance = halfduplexErc20token().balanceOf(address(this));
    require(curHDTokenBalance > 0);

    uint256 curFDTokenBalance = erc20token().balanceOf(address(this));

    require(halfduplexErc20token().approve(migrationContract, curHDTokenBalance));
    IScdMcdMigration(migrationContract).swapSaiToDai(curHDTokenBalance);

    require(erc20token().balanceOf(address(this)) - curFDTokenBalance == curHDTokenBalance);

    emit TokensSwapped(halfduplexErc20token(), erc20token(), curHDTokenBalance);
}
```

The migration contract is explicitly specified in the contract since it is assumed that the changes prepared for the Phase 2 are specific for particular pair of tokens.

#### Interop with alternative receiver feature

Since the scenario to use the alternative receiver feature directly involves the bridge contract to deposit the tokens it allows to swap SCD tokens to MCD in the same transaction.

The current implementation of the feature assumes that only one token contract exists for the bridge operations. Support of two tokens will require extension of existing interfaces as so the contract that deposits tokens by specifying an alternative receiver will need to specify the specific token contract address. But in order to make behavior compatible with other bridge modes four ways to relay tokens will be supported:
  1. `relayTokens(address _receiver, uint256 _amount)`
  2. `relayTokens(address _from, address _receiver, uint256 _amount)`
  3. `relayTokens(address _receiver, uint256 _amount, address _token)`
  4. `relayTokens(address _from, address _receiver, uint256 _amount, address _token)`

If the token contract address is not specified (the cases 1 and 2) or if it is zero (for the cases 3 and 4), the MCD token contract is used.

Bearing in mind described above, the following changes for the method defined in `contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol` are suggested: 

```solidity
function _relayTokens(address _sender, address _receiver, uint256 _amount, address _token) internal {
    require(_receiver != bridgeContractOnOtherSide());
    require(_receiver != address(0));
    require(_receiver != address(this));
    require(_amount > 0);
    ERC20Basic tokenToOperate = ERC20Basic(_token);
    ERC20Basic FDToken = erc20token();
    ERC20Basic HDToken = halfduplexErc20token();
    require(tokenToOperate == FDToken || 
            tokenToOperate == ERC20Basic(0x0) || 
            tokenToOperate == HDToken);
            
    require(withinLimit(_amount));
    setTotalSpentPerDay(getCurrentDay(), totalSpentPerDay(getCurrentDay()).add(_amount));

    if (tokenToOperate == ERC20Basic(0x0)) {
        tokenToOperate = FDToken;
    }
    
    tokenToOperate.transferFrom(_sender, address(this), _amount);
    emit UserRequestForAffirmation(_receiver, _amount);
    if (tokenToOperate == HDToken) {
        swapTokens();
    }
}
```

#### Implicit tokens swap

The current bridge architecture allows to implement the implicit token swap operation which will handle non-zero balance of the SCD tokens. This balance could appear if a SCD deposit occurred with value less or equal the minimal balance threshold. Another reason of non-zero balance is the case when the bridge validator skips his duty to call the `swapTokens()` method.

The implicit token swap is integrated in the flow when the method `executeSignatures` is being executed:

```solidity
function onExecuteMessage(
    address _recipient,
    uint256 _amount,
    bytes32 /*_txHash*/
) internal returns (bool) {
    uint256 amount = _amount.div(10**decimalShift());
    uint256 amount = _amount;
    bool res = erc20token().transfer(_recipient, amount);
    
    if (halfduplexErc20token().balanceOf(address(this)) > 0) {
        address(this).call(abi.encodeWithSelector(bytes4(keccak256(bytes("swapTokens()")))));
    }

    return res;
}
``` 

#### Limited ability to claim SCD tokens

The TokenBridge contracts provides ability for the bridge owners to claim tokens (native or ERC20) owned by the bridge contract if they were occasionally transferred to the contract. Any token except the bridgeable one could be claimed and sent to the initial originator.

As per considered design the SCD token is bridgeable until SCD ES is executed. So, the ability to claim SCD tokens must be restricted till the SCD ES execution as well.

For this, the method `claimTokens` of `contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol` should be extended like this:

```solidity
function claimTokens(address _token, address _to) public {
    require(_token != address(erc20token()));
    if (_token == address(halfduplexErc20token())) {
        // SCD is not claimable if the bridge accepts deposits of this token
        require(!isTokenSwapAllowed(now));
    }
    super.claimTokens(_token, _to);
}
```

### xDai chain side

Since the only bridge operation that is possible with the SCD token is the deposit no changes are needed on the xDai chain side.

patitonar:
Closes #318


contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -71,7 +79,13 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
     ) internal returns (bool) {
         setTotalExecutedPerDay(getCurrentDay(), totalExecutedPerDay(getCurrentDay()).add(_amount));
         uint256 amount = _amount.div(10**decimalShift());
-        return erc20token().transfer(_recipient, amount);
+        bool res = erc20token().transfer(_recipient, amount);
+
+        if (AddressUtils.isContract(halfDuplexErc20token()) && tokenBalance(halfDuplexErc20token()) > 0) {
```
akolotov: do you think it make sense to check if it is a contract for the case when we hardcode the return value of the method `halfDuplexErc20token`? When we hardcode the value we know that it is the token contract.

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -102,4 +116,100 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
+            return true;
+        }
+        return false;
+    }
+
+    function tokenBalance(ERC20 _token) internal view returns (uint256) {
+        return _token.balanceOf(address(this));
+    }
+
+    function migrationContract() internal pure returns (IScdMcdMigration) {
```
akolotov: we don't need to specify the address of the migration contract in `migrateToMCD` anymore, do we?

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -102,4 +116,100 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
+        return IScdMcdMigration(0xc73e0383F3Aff3215E6f04B0331D58CeCf0Ab849);
+    }
+
+    function swapTokens() public {
```
akolotov: Can this method be used in `migrateToMCD`?

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -102,4 +116,100 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
+            return false;
+        }
+        return true;
+    }
+
+    function halfDuplexErc20token() public pure returns (ERC20) {
+        return ERC20(0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359);
+    }
+
+    function setMinHDTokenBalance(uint256 _minBalance) external onlyOwner {
```
akolotov: Let's set some initial value (e.g. `10 ether`) as part of the `migrateToMCD` call

patitonar:
Tested the implementation with [MakerDAO contracts in Kovan testnet](https://changelog.makerdao.com/releases/kovan/0.2.17/index.html) which are identical to the deployed in mainnet.

Sai Address: 0xc4375b7de8af5a38a93548eb8453a498222c4ff2
Dai Address: 0x4f96fe3b7a6cf9725f59d353f723c1bdb64ca6aa
ScdMcdMigration Address: 0x411b2faa662c8e3e5cf8f01dfdae0aee482ca7b0
Sai Top: 0x5f00393547561da3030ebf30e52f5dc0d5d3362c (from https://developer.makerdao.com/dai/1/api/)

Foreign Bridge Address: 0xb644F2c13D1e2410d224825991455fCBE95ACcd9

List of tx examples
- migrateToMCD [0x1470e2434ea73e3b4fcb10a8007b2f3def9e107746c9f615d9aea86dd3774c7f](https://blockscout.com/eth/kovan/tx/0x1470e2434ea73e3b4fcb10a8007b2f3def9e107746c9f615d9aea86dd3774c7f)
- swapTokens [0x3e2b6b891aec411f378cf60fc04ebe6554f19ec14827ccc9d559e1ae8b183e02](https://blockscout.com/eth/kovan/tx/0x3e2b6b891aec411f378cf60fc04ebe6554f19ec14827ccc9d559e1ae8b183e02)
- relayTokens sai [0x34ecac9a85cc1d44b86e22d70c2906294ab08d4f124073eb9296c4c904cd8eee](https://blockscout.com/eth/kovan/tx/0x34ecac9a85cc1d44b86e22d70c2906294ab08d4f124073eb9296c4c904cd8eee)
- relayTokens dai [0x1b1c1141ec38ba9e9321e05c91f34a9d7e610812954bf75ff5f868095484a4ee](https://blockscout.com/eth/kovan/tx/0x1b1c1141ec38ba9e9321e05c91f34a9d7e610812954bf75ff5f868095484a4ee)

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
patitonar: Thanks, fixed https://github.com/poanetwork/tokenbridge-contracts/pull/319/commits/360dcaaafefa2069f363becae01c3a71bf4a3a8c

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
patitonar: Updated `migrateToMCD` to call `swapTokens`

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
patitonar: Added initial value https://github.com/poanetwork/tokenbridge-contracts/pull/319/commits/360dcaaafefa2069f363becae01c3a71bf4a3a8c

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
patitonar: You are correct, the check is not needed. Fixed https://github.com/poanetwork/tokenbridge-contracts/pull/319/commits/055bcb62a794ed913dfec10b6558e3b57a314151

akolotov:
> Tested the implementation with [MakerDAO contracts in Kovan testnet](https://changelog.makerdao.com/releases/kovan/0.2.17/index.html) which are identical to the deployed in mainnet.

thanks!

contracts/upgradeable_contracts/BaseBridgeValidators.sol
```diff
@@ -102,4 +102,23 @@ contract BaseBridgeValidators is InitializableBridge, Ownable {
     function setNextValidator(address _prevValidator, address _validator) internal {
         addressStorage[keccak256(abi.encodePacked("validatorsList", _prevValidator))] = _validator;
     }
+
+    function isValidatorDuty(address _validator) external view returns (bool) {
```
akolotov: should we increase a version in `getBridgeValidatorsInterfacesVersion`?

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -83,23 +97,113 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
+
+    function tokenBalance(ERC20 _token) internal view returns (uint256) {
+        return _token.balanceOf(address(this));
+    }
+
+    function migrationContract() internal pure returns (IScdMcdMigration) {
+        return IScdMcdMigration(0xc73e0383F3Aff3215E6f04B0331D58CeCf0Ab849);
+    }
+
+    function swapTokens() public {
```
akolotov: should we increase the version returned by `getBridgeInterfacesVersion`?

contracts/upgradeable_contracts/BaseBridgeValidators.sol
patitonar: I think we should. Updated version to `2.3.0` https://github.com/poanetwork/tokenbridge-contracts/pull/319/commits/be3cd335b2bf2f4077759dfc0996e52e00b367a7

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
patitonar: Yes, it makes sense. Updated to `2.5.0` https://github.com/poanetwork/tokenbridge-contracts/pull/319/commits/c2c2c4cc2b5308b1fbe5a7bdd872d5238e297636

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -83,23 +97,113 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
         super._relayTokens(_sender, _receiver, _amount);
     }
 
-    function migrateToMCD(address _migrationContract) external onlyOwner {
+    function migrateToMCD() external onlyOwner {
```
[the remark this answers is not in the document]
patitonar: Updated the method https://github.com/poanetwork/tokenbridge-contracts/pull/319/commits/251f9495578ca63202bc731068dc8064c04fdded

patitonar:
Tested the upgrade with full changes following the steps mentioned [here](https://hackmd.io/@1oqceZGlR6WSWAv2TTWJeA/Bk76GRPpS). Everything worked as expected and no issues were found.

Here are the new contract implementations:
* **[xDai] HomeBridgeErcToNative**: no migration method required. Gist: https://gist.github.com/patitonar/a19287408dfa878d7be1dca95835625f
* **[mainnet] BridgeValidators**: added migration method to create the list of validators. Gist: https://gist.github.com/patitonar/fe4888e786e9b6cdbfaeedaf8fe99b78
```javascript
function upgradeToV230() public {
    bytes32 upgradeStorage = 0x2e8a0420c66b3c19df2ac1cd4ffadb5637da547c6fd49388c7fe28dc8c13a8dd; // keccak256(abi.encodePacked('isUpgradedToV230'))
    require(!boolStorage[upgradeStorage]);
    address validator1 = 0x9ad83402C19Af24F76afa1930a2B2EEC2F47A8C5;
    address validator2 = 0x4D1c96B9A49C4469A0b720a22b74b034EDdFe051;
    address validator3 = 0x491FC792e78CDadd7D31446Bb7bDDef876a69AD6;
    address validator4 = 0xc073C8E5ED9Aa11CF6776C69b3e13b259Ba9F506;

    setNextValidator(F_ADDR, validator1);
    setNextValidator(validator1, validator2);
    setNextValidator(validator2, validator3);
    setNextValidator(validator3, validator4);
    setNextValidator(validator4, F_ADDR);
    boolStorage[upgradeStorage] = true;
}
```
* **[mainnet] ForeignBridgeErcToNative**: added migration method for missing parameters. Gist: https://gist.github.com/patitonar/c0b77a59eceb76aad05e010c40569d47
```javascript
function upgradeToV250() public {
    bytes32 upgradeStorage = 0x221181d03d3c423368b412861518cd98276a115f4c2d9594ab6697d90fc99c19; // keccak256(abi.encodePacked('isUpgradedToV250'))
    require(!boolStorage[upgradeStorage]);
    uintStorage[DAILY_LIMIT] = 100000 ether;
    emit DailyLimitChanged(100000 ether);
    uintStorage[MIN_PER_TX] = 5000000000000000;
    _setBridgeContractOnOtherSide(0x7301CFA0e1756B71869E93d4e4Dca5c7d0eb0AA6);
    boolStorage[upgradeStorage] = true;
}
```

I tested all steps using the strategy option 1 on Sokol and Kovan, using the set of contracts provided by makerDAO on Kovan. Except for step `9. Test Emergency Shutdown` which was not possible to generate such a scenario.

I also tested upgrading in foreign side `BridgeValidators` and `ForeignBridgeErcToNative` with option 2 (using `ganache-cli` to fork the networks). I was able to deploy the new implementations and upgrade the contracts. I checked the state was not lost and new variables were correctly set. But I wasn't able to test the ui, oracle, and monitor in this case, because the forked local network generated by ganache was slow and sometimes it just stopped working when received a lot of requests, especially when trying to get events. Also, I wasn't able to deploy the new implementation for `HomeBridgeErcToNative` in the forked xDai, probably because of some issues with `ganache-cli`. I tested deploying the same contract directly in xDai and it worked correctly there.

PR #328, opened 2019-12-09
Possibility of increasing/decreasing allowance in ERC677

### LINKED ISSUE #325: Add increase/decrease allowance methods to the token contracts
As per the security audit it was found that token contracts supplied with the tokenbridge contracts source code does not have `increaseAllowance` and `decreaseAllowance` implemented that [the reference implementation of ERC20 token](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol). Rationale of these methods is described [here](https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729).

k1rill-fedoseev:
Closes #325


akolotov:
Will be merged just after #319.

PR #329, opened 2019-12-09
Insert missing variables into the example configuration

ArseniiPetrovich:
Contracts deployment would not start if `HOME_REWARDABLE` or `FOREIGN_REWARDABLE` is not defined, however, they are not defined at ERC_TO_ERC example config.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -284,6 +284,9 @@ REQUIRED_NUMBER_OF_VALIDATORS=1
 # correctly to the Foreign network.
 VALIDATORS=0x 0x 0x
 
+HOME_REWARDABLE=false
+FOREIGN_REWARDABLE=false
+
 # The flag defining whether to use ERC677BridgeTokenRewardable contract instead of
 # ERC677BridgeToken.
 DEPLOY_REWARDABLE_TOKEN=false


deploy/README.md
```diff
@@ -284,6 +284,9 @@ REQUIRED_NUMBER_OF_VALIDATORS=1
 # correctly to the Foreign network.
 VALIDATORS=0x 0x 0x
 
+HOME_REWARDABLE=false
```
akolotov: Please add descriptions how it is done for these variables in other sections.

deploy/README.md
```diff
@@ -284,6 +284,9 @@ REQUIRED_NUMBER_OF_VALIDATORS=1
...
+FOREIGN_REWARDABLE=false
+
 # The flag defining whether to use ERC677BridgeTokenRewardable contract instead of
 # ERC677BridgeToken.
 DEPLOY_REWARDABLE_TOKEN=false
```
akolotov: Does it make sense to comment `DPOS_STAKING_ADDRESS ` and `DPOS_STAKING_ADDRESS ` in order to not allow user to make a mistake by initialising them in some value even if a rewardable token is not used.

ArseniiPetrovich: If these values are not required - it is a good idea to comment them out, from my point of view.

deploy/README.md
ArseniiPetrovich: Which modes of `REWARDABLE` are supported in this bridge mode?

patitonar: In ERC_TO_ERC mode, for `HOME_REWARDABLE` variable only  `BOTH_DIRECTIONS` is supported. For `FOREIGN_REWARDABLE` variable collecting fees on Foreign network is not supported

deploy/README.md
akolotov: yes, they are not required for the generic `erc20-to-erc20` bridge

PR #330, opened 2019-12-12
Updated checks in setMinPerTx and setMaxPerTx

### LINKED ISSUE #322: checks in setMinPerTx and setMaxPerTx inconsistent with initialization checks
As part of a new security audit it was found that checks used in `setMinPerTx` and `setMaxPerTx` are different from the checks used during intialization procedure.

For example,

https://github.com/poanetwork/tokenbridge-contracts/blob/20d262702defd3aed39870f3379f1cf0c314305e/contracts/upgradeable_contracts/BasicTokenBridge.sol#L92-L100

and 

https://github.com/poanetwork/tokenbridge-contracts/blob/20d262702defd3aed39870f3379f1cf0c314305e/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol#L135-L139

as it is presented above `setMaxPerTx` checks that the new value is less than `dailyLimit`, `setMinPerTx` checks that the new value is less than `maxPerTx`. At the same time the initialization sequence additionally ensures that `maxPerTx` value is greater than `minPerTx` and `minPerTx` is greater than `0`.

Consider to change `setMaxPerTx` and `setMinPerTx` to have checks consistent with the initialization sequence. 

`setMaxPerTx` should also allow to set the value to `0` in order to have two ways to stop the bridge operations by both `setDailyLimit` and `swtMaxPerTx` (it will be used later in the _Relative Daily Limit_ feature.)

k1rill-fedoseev:
Closes #322 
`setMaxPerTx` now allows to set `0` limit in order to temporarily stop the bridge. Otherwise, if not `0`, new max limit should be greater than `minPerTx()`.
`setMinPerTx` additionally checks that new limit is positive.

### DIFF
--- a/contracts/upgradeable_contracts/BasicTokenBridge.sol
+++ b/contracts/upgradeable_contracts/BasicTokenBridge.sol
@@ -90,12 +90,12 @@ contract BasicTokenBridge is EternalStorage, Ownable {
     }
 
     function setMaxPerTx(uint256 _maxPerTx) external onlyOwner {
-        require(_maxPerTx < dailyLimit());
+        require(_maxPerTx == 0 || (_maxPerTx > minPerTx() && _maxPerTx < dailyLimit()));
         uintStorage[MAX_PER_TX] = _maxPerTx;
     }
 
     function setMinPerTx(uint256 _minPerTx) external onlyOwner {
-        require(_minPerTx < dailyLimit() && _minPerTx < maxPerTx());
+        require(_minPerTx > 0 && _minPerTx < dailyLimit() && _minPerTx < maxPerTx());
         uintStorage[MIN_PER_TX] = _minPerTx;
     }
 }
--- a/test/erc_to_native/home_bridge.test.js
+++ b/test/erc_to_native/home_bridge.test.js
@@ -845,6 +845,13 @@ contract('HomeBridge_ERC20_to_Native', async accounts => {
       minPerTx.should.be.bignumber.equal(toBN(1))
     })
 
+    it('setMaxPerTx allows to set limit to zero', async () => {
+      await homeContract.setMaxPerTx(0, { from: owner }).should.be.fulfilled
+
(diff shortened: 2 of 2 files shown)


contracts/upgradeable_contracts/BasicTokenBridge.sol
```diff
@@ -90,12 +90,12 @@ contract BasicTokenBridge is EternalStorage, Ownable {
...
 
     function setMaxPerTx(uint256 _maxPerTx) external onlyOwner {
-        require(_maxPerTx < dailyLimit());
+        require(_maxPerTx == 0 || (_maxPerTx > minPerTx() && _maxPerTx < dailyLimit()));
         uintStorage[MAX_PER_TX] = _maxPerTx;
     }
 
     function setMinPerTx(uint256 _minPerTx) external onlyOwner {
-        require(_minPerTx < dailyLimit() && _minPerTx < maxPerTx());
+        require(_minPerTx > 0 && _minPerTx < dailyLimit() && _minPerTx < maxPerTx());
```
patitonar: Given that the limits are consistently set by initialization method, `setMaxPerTx` and `setDailyLimit`, is it ensured that `maxPerTx() < dailyLimit()`. Does it makes sense to only check that `_minPerTx < maxPerTx()`?
```suggestion
        require(_minPerTx > 0 && _minPerTx < maxPerTx());
```

k1rill-fedoseev: If we make it so, then I guess that we will be able to change `minPerTx` limit when the bridge is stopped by `setDailyLimit(0)`, while at the same time, we won't be able to change it when the bridge is stopped by `setMaxPerTx(0)`.
Should it behave like this?

patitonar: I see. I think we should keep the implementation as it is for consistency, so no other limits can be updated when the bridge is stopped. Thanks

PR #332, opened 2019-12-12
Include upgradeability contracts as a dependency

### LINKED ISSUE #324: Use OpenZeppelin eternal storage contracts as dependency
As part of the security audit the following finding was raised:

**Description**: The clone-and-own approach involves copying and adjusting open source code at one's own discretion. From the development perspective, it is initially beneficial as it reduces the amount of effort. However, from the security perspective, it involves some risks as the code may not follow the best practices, may contain a security vulnerability, or may include intentionally or unintentionally modified upstream libraries. Rather than the clone- and-own approach, a good industry practice is to use the Truffle framework for managing library dependencies. This eliminates the clone-and-own risks yet allows for following best practices, such as, using libraries.

**Recommendation**: Since upgradeability clones most of the behavior in `https://github.com/OpenZeppelin/openzeppelin- labs/tree/master/upgradeability_using_eternal_storage`, we recommend using the latter as a dependency, rather than cloning its code.

Since the tokenbridge contracts are using a customised version of the sources mentioned above, research and implement a way how to use the original code as dependency for the bridge.

k1rill-fedoseev:
Closes #324 
Now, using `openzeppelin-labs` repository as a dependency. 
Check for code size from local copy was moved to `EternalStorageProxy.sol`.
Implementation version is saved as a `string` (not as a `uint256`, as it was in local copy).


akolotov:
> What do you think @akolotov?

I agree with you @patitonar. That's why the purpose of issue is `research and implement a way how to use the original code`. For example, one of the research result could be to suggest leave everything as is but with code comments in the contracts what was changed in comparison with the original version -- these comments will be used as explanations for further security audits.

Another possible research result could be a recommendation to replace only few of contracts by the original ones whereas the rest could be left modified. Again, the reasoning must be added to the code comments.

Or we could modify the contracts on the fly as part of the compilation process. And of course, this is the most undesired solution since it is very hard to maintain the further modifications in this case.

akolotov:
@patitonar please take a look. If you are OK we will merge the PR and will start a procedure to tag a new release.

PR #334, opened 2019-12-16
Add automatic contracts verification in deployment script

### LINKED ISSUE #193: Automatic contracts verification in Blockscout/Etherescan
Associating the contract source code with the contract in the chain is not trivial procedure. Especially since the proxy template is used for the bridge functionality. 

Consider to user [the BlockScout API](https://blockscout.com/eth/kovan/api_docs) to verify the deployed contracts. 

New parameters like `HOME_BLOCKSCOUT_URL` and `FOREIGN_BLOCKSCOUT_URL` could be added to the `.env` file. If they are configured and the URL is accessible, the deployment script will try to verify every deployed contract as soon as the deployment transaction returns success.

patitonar:
Closes #193

Support Blockscout and Etherscan explorers. In the case of Etherscan, an API KEY is also needed besides the URL.

For that I added 4 new parameters:
* `HOME_EXPLORER_URL`
* `HOME_EXPLORER_API_KEY`
* `FOREIGN_EXPLORER_URL`
* `FOREIGN_EXPLORER_API_KEY`

In the deployment script, every time a contract is deployed, it will try to verify the source code in the explorer if the URL is provided.
The `verifier` component will get most of the information (name, solidity version, evm version, optimizer) from the truffle-artifact provided by parameter and also will use `truffle-flattener` to generate the flat version. It uses `promise-retry` to retry verifying the contract a couple of times in the case that the explorer didn't index the contract fast enough.


deploy/src/utils/verifier.js
```diff
@@ -0,0 +1,102 @@
+const axios = require('axios')
+const querystring = require('querystring')
+const flattener = require('truffle-flattener')
+const { EXPLORER_TYPES, REQUEST_STATUS } = require('../constants')
+const promiseRetry = require('promise-retry')
+
+const flat = async contractPath => flattener([contractPath], process.cwd())
```
akolotov: do we need to flatten contracts one more time? We already have them flattened in `flats` directory. The idea is to use files from `flats` in order to make sure that deployed contracts can be built from them. My plan is to provide content of `flats` directory as a release artefacts.

patitonar: I thought about using the `flats` directory but here are some reasons why I chose to not use it:
- It will require that the `flatten` script is run before the deployment script, and it should be up-to-date in case of adding new contracts. For this, we can create a `deploy` npm script instead of running `node deploy.js` and call first the `flatten` script to make sure they are created. 
- In the implementation, we need to make a relation between the deployed contract and the flatten location. The truffle artifact provides the location of the source code and the script uses it to call truffle-flattener for that contract. In this case, I think we can calculate the location of the flatten contract from the source location, but it will be prone to break if any directory is changed

akolotov: Thanks for analysis!

> It will require that the flatten script is run before the deployment script

I have no objections for this.

patitonar: Updated the logic to use `flats` directory https://github.com/poanetwork/tokenbridge-contracts/pull/334/commits/04bb5785a6451fd09daa05d82e02de388dd2bc92

deploy/package.json
```diff
@@ -5,7 +5,8 @@
   "main": "deploy.js",
   "scripts": {
     "compile": "pkg deploy.js --out-path build/tmp_build",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "deploy": "cd .. && npm run flatten && cd deploy && node deploy.js"
```
akolotov: could you consider to improve it by a way like this:
```
if [ ! "${NOFLAT}" == "true" ]; then cd ..; npm run flatten; cd deploy; fi && node deploy.js
```
and update the Docker file with `ENV NOFLAT "true"` ?

the rationale is the following: if the deployment is run within the docker container there is no need to flat contracts since they are already flattened. In most cases users will not re-compile contracts within the container so it is safe to assume that `flats` directory will not be updated without update entire docker image.

deploy/src/utils/verifier.js
akolotov: thanks

deploy/package.json
patitonar: Updated the script and Dockerfile https://github.com/poanetwork/tokenbridge-contracts/pull/334/commits/e3ef705908a29e0ef5a67a12d30bb58e633d9fcc

akolotov: cool!

PR #337, opened 2019-12-17
Add upgrade scripts

### LINKED ISSUE #336: Simplify the upgrade to new version of contracts by introduction of dedicated scripts
Currently upgrade of bridge contracts is not trivial (https://docs.tokenbridge.net/xdai-bridge/xdai-bridge-contracts-management) especially with the case when a multisig wallet is used as a management interface for bridge owners.

Consider to implement scripts that will simplify the bridge owners actions. Particularly this issue is to help with upgrade to the new contract introduced under #318.

As per [testing suit](https://github.com/poanetwork/tokenbridge-contracts/pull/319) validators need to perform 4 steps:
  * call `upgradeToAndCall` from the Validators contract on the Foreign side.
  * call `upgradeTo` from the bridge contract on the Home side
  * call `upgradeToAndCall`  from the bridge contract on the Foreign side
  * call `migrateToMCD`  from the bridge contract on the Foreign side

Every such call requires one validator to initialize the transaction on the multisig wallet and confirmations from other validators.
So, there would be 4 scripts for every call. Each script must support two parameters `init` and `confirm`. The scripts should be located in a new `upgrade` directory.

The scripts should be able to get transactionId for confirmations automatically. In order to achieve this, the suggestion is to have a start block defined in the config files.
So, then the leader will call first script (e.g. it could be called `upgradeValidatorsContractOnForeign`) with the `init` parameter. As soon the leader gets the result for the transaction as successfully mined (exit status of the script?). They notifies other validators to send confirmations. Others call `upgradeValidatorsContractOnForeign confirm`.
The script will look for `Submissions` events in all the blocks starting from the configured one. Then the script checks that the corresponding transaction is for the `upgradeToAndCall` call and there is no corresponding `Execution`. If it is true the confirmation is sent. If the submissions were not found or other conditions were not met the script will report the corresponding error.

So, the configuration file will be similar to:
```
HOME_RPC_URL=
HOME_PRIVKEY=
HOME_BRIDGE_ADDRESS=
HOME_START_BLOCK=

FOREIGN_RPC_URL=
FOREIGN_PRIVKEY=
FOREING_BRIDGE_ADDRESS=
FOREIGN_START_BLOCK=
```

It assumes that all other information could be extracted from the bridge contracts.

patitonar:
Closes #336 

Besides the mentioned variable for the scripts, I added `HOME_GAS_PRICE`,  `FOREIGN_GAS_PRICE` and `ROLE`. The last one can be set to `leader` or `confirm`. If it is set to `leader` it will create the transactions on the multisig wallet when any of the scripts are run, otherwise, it will look for the submissions and send the confirmation

The scripts assume the new implementations contracts are deployed and the addresses are available in `upgrade/src/constants.js`. Right now, they are set to the current implementation addresses, **once we have the addresses of the new implementations contracts we should update it before merging this.**


upgrade/src/constants.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
akolotov: Do you think it makes sense to move all these parameters to `.env` file as well? Sorry that I forgot to mention this in the issue description.

upgrade/package.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "upgrade-xdai",
+    "version": "1.0.0",
+    "main": "index.js",
+    "license": "MIT",
```
akolotov: the license of the entire repo is GPL-3. Do you think the license of the scripts should differ?

upgrade/README.md
```diff
@@ -0,0 +1,26 @@
...
+Install dependencies in root project and compile the contracts
+```bash
+cd ..
+npm i
+npm run compile
+```
+
+Install dependencies from `upgrade` folder
+```bash
+npm i
```
akolotov: should go the `upgrade` directory first, right?

upgrade/src/migrateToMCD.js
```diff
@@ -0,0 +1,53 @@
...
+  FOREIGN_PRIVKEY,
+  FOREIGN_RPC_URL,
+  FOREING_BRIDGE_ADDRESS,
+  ROLE,
+  FOREIGN_START_BLOCK,
+  FOREIGN_GAS_PRICE
+} = process.env
+
+const web3 = new Web3(new Web3.providers.HttpProvider(FOREIGN_RPC_URL))
+const { address } = web3.eth.accounts.wallet.add(FOREIGN_PRIVKEY)
```
akolotov: May we output the address of this account and check that its balance is not zero?

upgrade/src/migrateToMCD.js
```diff
@@ -0,0 +1,53 @@
...
+
+const migrateToMCD = async () => {
+  try {
+    const proxy = new web3.eth.Contract(proxyAbi, FOREING_BRIDGE_ADDRESS)
+    const bridge = new web3.eth.Contract(foreignBridgeAbi, FOREING_BRIDGE_ADDRESS)
+
+    const ownerAddress = await proxy.methods.upgradeabilityOwner().call()
+    const multiSigWallet = new web3.eth.Contract(multiSigWalletAbi, ownerAddress)
```
akolotov: is it worth to check that the account that is going to send a transaction participates in the multisig wallet?

upgrade/src/upgradeBridgeOnForeign.js
```diff
@@ -0,0 +1,51 @@
...
+
+const web3 = new Web3(new Web3.providers.HttpProvider(FOREIGN_RPC_URL))
+const { address } = web3.eth.accounts.wallet.add(FOREIGN_PRIVKEY)
+
+const upgradeBridgeOnForeign = async () => {
+  try {
+    const proxy = new web3.eth.Contract(proxyAbi, FOREING_BRIDGE_ADDRESS)
+    const ownerAddress = await proxy.methods.upgradeabilityOwner().call()
+    const multiSigWallet = new web3.eth.Contract(multiSigWalletAbi, ownerAddress)
+    // 0x46016a67 = upgradeToV250()
```
akolotov: you use `bridge.methods.migrateToMCD().encodeABI()` in `migrateToMCD.js`. Do you think it is worth to use similar approach here?

upgrade/src/migrateToMCD.js
```diff
@@ -0,0 +1,53 @@
...
+    const data = bridge.methods.migrateToMCD().encodeABI()
+
+    if (ROLE === 'leader') {
+      const gas = await multiSigWallet.methods
+        .submitTransaction(FOREING_BRIDGE_ADDRESS, 0, data)
+        .estimateGas({ from: address })
+      const receipt = await multiSigWallet.methods
+        .submitTransaction(FOREING_BRIDGE_ADDRESS, 0, data)
+        .send({ from: address, gas, gasPrice: FOREIGN_GAS_PRICE })
+      console.log(receipt)
```
akolotov: May we highlight explicitly if the status of the transaction is zero?

upgrade/src/confirmTransaction.js
```diff
@@ -0,0 +1,39 @@
...
+
+  if (filteredSubmission.length > 1) {
+    throw new Error('More than one transaction found.')
+  } else if (filteredSubmission.length === 0) {
+    throw new Error('No transaction to confirm.')
+  }
+  const { transactionId } = filteredSubmission[0].returnValues
+
+  const gas = await contract.methods.confirmTransaction(transactionId).estimateGas({ from: address })
+  const receipt = await contract.methods.confirmTransaction(transactionId).send({ from: address, gas, gasPrice })
```
akolotov: Imagine that all validators performs a particular confirmation at the same time. The last confirmation will consume more gas than it was estimated since the transaction will not only change the state of the multisig wallet contract but also will invoke the method in a destination contract. That's why in the instructions (e.g. https://docs.tokenbridge.net/xdai-bridge/xdai-bridge-contracts-management/upgrade-contracts, the step 6) suggest to multiply the gas usage parameter for the confirmation transaction. Is it worth to reflect this behaviour here?

upgrade/src/constants.js
patitonar: Sure, I moved them to `.env` file https://github.com/poanetwork/tokenbridge-contracts/pull/337/commits/f1aba7ffbd778ad927c731b90a568dfb14e2a11a

upgrade/package.json
patitonar: Sorry, this was autogenerated. Fixed https://github.com/poanetwork/tokenbridge-contracts/pull/337/commits/30b4fa909200ab60185ca11f591fa3324d3c5a6c

upgrade/README.md
patitonar: Thanks, fixed https://github.com/poanetwork/tokenbridge-contracts/pull/337/commits/654048c037eed41f6bd689d2dc44099ca00ef7ef

upgrade/src/migrateToMCD.js
patitonar: Added the file `upgrade/src/validatorState.js` that log the address of the validator and also check if it is not zero

upgrade/src/migrateToMCD.js
patitonar: Added the file `upgrade/src/validatorState.js` that check that the validator is part of the multisig wallet

upgrade/src/upgradeBridgeOnForeign.js
patitonar: Yes, it makes sense to use a similar approach. Updated https://github.com/poanetwork/tokenbridge-contracts/pull/337/commits/79d21103b6f716e23da74c972a09e8d0b29db256

upgrade/src/migrateToMCD.js
patitonar: Updated the logs messages to remove unnecessary data https://github.com/poanetwork/tokenbridge-contracts/pull/337/commits/ca488abd46270cf258802b0023c5e33539887277

upgrade/src/confirmTransaction.js
patitonar: Yes, it makes sense to do it, updated the logic to multiply the estimated gas by 4 times. https://github.com/poanetwork/tokenbridge-contracts/pull/337/commits/a3410d1ec1e4280b98412b19c8fc44020fc5d17b

PR #346, opened 2019-12-30
rDai integration

maxaleks:
The feature is described in https://github.com/poanetwork/tokenbridge-contracts/issues/317


contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+    bytes32 internal constant RTOKEN = 0xdce7cb71f7466970418465f97f1fb7d721c1661bc59b1f0caa7d1f8ce374a47e; // keccak256(abi.encodePacked("rToken"))
+
+    function initializeRToken(
+        address _rToken,
+        address[] _recipients,
+        uint32[] _proportions
+    ) external onlyOwner returns (uint256) {
+        require(AddressUtils.isContract(_rToken));
+        addressStorage[RTOKEN] = _rToken;
+        erc20token().approve(_rToken, uint256(-1));
```
akolotov: Is it safe to approve operations for maximum amount of tokens? Does it make sense to approve the corresponding amount of tokens as part of the `mintRToken` operation?

contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+        return createAndChangeRTokenHat(_recipients, _proportions);
+    }
+
+    function removeRToken() external onlyOwner {
```
akolotov: should we make sure that the bridge has no balance in rDai before removing it? Does it make sense to redeem all tokens in that case?

contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+        addressStorage[RTOKEN] = address(0);
+    }
+
+    function redeemRToken(uint256 _redeemTokens) external onlyOwner returns (bool) {
+        return _redeemRToken(_redeemTokens);
+    }
+
+    function mintRToken(uint256 _mintAmount) public returns (bool) {
```
akolotov: what is the case to have any value returned from the method?

contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+        if (rToken() == address(0)) return;
```
akolotov: For this and all following cases where it is applicable does it make sense to explicitly return `false` in such construction?

contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    function initializeRToken(
+        address _rToken,
+        address[] _recipients,
+        uint32[] _proportions
+    ) external onlyOwner returns (uint256) {
+        require(AddressUtils.isContract(_rToken));
+        addressStorage[RTOKEN] = _rToken;
+        erc20token().approve(_rToken, uint256(-1));
+        return createAndChangeRTokenHat(_recipients, _proportions);
```
akolotov: for this and all following cases where it is applicable will the corresponding method revert in case of an error or will return `false`?

contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+        return IRToken(rToken()).createHat(_recipients, _proportions, true);
+    }
+
+    function rToken() public view returns (address) {
+        return addressStorage[RTOKEN];
+    }
+
+    function _redeemRToken(uint256 _redeemTokens) internal returns (bool) {
+        if (rToken() == address(0)) return;
+        return IRToken(rToken()).redeem(_redeemTokens);
```
akolotov: Since we rely here on a 3rd party contracts should we make sure that all requested amount of tokens was redeemed?

contracts/upgradeable_contracts/RTokenConnector.sol
```diff
@@ -0,0 +1,51 @@
...
+        addressStorage[RTOKEN] = address(0);
+    }
+
+    function redeemRToken(uint256 _redeemTokens) external onlyOwner returns (bool) {
+        return _redeemRToken(_redeemTokens);
+    }
+
+    function mintRToken(uint256 _mintAmount) public returns (bool) {
+        if (rToken() == address(0)) return;
+        return IRToken(rToken()).mint(_mintAmount);
```
akolotov: Since we rely here on a 3rd party contracts should we make sure that all requested amount of tokens was minted?

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -157,6 +164,8 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
 
         require(tokenBalance(fdToken).sub(curFDTokenBalance) == curHDTokenBalance);
 
+        mintRToken(_getTokenBalance());
```
akolotov: the oracles will call `swapTokens` till the moment when the Emergency Shutdown is executed. That's why I think the features "Two tokens support" and "Earning interest with locked DAI" should be independent: a new method to be called by oracles to convert Dai to rDai must be implemented, oracle must be changed to call this method from time to time, the monitor and the UI must be modified to operates with sum of balances Dai and rDai.

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Agree with you. I thought about lower costs for transactions but this change won't make much difference

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Good point, I will add it

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: I think I left it for something else 😅
I will remove it

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: It will revert. And I understood you, the return statement will be removed

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: If something goes wrong then the transfer of token will fail after `redeem` (where it is used inside the contracts)

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Then it will block `relayTokens` and `swapTokens` if something goes wrong

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
maxaleks: So you suggest removing this call, right?

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Should we handle the case when something is wrong with `rToken` and `redeemAll` method reverts? Then we won't be able to remove `rToken`

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
akolotov: yes

contracts/upgradeable_contracts/RTokenConnector.sol
akolotov: why do we need to return anything if there is no way to get result of transaction except its status?

contracts/upgradeable_contracts/RTokenConnector.sol
akolotov: should we introduce a flag `force` to handle that case?

contracts/upgradeable_contracts/RTokenConnector.sol
akolotov: And for my perspective it is a desirable behaviour.

contracts/upgradeable_contracts/RTokenConnector.sol
akolotov: Correct. And it will spend additional gas to call and execute the transfer. Moreover it will be hard to identify why it was failed on `transfer` by using traces. If it fails on `_redeemRTokens` we will see it in the block explorer (or by traces).

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Sorry, I do not quite understand you. Why is there no way to get result of transaction? And do you just want to remove all returned values?

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Good idea, I will add it

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Ok

contracts/upgradeable_contracts/RTokenConnector.sol
akolotov: how do you get a result of a method call if it is executed by `sendRawTransaction`?

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: Ok, I got your point. I will add this check

contracts/upgradeable_contracts/RTokenConnector.sol
maxaleks: There is no way to get it I think. I will remove returned values

akolotov:
Closed on account of #356

PR #353, opened 2020-01-01
Simplified require conditions for AMB bridge

k1rill-fedoseev:
It seems to me, that some of the `requires` for AMB bridge can be simplified, doing so may save some gas during execution.

### DIFF
--- a/contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
+++ b/contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
@@ -13,7 +13,7 @@ contract BasicAMB is BasicBridge {
         address _owner
     ) external onlyRelevantSender returns (bool) {
         require(!isInitialized());
-        require(_validatorContract != address(0) && AddressUtils.isContract(_validatorContract));
+        require(AddressUtils.isContract(_validatorContract));
         require(_gasPrice > 0);
         require(_requiredBlockConfirmations > 0);
 
--- a/contracts/libraries/Message.sol
+++ b/contracts/libraries/Message.sol
@@ -104,7 +104,7 @@ library Message {
         IBridgeValidators _validatorContract,
         bool isAMBMessage
     ) internal view {
-        require(isAMBMessage || (!isAMBMessage && isMessageValid(_message)));
+        require(isAMBMessage || isMessageValid(_message));
         uint256 requiredSignatures = _validatorContract.requiredSignatures();
         // It is not necessary to check that arrays have the same length since it will be handled
         // during attempt to access to the corresponding elements in the loop and the call will be reverted.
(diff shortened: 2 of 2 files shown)


contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -13,7 +13,7 @@ contract BasicAMB is BasicBridge {
         address _owner
     ) external onlyRelevantSender returns (bool) {
         require(!isInitialized());
-        require(_validatorContract != address(0) && AddressUtils.isContract(_validatorContract));
```
akolotov: There are similar checks in `erc20_to_native/HomeBridgeErcToNative.sol` and `RewardableBridge.sol`. Consider to re-factor code there as well.

k1rill-fedoseev: In https://github.com/poanetwork/tokenbridge-contracts/blob/c85aeb737bfd684b20b99520bd632ba4b8866583/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol#L140 we also want to allow to set `0x00..00` as a `blockReward` address, so here we cannot simplify this check.

Similarly, in https://github.com/poanetwork/tokenbridge-contracts/blob/c85aeb737bfd684b20b99520bd632ba4b8866583/contracts/upgradeable_contracts/RewardableBridge.sol#L59-L62 we want to allow set `0x00..00` as a fee manager contract address.

k1rill-fedoseev: By the way, is it supposed to be so?
In `erc20_to_erc20/HomeBridgeErcToErc.sol`, `erc20_to_native/HomeBridgeErcToNative.sol`, `native_to_erc20/ForeignBridgeNativeToErc.sol` and `native_to_erc20/HomeBridgeNativeToErc.sol` we allow `_feeManager` to be a contract address only, as shown here https://github.com/poanetwork/tokenbridge-contracts/blob/c85aeb737bfd684b20b99520bd632ba4b8866583/contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol#L94, but later, in https://github.com/poanetwork/tokenbridge-contracts/blob/c85aeb737bfd684b20b99520bd632ba4b8866583/contracts/upgradeable_contracts/RewardableBridge.sol#L59-L62 we can also use `address(0)` as a fee manager address.

Should we refactor this checks to make them consistent?

akolotov: > we also want to allow to set `0x00..00` as a `blockReward` address, so here we cannot simplify this check.

Sorry, you are right. The checks are not the same

akolotov: > we can also use `address(0)` as a fee manager address.
> Should we refactor this checks to make them consistent?

It was done to provide ability to disable a Fee Manager without upgrade of the bridge contract. So, the current implementation is correct.

Issue #317, opened 2019-11-15
Earning interest with locked DAI


maxaleks:
### Solutions
Some of the possible solutions to earn interest are [DAI Savings Rate](https://blog.makerdao.com/an-update-on-the-dai-savings-rate-in-multi-collateral-dai/), [Compound](https://compound.finance), [rDAI](https://rdai.money) and [Pool DAI](https://zeframlou.github.io/pool-dai/)

Let's take a closer look at each of them.

#### 1. DAI Savings Rate (DSR)
This feature will be available with the DAI update. It allows us to lock our DAI and earn interest without any risks. The MakerDAO provides us [the contract](https://github.com/makerdao/dss-proxy-actions/blob/master/src/DssProxyActions.sol#L891) with `join/exit` methods which we should call using [DSProxy](https://github.com/dapphub/ds-proxy/blob/master/src/proxy.sol) pattern ([here](https://github.com/makerdao/developerguides/blob/master/devtools/working-with-dsproxy/working-with-dsproxy.md) is the guide on how to use the pattern)

Example transactions: [join](https://kovan.etherscan.io/tx/0xf2f89f8811f847c18ca02ef5d36e3414f49f9d993ecfea3ceac9cd75778d9fac) (3 DAI) -> [join](https://kovan.etherscan.io/tx/0x9df4c4e467f6f4018435b58c3d122fac2cf52dd170cabf3215cd4aae4689c97e) (4 DAI) -> [exitAll](https://kovan.etherscan.io/tx/0xa681e2bb558dcdf0aa47dacc77cb0ecfb3c9475f205d18b11d1275ac150cb4c8) (7 DAI + earned interest)

The implementation steps will look like this:
1. deploy DSProxy contracts or use `DSProxyFactory.build()` in production
2. approve DAI tokens for deployed proxy
3. add the following code to the bridge contracts:
```
function _join(uint256 _amount) internal {
    bytes memory _data = abi.encodeWithSignature("join(address,address,uint256)", daiJoinAddress, potAddress, _amount);
    _execute(_data);
}

function _exit(uint256 _amount) internal {
    bytes memory _data = abi.encodeWithSignature("exit(address,address,uint256)", daiJoinAddress, potAddress, _amount);
    _execute(_data);
}

function _exitAll() internal {
    bytes memory _data = abi.encodeWithSignature("exitAll(address,address)", daiJoinAddress, potAddress);
    _execute(_data);
}

[...]

PR #356, opened 2020-01-10
Chai integration

k1rill-fedoseev:
Related to #317 and #346


k1rill-fedoseev:
I have tested contracts in the Kovan network.
Used Dai contract at [0xbf7a7169562078c96f0ec1a8afd6ae50f12e5a99](https://kovan.etherscan.io/address/0xbf7a7169562078c96f0ec1a8afd6ae50f12e5a99),
rDai token contract at [0x3183683cEeAb01699722053A2cb6A945cE0D7CeC](https://kovan.etherscan.io/address/0x3183683ceeab01699722053a2cb6a945ce0d7cec).

ForeignBridgeErcToNative was [deployed](https://kovan.etherscan.io/tx/0x47f004940200fab39782c6368757f674a78c5ff87d6ffe06d55e0845bae703b5) at [0x1426b5e4dd9a8efc7cc3ba02c09a5c78e88654d5](https://kovan.etherscan.io/address/0x1426b5e4dd9a8efc7cc3ba02c09a5c78e88654d5) and initialised.
Then, rToken was [initialised](https://kovan.etherscan.io/tx/0x78b001c4e03c6046d8736dfc3be95e2ccc66957c331030d6decba64a82d0ba02), rToken hat was [updated](https://kovan.etherscan.io/tx/0x34a240bf47d05499f19d3dae960dbeb27c816ee589bc757d5112e32b44db98cf).
Dai on the bridge account was [converted to rDai](https://kovan.etherscan.io/tx/0x1e67159f644ed50e67ba0a2aac7de4434a81af7c9664edd0222ff178b2cfbd6e).
In some time, all invested tokens were [redeemed](https://kovan.etherscan.io/tx/0xbbee2d94124b6082cf637b224f9fffb5f185bfcd9bb30a1ceecd110ef03ce475) back to the bridge address. RToken was [removed](https://kovan.etherscan.io/tx/0x060fa9b08aa929f9f3124f2c85167837e932963cec3d520f065d05ece45eda42).
Interest was payed to the specified account using rToken's `payInterest` method during [tx1](https://kovan.etherscan.io/tx/0xe7c211a3ce87c78cd37b8f7c96f81853272297906d62e716c60a6c05e3abab88), [tx2](https://kovan.etherscan.io/tx/0x21e5f5e97d817027c42d4ebee7a37e69107c9c6b940eba71093995f824ff571b). Collected interest was [converted](https://kovan.etherscan.io/tx/0x83abc5c993cb6dfcfb743635cb5433d11d2460479b3be3090571cc83e04c8cbb) back into regular Dai using a rToken's `redeem` method.

k1rill-fedoseev:
I have tested contracts in Kovan Testnet for some simple use cases.
There is no deployed Chai token, so I have deployed my own one at [0x623f148dabd40dd221807c5e98235bc40e54a106](https://kovan.etherscan.io/address/0x623f148dabd40dd221807c5e98235bc40e54a106), using existing [Kovan MakerDAO contracts](https://changelog.makerdao.com/releases/kovan/0.2.17/contracts.json).
- ForeignBridgeErcToNative was deployed at [0xaa5cf36b6d97a709fccf3fd4bea94da9753d9ca4](https://kovan.etherscan.io/address/0xaa5cf36b6d97a709fccf3fd4bea94da9753d9ca4) and initialised during [tx](https://kovan.etherscan.io/tx/0xb40a9b9dd4372859bb2a194dcf88e6bb1c58b9d2328768f29c51a98f5d426ae1)
- Call to `initializeChaiToken` at [tx](https://kovan.etherscan.io/tx/0xa971139eddfea7df882c79e176902977431cd7f36ff774d5af74a8729820d4b7)
- Call to `convertDaiToChai` at [tx](https://kovan.etherscan.io/tx/0x62d75dc1a5c95ead5371d09a62a23ae1456a25c03834996b8bf8f3fa89e714ad)
- Call to `payInterest` at [tx1](https://kovan.etherscan.io/tx/0x6aec389821873da104be3b5080ad9e230b8e8a9d9e62be78d460e4377a5ee772) and [tx2](https://kovan.etherscan.io/tx/0x87a898a5452badb1030d70b9add281bcfb4810add43a8ed6d97f72c924346d62)
- Call to `convertDaiToChai` at [tx](https://kovan.etherscan.io/tx/0xafcb40ced5eefed9774141e311e02e13e39527d2768456b8b900135a4db1a3ad)
- Call to `removeChaiToken` at [tx](https://kovan.etherscan.io/tx/0xdc393140b7399a65841576a64f573ce676f6046f74ab6d9bc3cf6944097f3323)

All performed conversions and transfers were executed as expected, balances were successfully updated. Calculation of interest worked as expected.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+    * @dev Pays all available interest, in Chai tokens
+    * @param recipient Account address to receive available interest
+    */
+    function payInterest(address recipient) external onlyOwner {
+        // since investedAmountInChai() returns a ceiled value,
+        // the value of chaiBalance() - investedAmountInChai() will be floored,
+        // leading to excess remaining chai balance
+        chaiToken().transfer(recipient, chaiBalance() - investedAmountInChai());
```
patitonar:
```suggestion
        chaiToken().transfer(recipient, chaiBalance().sub(investedAmountInChai()));
```

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+    * @dev Pays interest, in Chai tokens
+    * @param recipient Account address to receive available interest
+    * @param amount Amount of Chai tokens to transfer
+    */
+    function payInterest(address recipient, uint256 amount) external onlyOwner {
+        // check that the remaining chai balance will be sufficient to cover all invested DAI tokens
+        require(chaiBalance() - amount >= investedAmountInChai());
```
patitonar:
```suggestion
        require(chaiBalance().sub(amount) >= investedAmountInChai());
```

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+    */
+    function _convertChaiToDai(uint256 amount) internal {
+        uint256 invested = investedAmountInDai();
+        if (amount >= invested) {
+            chaiToken().draw(address(this), invested);
+            setInvestedAmointInDai(0);
+        } else if (amount > 0) {
+            uint256 initialDaiBalance = erc20token().balanceOf(address(this));
+            chaiToken().draw(address(this), amount);
+            uint256 redeemed = erc20token().balanceOf(address(this)) - initialDaiBalance;
```
patitonar:
```suggestion
            uint256 redeemed = erc20token().balanceOf(address(this)).sub(initialDaiBalance);
```

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+            setInvestedAmointInDai(redeemed < invested ? invested - redeemed : 0);
```
patitonar:
```suggestion
            setInvestedAmointInDai(redeemed < invested ? invested.sub(redeemed) : 0);
```

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+    * @param _chai Chai token contract address
+    */
+    function initializeChaiToken(address _chai) external onlyOwner {
+        require(address(chaiToken()) == address(0));
+        require(address(IChai(_chai).daiToken()) == address(erc20token()));
+        addressStorage[CHAI_TOKEN] = _chai;
+    }
+
+    /**
+    * @dev Withdraws all invested tokens, pays remaining interest, removes chai token from contrac storage
```
patitonar: Typo fix
```suggestion
    * @dev Withdraws all invested tokens, pays remaining interest, removes chai token from contract storage
```

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: Seems that here a usage of SafeMath is redundant, ternary operator will handle possible overflow.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: Here, SafeMath looks redundant as well. Balance after `draw` cannot be less than the initial one.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+        chaiToken().join(address(this), amount);
+    }
+
+    /**
+    * @dev Redeems DAI from Chai, the total redeemed amount will be at least equal to specified amount
+    * @param amount Amount of DAI to redeem
+    */
+    function _convertChaiToDai(uint256 amount) internal {
+        uint256 invested = investedAmountInDai();
+        if (amount >= invested) {
```
akolotov: What is the case when `amount >= invested`? Could this appear as a valid scenario?

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+        _convertChaiToDai(investedAmountInDai());
+        chaiToken().transfer(recipient, chaiBalance());
+        delete addressStorage[CHAI_TOKEN];
+    }
+
+    /**
+    * @dev Pays all available interest, in Chai tokens
+    * @param recipient Account address to receive available interest
+    */
+    function payInterest(address recipient) external onlyOwner {
```
akolotov: Having in mind the scheme required by the PoS introduction in the future (https://www.staketoken.net/token-model/xdai-stable-token) consider to do the following:
  * change the existing method `payInterest` as so to be available for every one. This method must pay the interest on a configured account (make sure that it is not zero). The method must allow its invocation with some minimal period (e.g. not often than once per week). If the method is called by the **upgradability owner** (as `claimTokens` in `BasicBridge.sol`) the rule to allow paying the interest with some minimal period is not applied. As soon as the interest payed, the configured account (if it is a contract) must be notified by calling a method (it seems logical to have `onTokenTransfer` there).
  * implement a method available for the owner to set up the account to receive the interest.
  * implement a method to configure min period between two subsequent attempts to call the method to pay the interest.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+
+        require(dsrBalance() >= investedAmountInDai());
+    }
+
+    /**
+    * @dev Pays interest, in Chai tokens
+    * @param recipient Account address to receive available interest
+    * @param amount Amount of Chai tokens to transfer
+    */
+    function payInterest(address recipient, uint256 amount) external onlyOwner {
```
akolotov: Consider to remove this method. We will not provide this functionality in such way. It will also allow to safe some gas during deployment.

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -46,6 +48,7 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
         uintStorage[EXECUTION_DAILY_LIMIT] = _homeDailyLimitHomeMaxPerTxArray[0];
         uintStorage[EXECUTION_MAX_PER_TX] = _homeDailyLimitHomeMaxPerTxArray[1];
         uintStorage[DECIMAL_SHIFT] = _decimalShift;
+        uintStorage[MIN_DAI_TOKEN_BALANCE] = 100 ether;
```
akolotov: Since it is related only to the introduction of Chai token into the bridge does it makes sense to initialize this parameter in `initializeChaiToken`?

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -171,7 +175,61 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
         emit TokensSwapped(hdToken, fdToken, curHDTokenBalance);
     }
 
-    function relayTokens(address _from, address _receiver, uint256 _amount, address _token) external {
+    function setMinDaiTokenBalance(uint256 _minBalance) external onlyOwner {
```
akolotov: If all these new methods are valid only in case of Chai used by the bridge does it makes sense to move them to the `ChaiConnector` module?

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -171,7 +175,61 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
+    }
+
+    /**
+    * @dev Converts all DAI into Chai tokens, keeping minDaiTokenBalance() DAI as a buffer
+    */
+    function convertDaiToChai() public {
+        _convertDaiToChai(tokenBalance(erc20token()).sub(minDaiTokenBalance()));
+    }
+
+    /**
```
akolotov: Consider to remove `convertDaiToChai(amount)` and `convertChaiToDai(amount)` to reduce the power of the owner. It is assumed that the owner is responsible only for changing the parameters of the bridge. It also will allows to safe some gas during the contract deployment.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,156 @@
...
+    */
+    function rdivup(uint256 x, uint256 y) internal pure returns (uint256) {
+        return x.mul(ONE).add(y.sub(1)) / y;
+    }
+
+    /**
+    * @dev Initializes chai token
+    * @param _chai Chai token contract address
+    */
+    function initializeChaiToken(address _chai) external onlyOwner {
```
akolotov: Consider to hardcode the address of Chai token contract.

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -79,6 +82,12 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
     ) internal returns (bool) {
         setTotalExecutedPerDay(getCurrentDay(), totalExecutedPerDay(getCurrentDay()).add(_amount));
         uint256 amount = _amount.div(10**decimalShift());
+
+        uint256 currentBalance = tokenBalance(erc20token());
+        if (currentBalance < amount) {
+            _convertChaiToDai(amount.sub(currentBalance).add(minDaiTokenBalance()));
```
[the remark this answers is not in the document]
k1rill-fedoseev: This seems to be an invalid scenario. 
If Chai token is not configured, then bridge should have everything stored in Dai, so it will have enough funds to cover any xDai => Dai transfer, in other words `currentBalance >= amount` in this case.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: The only case is related to the `convertChaiToDai(uint256 _amount)`, so that it will just convert all remaining chai into dai instead of revert on too high `_amount`.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: Should we pay the interest in Dai directly, so than the bridge will have to withdraw Dai first?

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: Should we still have a functionality for enabling/disabling chai token then?

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: There is another case, by the way. In `onExecuteMessage` bridge will try to withdraw tokens, so that the balance will be equal to `minDaiTokenBalance()` after the operation. If the overall bridge balance is less than this limit, it will just withdraw all remaining tokens.

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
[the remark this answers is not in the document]
akolotov: Understood. Please add a comment here to reduce number of comments from the security auditors.

contracts/upgradeable_contracts/ChaiConnector.sol
akolotov: Thanks for the question. My view is that the exchange from Chai to Dai will be responsibility of the receiver (either EoA or a contract). Do you see any objections?

contracts/upgradeable_contracts/ChaiConnector.sol
akolotov: Yes, calling of `initializeChaiToken` will enable the functionality (the Chai token contract address will be hardcoded here instead of specifying be the transaction sender). `removeChaiToken` will disable it. Do you see it in different way?

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: If the users are not going to operate with Chai token directly, then I think it is better to keep all the functionality related to Chai in one place. 
This will also simplify switching to other investing approach in future updates, if needed.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: No, this approach should work fine.

contracts/upgradeable_contracts/ChaiConnector.sol
akolotov: > The only case is related to the convertChaiToDai(uint256 _amount), so that it will just convert all remaining chai into dai instead of revert on too high _amount.

From my perspective it is better to revert by requiring explicit actions from the owners. But I would insist of removing `convertChaiToDai(amount)` at all.

akolotov: > If the overall bridge balance is less than this limit, it will just withdraw all remaining tokens.

Got you! Please add the corresponding comment to the code to not forget this case. Is this scenario covered in tests?

contracts/upgradeable_contracts/ChaiConnector.sol
akolotov: Good reasoning. Let's implement the functionality in the way you suggested. Thanks!

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: Not properly. I will add an explicit test on that.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,191 @@
...
+    /**
+    * @dev Redeems DAI from Chai, the total redeemed amount will be at least equal to specified amount
+    * @param amount Amount of DAI to redeem
+    */
+    function _convertChaiToDai(uint256 amount) internal {
+        uint256 invested = investedAmountInDai();
+        if (amount >= invested) {
+            // onExecuteMessage can call a convert operation with argument greater than the current invested amount,
+            // in this case bridge should withdraw all invested funds
+            chaiToken().draw(address(this), invested);
```
akolotov: The question about Chai/DSR: what is more profitable: pay the interest before Dai withdrawal or continue to keep the interest until the method `payInterest` is called?

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,191 @@
...
+    * @dev Internal function for paying all available interest, in Dai tokens
+    * @param recipient Account address to receive available interest
+    */
+    function _payInterest(address recipient) internal {
+        // since investedAmountInChai() returns a ceiled value,
+        // the value of chaiBalance() - investedAmountInChai() will be floored,
+        // leading to excess remaining chai balance
+        uint256 balanceBefore = erc20token().balanceOf(address(this));
+        chaiToken().exit(address(this), chaiBalance().sub(investedAmountInChai()));
+        uint256 interestInDai = erc20token().balanceOf(address(this)) - balanceBefore;
```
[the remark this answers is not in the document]
k1rill-fedoseev: If we look into the implementation of Chai and MakerDAO contracts, it is clear that it is not possible for balance to decrease.
Why then we need to consider such additional checks?

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,191 @@
...
+        uint256 chi = (now > pot.rho()) ? pot.drip() : pot.chi();
+        return rdivup(investedAmountInDai(), chi);
+    }
+
+    /**
+    * @dev Invests DAI into Chai
+    * @param amount Amount of DAI to invest
+    */
+    function _convertDaiToChai(uint256 amount) internal {
+        setInvestedAmointInDai(investedAmountInDai() + amount);
```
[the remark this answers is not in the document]
k1rill-fedoseev: No, it is not necessary to use SafeMath here. The only place where this function is used, uses `tokenBalance(erc20token()).sub(minDaiTokenBalance())` as an argument, which cannot be an extremely high value.
I can add a comment for that.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: There is no any profit in paying interest before Dai withdrawal. 
In DSR, after each second the interest is reevaluated for your account. The formula depends only on the current amount of DAI saved in DSR. So there is no need to explicitly
call `payInterest` before withdrawal.
The optimal way is to keep DAI in DSR as long as you can, and then pay the interest only when it is needed.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,191 @@
...
+    */
+    function _payInterest(address recipient) internal {
+        // since investedAmountInChai() returns a ceiled value,
+        // the value of chaiBalance() - investedAmountInChai() will be floored,
+        // leading to excess remaining chai balance
+        uint256 balanceBefore = erc20token().balanceOf(address(this));
+        chaiToken().exit(address(this), chaiBalance().sub(investedAmountInChai()));
+        uint256 interestInDai = erc20token().balanceOf(address(this)) - balanceBefore;
+
+        erc20token().transfer(recipient, interestInDai);
```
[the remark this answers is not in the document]
k1rill-fedoseev: What about a signature of `onTokenTransfer`. Should it be `onTokenTransfer(uint256 value)`, `onTokenTransfer(address from, uint256 value)`, or something different?

contracts/upgradeable_contracts/ChaiConnector.sol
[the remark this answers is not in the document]
akolotov: I see.

> Why then we need to consider such additional checks?

1) Just to make sure that everything will behave as expected.
2) At least we need to add a comment to the code that we intentionally leave there the origin minus operation. Probably in the future the security auditors will have less questions after reading this comment. Thanks in advance.

contracts/upgradeable_contracts/ChaiConnector.sol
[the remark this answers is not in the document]
akolotov: From my perspective we should be consistent and expect that the contract supports `onTokenTransfer`should be as it is described in [EIP677](https://github.com/ethereum/EIPs/issues/677).

contracts/upgradeable_contracts/ChaiConnector.sol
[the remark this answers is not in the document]
akolotov: Please add the comment.

contracts/upgradeable_contracts/ChaiConnector.sol
akolotov: Ok. Thanks for clarification.

contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -171,7 +178,35 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
-    function relayTokens(address _from, address _receiver, uint256 _amount, address _token) external {
+    /**
+    * @dev Checks if DAI balance is high enough to be partially converted to Chai
+    * Twice limit is used in order to decrease frequency of convertDaiToChai calls,
+    * In case of high bridge utilization in DAI => xDAI direction,
+    * convertDaiToChai() will be called as soon as DAI balance reaches 2 * limit,
+    * limit DAI will be left as a buffer for future operations.
+    * @return true if convertDaiToChai() call is needed to be performed by the oracle
+    */
+    function isDaiNeedsToBeInvested() public view returns (bool) {
```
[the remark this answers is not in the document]
k1rill-fedoseev: Moved these methods to `ChaiConnector`

contracts/upgradeable_contracts/ChaiConnector.sol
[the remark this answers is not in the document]
k1rill-fedoseev: Ok. Added SafeMath for making sure everything works fine

contracts/upgradeable_contracts/ChaiConnector.sol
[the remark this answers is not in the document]
k1rill-fedoseev: Added functionality for recipient notification via `onTokenTransfer(address,uint256,bytes)`, added possibility of restrictions on frequency of `payInterest` call. Made `payInterest` to be publicly available.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,269 @@
...
+        // the value of chaiBalance() - investedAmountInChai() will be floored,
+        // leading to excess remaining chai balance
+        uint256 balanceBefore = daiBalance();
+        chaiToken().exit(address(this), chaiBalance().sub(investedAmountInChai()));
+        uint256 interestInDai = daiBalance().sub(balanceBefore);
+
+        erc20token().transfer(interestReceiver(), interestInDai);
+
+        if (AddressUtils.isContract(interestReceiver())) {
+            interestReceiver().onTokenTransfer(address(this), interestInDai, "");
```
akolotov: Consider to call `onTransferToken` in more safe way similar we call in the bridgeable token:

https://github.com/poanetwork/tokenbridge-contracts/blob/efbbdfade1d2eae6a6c6f12f1a59aa936470ad6f/contracts/ERC677BridgeToken.sol#L68

It will guarantee that the call will not be reverted even if the contract receiver does not support this method call.

contracts/mocks/ForeignBridgeErcToNativeMock.sol
```diff
@@ -17,4 +17,14 @@ contract ForeignBridgeErcToNativeMock is ForeignBridgeErcToNative {
...
         return ISaiTop(0x96bc48adACdB60E6536E55a6727919a397F14540);
     }
+
+    bytes32 internal constant CHAI_TOKEN_MOCK = 0x5d6f4e61a116947624416975e8d26d4aff8f32e21ea6308dfa35cee98ed41fd8; // keccak256(abi.encodePacked("chaiTokenMock"))
+
+    function setChaiToken(IChai _chaiToken) external {
+        addressStorage[CHAI_TOKEN_MOCK] = _chaiToken;
+    }
+
+    function chaiToken() public view returns (IChai) {
```
akolotov: Does it make sense to redefine `uintStorage[INTEREST_COLLECTION_PERIOD]` in this mock as well in order to cover the following functionality by tests: the paying interest is possible after `INTEREST_COLLECTION_PERIOD` expiration?

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        require(dsrBalance() >= investedAmountInDai());
```
akolotov: What is purpose of this check? To rollback the transaction if something has gone wrong? Does it make sense to introduce similar check into `convertDaiToChai` and `_ convertChaiToDai`?

akolotov:
@patitonar Gerardo, may I ask you to double check if all corner cases are covered by tests? Thanks!

contracts/mocks/ForeignBridgeErcToNativeMock.sol
k1rill-fedoseev: Yes, this makes sense.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: I didn't put such check in convert operations, since there are not changing the total balance of a bridge. On the other hand, `payInterest` decreases the overall balance of the bridge, so I have added this extra check.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,269 @@
...
+    bytes32 internal constant MIN_DAI_TOKEN_BALANCE = 0xce70e1dac97909c26a87aa4ada3d490673a153b3a75b22ea3364c4c7df7c551f; // keccak256(abi.encodePacked("minDaiTokenBalance"))
+
+    uint256 internal constant ONE = 10**27;
+
+    /**
+    * @dev Throws if chai token is not enabled
+    */
+    modifier chaiTokenEnabled {
+        require(isChaiTokenEnabled());
+        _;
```
patitonar: Let's add the following line, as in the other cases,  before `_;` :
```
/* solcov ignore next */
```
so the coverage tool doesn't incorrectly interpret it as an uncovered line

test/erc_to_native/foreign_bridge.test.js
```diff
@@ -1312,4 +1590,290 @@ contract('ForeignBridge_ERC20_to_Native', async accounts => {
...
+        [homeDailyLimit, homeMaxPerTx],
+        owner,
+        decimalShiftZero,
+        otherSideBridge.address
+      )
+      chaiToken = await createChaiToken(token, foreignBridge, owner)
+      interestRecipient = await ERC677ReceiverMock.new()
+    })
+
+    describe('initializeChaiToken', () => {
```
patitonar: I think we should add a unit test that calls `initializeChaiToken()` two times, in order to test that it can't be initialized twice

test/erc_to_native/foreign_bridge.test.js
```diff
@@ -358,6 +378,127 @@ contract('ForeignBridge_ERC20_to_Native', async accounts => {
...
+
+      const { logs } = await foreignBridge.executeSignatures(message, oneSignature).should.be.fulfilled
+
+      logs[0].event.should.be.equal('RelayedMessage')
+      logs[0].args.recipient.should.be.equal(recipientAccount)
+      logs[0].args.value.should.be.bignumber.equal(value)
+      const balanceAfter = await token.balanceOf(recipientAccount)
+      const balanceAfterBridge = await token.balanceOf(foreignBridge.address)
+      balanceAfter.should.be.bignumber.equal(balanceBefore.add(value))
+      balanceAfterBridge.should.be.bignumber.gte(ether('1'))
```
patitonar: Is it correct that in this scenario all the tokens invested were converted from chai to dai when executing the signatures? Does it make sense to also check that the chai balance of the bridge should be `ZERO`?

test/erc_to_native/foreign_bridge.test.js
k1rill-fedoseev: Sure, I will add this test

test/erc_to_native/foreign_bridge.test.js
k1rill-fedoseev: Yeah, it makes sense

k1rill-fedoseev: Actually, the chai balance won't be 0, a small interest will be already collected between calls to `convertDaiToChai()` and `executeSignatures()`.
Used a comparison with a relatively small number for that.

contracts/mocks/ForeignBridgeErcToNativeMock.sol
k1rill-fedoseev: Used `setInterestCollectionPeriod()` for testing this case.

contracts/upgradeable_contracts/ChaiConnector.sol
k1rill-fedoseev: So, from your perspective, what is better to do with this checks?

akolotov: I would add such check to `convertDaiToChai` just to make sure that the Chai/DSR/whatever contract which we do not control has not started to behave unexpectedly. The situation when we cannot convert Dai to Chai due to reverted transaction is better than unexpected behavior leading to unbalancing of the bridge. 

I would add such check to `_convertChaiToDai` since it will lead to failure of `executeSignatures` call. Failure of the call will bring our attention on the problem. So, if due to some reason (probably a bug) Chai/DSR/whatever will not convert the tokens properly we will notice and will try to handle this situation. Otherwise it could lead to unbalancing of the bridge as well.

What do you think?

k1rill-fedoseev: I think, it is not a problem to have such checks. So let's add them.

In case of `_convertChaiToDai` it is not a problem at all. However, there is a problematic moment in `convertDaiToChai`. Let me explain:

When evaluating the amount of DAI kept in Chai using `dsrBalance()`, there are some fixed point truncations. 
The dependency between invested amount of DAI - `value` and returned value of `dsrBalance()` - `res` is the following:
`res = floor(floor(value / chi) * chi))`, 
where `chi` is a fixed point number with precision of 27 digits. This coefficient is increasing every second, currently it is ~ `1.012`.
This, if I am not mistaken, can lead up to losses of `ceil(chi)` DAI in this balance evaluation.
 
This is not a problem for the bridge itself, since after even one block, the collected interest will be orders more than this error.
But, this reasoning implies that the check should be different:
`require(dsrBalance() >= investedAmountInDai() - (1 + chaiToken().pot().chi() / 10**27));`

k1rill-fedoseev: Also, I am not quite sure if this will work properly for several calls to `convertDaiToChai()` in the same block. I will try to test such scenario, to figure that out.

k1rill-fedoseev: One more option for this check, is to use high enough constant in inequality (e. g. `require(dsrBalance() >= investedAmountInDai() - 1000000)`)
Such approach will definitely work for all cases.

akolotov: I am OK with any option especially if it is well documented in the code.

k1rill-fedoseev: Done

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -0,0 +1,278 @@
...
+            chaiToken().draw(address(this), invested);
+            setInvestedAmointInDai(0);
+        } else if (amount > 0) {
+            uint256 initialDaiBalance = daiBalance();
+            chaiToken().draw(address(this), amount);
+            uint256 redeemed = daiBalance() - initialDaiBalance;
+            setInvestedAmointInDai(redeemed < invested ? invested - redeemed : 0);
+        }
+
+        require(dsrBalance() >= investedAmountInDai());
```
akolotov: Do you thinks that is more logical to check if the new balance of the bridge is as expected?

k1rill-fedoseev: What do you mean? 
Seems to me that if a new balance will not be as expected, we will get a revert anyway while trying to transfer DAI to user's account in `onExecuteMessage`.

akolotov: I meant that if you requested 300 dai (200 to execute transfer and 100 to have minimal balance) and by some reason receives only 250 dai as part of the execution `chaiToken().draw`, so we are in trouble.

k1rill-fedoseev: Ok, reasonable, I will add it.
What about a similar check in `convertDaiToChai`? We can also check that all approved amount was used.

akolotov: WRT `converDaiToChai`... Our main worry in this case is that all tokens we send will be considered as invested. So, correct me if I wrong: `chaiToken().join` performs an action as so the result of these actions will be reflected later in the result of `dsrBalance`. And the expected amount of invested tokens is calculated by `setInvestedAmointInDai(investedAmountInDai() + amount)`. If my reasoning is correct then  `require(dsrBalance() >= investedAmountInDai() - 10000)` checks that the expected amount not less than the actual amount of tokens. Since actions in `converDaiToChai` will not send more token than expected (`amount`) we don't need to check if the bridge balance was changed unexpectedly, do we?

But in `_convertChaiToDai` we could check both: expected amount of actual balance  and expected amount of invested tokens. In case if the expectation does not match -- revert the method call. Please share your view on this.

k1rill-fedoseev: Sounds correct. 
But what if something goes wrong in `converDaiToChai`, so that not all specified tokens are invested? For example, bridge approves `200` tokens, but Chai/MakerDAO contracts used only `150` of them, and `50` are still kept in approval. For such case, we also need to pay attention, don't we?

akolotov: they are still on the balance of the bridge so it is not the issue.

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -270,6 +270,9 @@ contract ChaiConnector is Ownable, ERC20Bridge {
             uint256 initialDaiBalance = daiBalance();
             chaiToken().draw(address(this), amount);
             uint256 redeemed = daiBalance() - initialDaiBalance;
+
+            require(redeemed >= amount);
```
akolotov: Don't you need a similar check in the case if `chaiToken().draw(address(this), invested)` is invoked?

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -176,8 +176,7 @@ contract ChaiConnector is Ownable, ERC20Bridge {
     * @return Balance in dai, truncated
     */
     function dsrBalance() public view returns (uint256) {
-        uint256 chi = chaiToken().pot().chi();
-        return rmul(chi, chaiBalance());
+        return chaiToken().dai(address(this));
```
akolotov: implementation of `rmul` can be removed now, right?

k1rill-fedoseev: Right, removed it

PR #357, opened 2020-01-15
Single execute signatures

### LINKED ISSUE #354: Unify usage of hasEnoughValidSignatures
As part of work under AMB feature a new way to transmit signatures was suggested and implemented as the alternative implementation of `hasEnoughValidSignatures`:

https://github.com/poanetwork/tokenbridge-contracts/blob/c85aeb737bfd684b20b99520bd632ba4b8866583/contracts/libraries/Message.sol#L124-L157

The reason of the suggestion is to achieve more cheaper transactions for the validators: instead of sending arrays `_vs`, `_rs` and `_ss` of the same size the signatures are transferred as a blob which consumes less gas for sending but could require more gas for handling. Tests showed that the high gas cost for nonzero bytes in the calldata (`68` gas) makes the new implementation more effective from gas usage point of view.

The [EIP-2028](https://eips.ethereum.org/EIPS/eip-2028) in the Istanbul hardfork reduces the cost of calldata from 68 to 16 gas per non-zero byte. It means that we need to re-examine if the new implementation is still effective for different number of signatures required for full confirmation. If it is still effective it is necessary to change the `executeSignatures` method in `BasicForeignBridge` as so the new implementation is used.

k1rill-fedoseev:
Resolves #354

### DIFF
--- a/contracts/libraries/Message.sol
+++ b/contracts/libraries/Message.sol
@@ -98,31 +98,6 @@ library Message {
 
     function hasEnoughValidSignatures(
         bytes _message,
-        uint8[] _vs,
-        bytes32[] _rs,
-        bytes32[] _ss,
-        IBridgeValidators _validatorContract,
-        bool isAMBMessage
-    ) internal view {
-        require(isAMBMessage || (!isAMBMessage && isMessageValid(_message)));
-        uint256 requiredSignatures = _validatorContract.requiredSignatures();
-        // It is not necessary to check that arrays have the same length since it will be handled
-        // during attempt to access to the corresponding elements in the loop and the call will be reverted.
-        // It will save gas for the rational validators actions and still be safe enough from security point of view
-        require(_vs.length >= requiredSignatures);
-        bytes32 hash = hashMessage(_message, isAMBMessage);
-        address[] memory encounteredAddresses = new address[](requiredSignatures);
-
-        for (uint256 i = 0; i < requiredSignatures; i++) {
-            address recoveredAddress = ecrecover(hash, _vs[i], _rs[i], _ss[i]);
-            require(_validatorContract.isValidator(recoveredAddress));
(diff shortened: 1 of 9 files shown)


contracts/libraries/Message.sol
```diff
@@ -96,31 +96,6 @@ library Message {
...
-
-        for (uint256 i = 0; i < requiredSignatures; i++) {
-            address recoveredAddress = ecrecover(hash, _vs[i], _rs[i], _ss[i]);
-            require(_validatorContract.isValidator(recoveredAddress));
-            require(!addressArrayContains(encounteredAddresses, recoveredAddress));
-            encounteredAddresses[i] = recoveredAddress;
-        }
-    }
-
     function hasEnoughValidSignatures(
```
akolotov: Please consider to add documentation to the method. Since we have https://github.com/poanetwork/tokenbridge-contracts/issues/320 it is more reasonable to address the concern raised there when the corresponding functionality is being changed rather than try to introduce all the comments in one activity.

contracts/upgradeable_contracts/BasicForeignBridge.sol
```diff
@@ -15,8 +15,9 @@ contract BasicForeignBridge is EternalStorage, Validatable, BasicBridge, BasicTo
     event RelayedMessage(address recipient, uint256 value, bytes32 transactionHash);
     event UserRequestForAffirmation(address recipient, uint256 value);
 
-    function executeSignatures(uint8[] vs, bytes32[] rs, bytes32[] ss, bytes message) external {
-        Message.hasEnoughValidSignatures(message, vs, rs, ss, validatorContract(), false);
+    function executeSignatures(bytes message, bytes signatures) external {
```
akolotov: Please consider to add the method documentation here.

Issue #65, opened 2018-08-13
Support of native-to-native mode


akolotov:
Latest `stable` version of Parity (1.11) allows extend standard block reward mechanism in networks which supports AuRA consensus (e.g. PoA network). New mechanism allows to create a contract which will be called as part of every block verification automatically. The method `reward` of this contract is called and should return a list of addresses and amount of ether which could get the corresponding addresses.
  * https://wiki.parity.io/Block-Reward-Contract.html 

This functionality can be used by `native-to-native` bridge when funds are sent from one network to another. After receiving the confirmation that funds needs to be received on another side the bridge registers address of account receiver in the `BlockRewardContract` as so new block is mined, the account gets the corresponding amount of ether. 
  * https://github.com/poanetwork/RFC/issues/22
  * https://github.com/varasev/test-block-reward

Consider to extend bridge functionality by enabling new mode of the bridge.

PR #359, opened 2020-01-20
Merge the develop branch to the master branch, preparation to v4.0.0-rc0

### LINKED ISSUE #352: Incorrect AMB minimum gas usage estimation after EIP-2028
The [EIP-2028](https://eips.ethereum.org/EIPS/eip-2028) reduces the cost of calldata from `68` to `16` gas per non-zero byte.
In `MessageDelivery` contract we have the following:
https://github.com/poanetwork/tokenbridge-contracts/blob/c85aeb737bfd684b20b99520bd632ba4b8866583/contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol#L14-L18
Should we reduce a multiplier to `16`?

### LINKED ISSUE #354: Unify usage of hasEnoughValidSignatures

akolotov:
This merge contains the following set of improvements and fixes:
  * [Improvement] Updated minimum gas usage estimation (#355), closes #352
  * [Improvement] Simplified some conditions in require() clauses (#353)
  * [Improvement] Single execute signatures (#357), closes #354
  * [Fix] Fixes in AMB deployment related documentation and scripts (#358)
  * [Other] badges links corrected to point to tokenbridge-contracts

It was decided to increase the major version of the release since the changes made under #357 are not backward compatible.

### DIFF
--- a/contracts/libraries/Message.sol
+++ b/contracts/libraries/Message.sol
@@ -96,31 +96,19 @@ library Message {
         }
     }
 
-    function hasEnoughValidSignatures(
-        bytes _message,
-        uint8[] _vs,
-        bytes32[] _rs,
-        bytes32[] _ss,
-        IBridgeValidators _validatorContract,
-        bool isAMBMessage
-    ) internal view {
-        require(isAMBMessage || isMessageValid(_message));
-        uint256 requiredSignatures = _validatorContract.requiredSignatures();
-        // It is not necessary to check that arrays have the same length since it will be handled
-        // during attempt to access to the corresponding elements in the loop and the call will be reverted.
-        // It will save gas for the rational validators actions and still be safe enough from security point of view
-        require(_vs.length >= requiredSignatures);
-        bytes32 hash = hashMessage(_message, isAMBMessage);
-        address[] memory encounteredAddresses = new address[](requiredSignatures);
-
-        for (uint256 i = 0; i < requiredSignatures; i++) {
-            address recoveredAddress = ecrecover(hash, _vs[i], _rs[i], _ss[i]);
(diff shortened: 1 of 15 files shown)


jazzholicbae:
@akolotov Hello, any chance to have this feature soon? https://github.com/poanetwork/tokenbridge-contracts/pull/359

akolotov:
> @akolotov Hello, any chance to have this feature soon? #65

Hi! My understanding is that we agreed in Gitter that your team will contribute this functionality:

image

jazzholicbae:
@akolotov Oops, my misunderstanding, sorry!

Issue #365, opened 2020-01-28
strategy to migrate an origin native-to-erc bridge to AMB


akolotov:
This issue is being opened to make research to identify ability of migrating a native-to-erc bridge (e.g. WETC bridge) to be an extension of AMB bridge.

The following is required to be considered:
  * the existing pair of proxy contracts (Ethereum Classic: `0x073081832B4Ecdce79d4D6753565c85Ba4b3BeA9`, Ethereum Mainnet: `0x0cB781EE62F815bdD9CD4c2210aE8600d43e7040`) must be the proxy for AMB mediators contracts
  * the current bridge state must preserve (balances and completed operations)

Please provide an analysis if the migration is possible, what limitations need to be taken into account (e.g. due to limited number of functionality existing in Ethereum Classic), what special handling must be done in comparison with [a generic native-to-erc20 extension for AMB](https://github.com/poanetwork/tokenbridge-contracts/issues/364).

The implementation stage will begin only after the review of the analysis described above.

patitonar:
Here is the result of my review.

## Security
AMB validators will listen to different events on different contracts with different types of messages and how they are processed. Also, validators could be different. So there is no risk to reprocess a completed message.

## Functionality
All functionality from the bridge can be kept as a mediator.
Regarding the fee distribution, the current fee manager implementation for this mode is intended to work with `RewardableValidators` contract. If the fees will be distributed to the AMB validators, which uses `BridgeValidators` contract, the fee manager should be adapted accordingly (i.e. the method `getValidatorRewardAddress` is not defined).


## Bridge state
Upgrading the implementation contract will not affect the locked ETC value in the proxy contract nor the ERC677 tokens hold by the users.

If `EternalStorage` is also used for the mediators, the list of parameters currently stored in the contract should continue to be accessible if the same key is used.

| Parameters in home bridge        | Possible parameters needed in home mediator           |
| ------------- |:-------------:|
| requiredBlockConfirmations |  |
| deployedAtBlock |  |
| gasPrice |  |
| isInitialized | isInitialized |
| owner | owner |
| validatorContract | validatorContract |
| dailyLimit | dailyLimit |
| maxPerTx | maxPerTx |
| minPerTx | minPerTx |
| executionDailyLimit | executionDailyLimit |
| executionMaxPerTx | executionMaxPerTx |
| feeManagerContract | feeManagerContract |
| homeFee | homeFee |
| foreignFee | foreignFee |
|  | bridgeContract |
[...]

Issue #364, opened 2020-01-28
native-to-erc20 AMB extension


akolotov:
Consider to create a generic AMB extension which could be used to transfer a native tokens in one chain to ERC20/677 tokens to another chain and back. It must support the functionality the current implementation of the generic native-to-erc20 bridge (e.g. WETC bridge) contains.

PR #367, opened 2020-02-13
Add native to erc20 mediator on top of AMB 

patitonar:
Relates to #364, #365


contracts/upgradeable_contracts/BasicAMBMediator.sol
```diff
@@ -0,0 +1,134 @@
...
+        require(_requestGasLimit <= maxGasPerTx());
+        uintStorage[REQUEST_GAS_LIMIT] = _requestGasLimit;
+    }
+
+    /**
+    * @dev Sets the unique identifier to be used for the next message to be sent to the AMB bridge.
+    * @param _hash representing the next message unique identifier.
+    */
+    function setNonce(bytes32 _hash) internal {
+        bytesStorage[NONCE] = abi.encodePacked(_hash);
```
k1rill-fedoseev: Here, and in all similar places (like `setTransactionHash()`, `setFailedMessageDataHash(bytes32)`), what is the purpose of storing data in `bytes` rather than in `bytes32`?
Seems to me, that if the value is always `32` bytes long, storing it as a `bytes32` will save some gas.

patitonar:
I think this is ready for review, here are some useful comments:

#### Fee Manager

In the existing implementation of Native-to-erc we have two ways of collecting fees:
- On both sides (current implementation in WETC bridge)
- Only on home side for transfers in both directions

To collect the fees only on home side for both transfers we used `onSignaturesCollected` method which is called after all signatures are collected, so we could safely distribute fees there.

In a mediator implementation, we don't have that opportunity to call logic when the signatures are collected (which is OK since it is an internal operation of AMB) so we can't implement the logic to distribute fees in that way.

That's why the implementation for the fee manager only considers the way to distribute the fees on both sides when the message is correctly executed on the other network, which is the same way used in the WETC bridge. This also simplified the logic in the fee manager.

Another change is that I implemented two different contracts to be deployed as fee managers, one for home and one for foreign. In this case, each fee manager has the logic it only needs and also I think it helps to better understand its usage in the mediator contracts.

#### WETC specific

Regarding the specific usage of this mediator for WETC bridge, while there are no restrictions now from the network, there is a restriction from the current proxy contract that is being used.

The proxy contract being used is `ClassicEternalStorageProxy`. It has the restriction that for every call delegated to the implementation contract it will return a fixed-size value of `32` unless we specified the size for a method signature as we did for `signature(bytes32,uint256)` and `message(bytes32)`.

In this case, it affects the method `rewardAccounts()` which returns the list of reward accounts that will get the fees that has a variable size. It doesn't affect the operations, fees are correctly distributed, but it is not possible to perform an external call if someone wants to know the list. That's why I implemented `ClassicHomeAMBNativeToErc20` that maintains the size to be returned by the proxy so it will work correctly with the `ClassicEternalStorageProxy` proxy contract.

I tested both implementations in Sokol - Kovan with the existing AMB bridge:

- Test with regular Home and Foreign versions
[   Home  ] Bridge Mediator: 0x21ec0Ccd8Cc8430dCb04AA4124AF7995E5Be50DF
[   Home  ] Fee Manager: 0x45220dec68F4733181D13e062dcF6106e4b0BdD8
[ Foreign ] Bridge Mediator: 0x9cEB11d7a50F29eB1100D6EfBfbCa3586c27D03F
[ Foreign ] Fee Manager: 0x604D0C32723E11A2E6b7D01717F3Afd730c76C4a
[ Foreign ] ERC677 Token: 0x3459EDf5CEd322fC911bE03ACD3E177dD064d2Ba

- Test with Classic version for Home and regular version for Foreign
[   Home  ] Bridge Mediator: 0x6c2bA57847e0a1B9eCC8292f6Dbe0776e4B9EAfc
[   Home  ] Fee Manager: 0x6d98Af026aBb1A2A8665f88C210f420A3615b3dA
[ Foreign ] Bridge Mediator: 0x86c070D543A8aa4bCdEf1c78182F207dA55236Aa
[ Foreign ] Fee Manager: 0x55Ff8C6D2aA87f291E1Ce74cbc74D775A7538E5e
[ Foreign ] ERC677 Token: 0xD41BC633C05d4Db17122Cc9353DF31a0D51f99A2

After we agree on the changes, we should prepare to test the WETC bridge upgrade. The following are the possible steps for the upgrades and the methods to be called to migrate the storage, they were **NOT tested**.

##### Possible migration Steps
0. An AMB bridge should be first deployed for ETC-ETH
1. Deploy Home Fee manager contract
2. Deploy Foreign Fee manager contract
3. Update the upgrade methods with the missing contract addresses
4. Deploy Classic home Mediator contract
5. Deploy Foreign Mediator contract
6. Stop bridge operations in both directions
7. In Home call the proxy method upgradeToAndCall to call migrate method
8. In Foreign call the proxy method upgradeToAndCall to call migrate method
9. Resume bridge operations in both directions

##### Upgrade Method for home
```solidity
/**
* @dev Method to migrate home WETC native-to-erc bridge to a mediator implementation on top of AMB
*/
function migrateToMediator() external {
    bytes32 REQUIRED_BLOCK_CONFIRMATIONS = 0x916daedf6915000ff68ced2f0b6773fe6f2582237f92c3c95bb4d79407230071; // keccak256(abi.encodePacked("requiredBlockConfirmations"))
    bytes32 GAS_PRICE = 0x55b3774520b5993024893d303890baa4e84b1244a43c60034d1ced2d3cf2b04b; // keccak256(abi.encodePacked("gasPrice"))
    bytes32 DEPLOYED_AT_BLOCK = 0xb120ceec05576ad0c710bc6e85f1768535e27554458f05dcbb5c65b8c7a749b0; // keccak256(abi.encodePacked("deployedAtBlock"))
    bytes32 HOME_FEE_STORAGE_KEY = 0xc3781f3cec62d28f56efe98358f59c2105504b194242dbcb2cc0806850c306e7; // keccak256(abi.encodePacked("homeFee"))
    bytes32 FOREIGN_FEE_STORAGE_KEY = 0x68c305f6c823f4d2fa4140f9cf28d32a1faccf9b8081ff1c2de11cf32c733efc; // keccak256(abi.encodePacked("foreignFee"))
    bytes32 VALIDATOR_CONTRACT = 0x5a74bb7e202fb8e4bf311841c7d64ec19df195fee77d7e7ae749b27921b6ddfe; // keccak256(abi.encodePacked("validatorContract"))

    bytes32 storageAddress = 0x131ab4848a6da904c5c205972a9dfe59f6d2afb8c9c3acd56915f89558369213; // keccak256(abi.encodePacked("migrationToMediator"))
    require(!boolStorage[storageAddress]);

    // Assign new AMB parameters
    _setBridgeContract(0x0000000000000000000000000000000000000000); // TODO set AMB bridge address when deployed on ETC
    _setMediatorContractOnOtherSide(0x0cB781EE62F815bdD9CD4c2210aE8600d43e7040);
    _setRequestGasLimit(0); // TODO define gas limit amount for foreign handleBridgedTokens method
    setNonce(keccak256(abi.encodePacked(address(this))));

    // Update fee parameters
    address feeManagerAddress = 0x0000000000000000000000000000000000000000; // TODO set new fee manager address
    addressStorage[FEE_MANAGER_CONTRACT] = feeManagerAddress;
    _setFee(feeManagerAddress, uintStorage[FOREIGN_FEE_STORAGE_KEY]);

    address[] memory list = new address[](1);
    list[0] = 0xCB576F21D18BD112D021813d4D6802d54Cacc0A0;
    _initializeRewardAccounts(list);

    // Free old storage
    delete addressStorage[VALIDATOR_CONTRACT];
    delete uintStorage[GAS_PRICE];
    delete uintStorage[DEPLOYED_AT_BLOCK];
    delete uintStorage[REQUIRED_BLOCK_CONFIRMATIONS];
    delete uintStorage[HOME_FEE_STORAGE_KEY];
    delete uintStorage[FOREIGN_FEE_STORAGE_KEY];

    // Free storage related to proxy size returns.
    delete uintStorage[0x5e16d82565fc7ee8775cc18db290ff4010745d3fd46274a7bc7ddbebb727fa54]; // keccak256(abi.encodePacked("dataSizes", bytes4(keccak256("signature(bytes32,uint256)"))))
    delete uintStorage[0x3b0a1ac531be1657049cf649eca2510ce9e3ef7df1be26d5c248fe8b298f4374]; // keccak256(abi.encodePacked("dataSizes", bytes4(keccak256("message(bytes32)"))))

    boolStorage[storageAddress] = true;
}
```

##### Upgrade Method for foreign
```solidity
/**
* @dev Method to migrate foreign WETC native-to-erc bridge to a mediator implementation on top of AMB
*/
function migrateToMediator() external {
    bytes32 REQUIRED_BLOCK_CONFIRMATIONS = 0x916daedf6915000ff68ced2f0b6773fe6f2582237f92c3c95bb4d79407230071; // keccak256(abi.encodePacked("requiredBlockConfirmations"))
    bytes32 GAS_PRICE = 0x55b3774520b5993024893d303890baa4e84b1244a43c60034d1ced2d3cf2b04b; // keccak256(abi.encodePacked("gasPrice"))
    bytes32 DEPLOYED_AT_BLOCK = 0xb120ceec05576ad0c710bc6e85f1768535e27554458f05dcbb5c65b8c7a749b0; // keccak256(abi.encodePacked("deployedAtBlock"))
    bytes32 HOME_FEE_STORAGE_KEY = 0xc3781f3cec62d28f56efe98358f59c2105504b194242dbcb2cc0806850c306e7; // keccak256(abi.encodePacked("homeFee"))
    bytes32 FOREIGN_FEE_STORAGE_KEY = 0x68c305f6c823f4d2fa4140f9cf28d32a1faccf9b8081ff1c2de11cf32c733efc; // keccak256(abi.encodePacked("foreignFee"))
    bytes32 VALIDATOR_CONTRACT = 0x5a74bb7e202fb8e4bf311841c7d64ec19df195fee77d7e7ae749b27921b6ddfe; // keccak256(abi.encodePacked("validatorContract"))

    bytes32 storageAddress = 0x131ab4848a6da904c5c205972a9dfe59f6d2afb8c9c3acd56915f89558369213; // keccak256(abi.encodePacked("migrationToMediator"))
    require(!boolStorage[storageAddress]);

    // Assign new AMB parameters
    _setBridgeContract(0x0000000000000000000000000000000000000000); // TODO set AMB bridge address when deployed on ETH
    _setMediatorContractOnOtherSide(0x073081832b4ecdce79d4d6753565c85ba4b3bea9);
    _setRequestGasLimit(0); // TODO define gas limit amount for home handleBridgedTokens method
    setNonce(keccak256(abi.encodePacked(address(this))));

    // Update fee parameters
    address feeManagerAddress = 0x0000000000000000000000000000000000000000; // TODO set new fee manager address
    addressStorage[FEE_MANAGER_CONTRACT] = feeManagerAddress;
    _setFee(feeManagerAddress, uintStorage[HOME_FEE_STORAGE_KEY]);

    address[] memory list = new address[](1);
    list[0] = 0xCB576F21D18BD112D021813d4D6802d54Cacc0A0;
    _initializeRewardAccounts(list);

    // Free old storage
    delete addressStorage[VALIDATOR_CONTRACT];
    delete uintStorage[GAS_PRICE];
    delete uintStorage[DEPLOYED_AT_BLOCK];
    delete uintStorage[REQUIRED_BLOCK_CONFIRMATIONS];
    delete uintStorage[HOME_FEE_STORAGE_KEY];
    delete uintStorage[FOREIGN_FEE_STORAGE_KEY];

    boolStorage[storageAddress] = true;
}
```

contracts/upgradeable_contracts/BasicAMBMediator.sol
patitonar: You are right, storing as `bytes32` will save some gas. The issue is that in the `EternalStorage` there is no mapping for storing `mapping(bytes32 => bytes32)` that's why we use the workaround to store it as `bytes` in order to not modify the storage layout.

k1rill-fedoseev: We can modify `EternalStorage` by adding a `mapping(bytes32 => bytes32)` at the end of the file? This shouldn't change the storage layout.

k1rill-fedoseev: Ok, this is not possible if the plan is to upgrade current bridge contracts to the new mediator contracts.

k1rill-fedoseev: By the way, can't we use `uintStorage` for storing `bytes32` values instead?

akolotov:
> The proxy contract being used is `ClassicEternalStorageProxy`. It has the restriction that for every call delegated to the implementation contract it will return a fixed-size value of `32` unless we specified the size for a method signature as we did for `signature(bytes32,uint256)` and `message(bytes32)`.
> 
> In this case, it affects the method `rewardAccounts()` which returns the list of reward accounts that will get the fees that has a variable size. It doesn't affect the operations, fees are correctly distributed, but it is not possible to perform an external call if someone wants to know the list. That's why I implemented `ClassicHomeAMBNativeToErc20` that maintains the size to be returned by the proxy so it will work correctly with the `ClassicEternalStorageProxy` proxy contract.

Thanks for highlighting this. I need to think more if it does make sense still have separate versions of contracts.

contracts/upgradeable_contracts/BasicAMBMediator.sol
patitonar: Thanks! Storing it in `uintStorage` saves some gas. I updated how the nonce is stored. Regarding `setTransactionHash()` and `setFailedMessageDataHash`, as they are part of AMB bridge itself, I think it makes sense to create an issue for this improvement to properly evaluate and apply the changes separately.  https://github.com/poanetwork/tokenbridge-contracts/pull/367/commits/4a615de4b340218245f154faeb5c22c7b14195bf

k1rill-fedoseev: Great. Created #373

patitonar:
In order to use the fee manager contract as a separate contract, the mediator contract will only store the address of the fee manager contract, and so the fee and the reward accounts list will be stored in the fee manager contract.

In initialization terms, this means that a new proxy will be deployed along with the fee manager contract, and then it will be initialized defining an owner, the fee and the reward accounts.
In the mediator, when initializing, the fee manager proxy address will be used or set to zero in case this feature is not needed in the `initialize` method, so `rewardableInitialize` is no longer needed.

The fee manager will accept external calls to return the stored fee, calculate the fee amount of a value using the stored fee and return the list of reward accounts. Also, it will accept external calls from the owner to update the fee and add/remove accounts from the list.

For the fee calculation, the method with the logic will be defined in the fee manager contract, so the mediator will perform a regular call to get the calculated fee amount.

The fee distribution method needs to be executed in the context(with the state) of the mediator contract so it can access the tokens. I think the best option is to define the `distributeFee` method in the mediator contract and interact with the fee manager to iterate the reward accounts. One advantage of this is that the fee manager will be generic enough to only have one version instead of having one for home and one for foreign. 

With this solution, we don't need a special mediator contract for the existing classic proxy because the `rewardAccounts` method will be in the fee manager contract using a regular proxy contract. Regarding the POA-POA20 bridge, the migration process and state variables updates require further analysis but so far I wasn't able to identify any restriction for which this solution won't work since the proxy contract used doesn't have any special handling and the rest of the code is upgradable.

Applying these changes doesn't require big efforts in the contracts since most of the methods are already implemented. First, we need to revert https://github.com/poanetwork/tokenbridge-contracts/pull/367/commits/8cc1e65b5465dcd8b09bfe4713d8a60fca95b71f. Then most of the work will be to move some methods, update the visibility/modifiers and update the initialization methods. The initialization of some of the unit tests will be affected and also the deployment script needs to be updated to deploy a new proxy and initialize the fee manager.

akolotov:
Thanks for investigation! Below are my comments.

> In initialisation terms, this means that a new proxy will be deployed along with the fee manager contract, and then it will be initialised defining an owner, the fee and the reward accounts.

What if we will not use the proxy template here at all and will refer to the implementation? What drawbacks of this solution do you see?

> The fee distribution method needs to be executed in the context(with the state) of the mediator contract so it can access the tokens. I think the best option is to define the `distributeFee` method in the mediator contract and interact with the fee manager to iterate the reward accounts.

It does not look as a straightforward way. Can you consider another options which can be implemented? As soon as the mediator needs to operate with fees, it sends the corresponding number of tokens to the fee manager and calls a method (`onTokenTransfer`?) from the fee manager contract. The fee manager contract will go through the list of reward recipients and send tokens to them. It will allow us: 
  - to use EOA instead of a manager
  - to not implement any logic to distribute fees and collect fees on the fee manager contract instead.

Both options are useful on early stage of migration to AMB.

This option could allow us still keep the logic of fee distribution on the fee manager side as so different actions with fees are possible. For example,
  - Instead of transfer tokens by the fee manager contract, the reward recipients could call a special method on the fee manager contract to withdraw the fee. It will allow to increase number of fee recipients and safe gas for the bridge validators.
  - Fee could be sent back to the bridge automatically as so the corresponding amount of tokens appeared on another side of the bridge.

patitonar:
Updated the implementation with the recommendation from the comments.
The fee manager is used as a separate contract and without a proxy. For fee distribution, the mediator sends/Mint the tokens to the fee manager and calls `onTokenTransfer` which will go through the list of reward recipients and send tokens to them.

> It will allow us:
>- to use EOA instead of a manager
>- to not implement any logic to distribute fees and collect fees on the fee manager contract instead.

In the way it is implemented right now I don't think we will be able to use an EOA instead of a manager. The fee parameter and the logic to calculate the fee of a transfer are stored/handled in the fee manager.


### Updated migration methods

##### Upgrade Method for home

Deploy `HomeFeeManagerAMBNativeToErc20` in ETC and initialize it with the following parameters
- `0x131c608EE32e18430eDb439296d1d40819b509E9` // Current upgradeabilityOwner of the bridge
- `2000000000000000` // Fee value
- `[0xCB576F21D18BD112D021813d4D6802d54Cacc0A0]` // Reward address value

```solidity
/**
* @dev Method to migrate home WETC native-to-erc bridge to a mediator implementation on top of AMB
*/
function migrateToMediator() external {
    bytes32 REQUIRED_BLOCK_CONFIRMATIONS = 0x916daedf6915000ff68ced2f0b6773fe6f2582237f92c3c95bb4d79407230071; // keccak256(abi.encodePacked("requiredBlockConfirmations"))
    bytes32 GAS_PRICE = 0x55b3774520b5993024893d303890baa4e84b1244a43c60034d1ced2d3cf2b04b; // keccak256(abi.encodePacked("gasPrice"))
    bytes32 DEPLOYED_AT_BLOCK = 0xb120ceec05576ad0c710bc6e85f1768535e27554458f05dcbb5c65b8c7a749b0; // keccak256(abi.encodePacked("deployedAtBlock"))
    bytes32 HOME_FEE_STORAGE_KEY = 0xc3781f3cec62d28f56efe98358f59c2105504b194242dbcb2cc0806850c306e7; // keccak256(abi.encodePacked("homeFee"))
    bytes32 FOREIGN_FEE_STORAGE_KEY = 0x68c305f6c823f4d2fa4140f9cf28d32a1faccf9b8081ff1c2de11cf32c733efc; // keccak256(abi.encodePacked("foreignFee"))
    bytes32 VALIDATOR_CONTRACT = 0x5a74bb7e202fb8e4bf311841c7d64ec19df195fee77d7e7ae749b27921b6ddfe; // keccak256(abi.encodePacked("validatorContract"))

    bytes32 migrationToMediatorStorage = 0x131ab4848a6da904c5c205972a9dfe59f6d2afb8c9c3acd56915f89558369213; // keccak256(abi.encodePacked("migrationToMediator"))
    require(!boolStorage[migrationToMediatorStorage]);

    // Assign new AMB parameters
    _setBridgeContract(0x0000000000000000000000000000000000000000); // TODO set AMB bridge address when deployed on ETC
    _setMediatorContractOnOtherSide(0x0cB781EE62F815bdD9CD4c2210aE8600d43e7040);
    _setRequestGasLimit(0); // TODO define gas limit amount for foreign handleBridgedTokens method
    setNonce(keccak256(abi.encodePacked(address(this))));

    // Update fee manager
    addressStorage[FEE_MANAGER_CONTRACT] = 0x0000000000000000000000000000000000000000; // TODO set new fee manager address

    // Free old storage
    delete addressStorage[VALIDATOR_CONTRACT];
    delete uintStorage[GAS_PRICE];
    delete uintStorage[DEPLOYED_AT_BLOCK];
    delete uintStorage[REQUIRED_BLOCK_CONFIRMATIONS];
    delete uintStorage[HOME_FEE_STORAGE_KEY];
    delete uintStorage[FOREIGN_FEE_STORAGE_KEY];

    // Free storage related to proxy size returns.
    delete uintStorage[0x5e16d82565fc7ee8775cc18db290ff4010745d3fd46274a7bc7ddbebb727fa54]; // keccak256(abi.encodePacked("dataSizes", bytes4(keccak256("signature(bytes32,uint256)"))))
    delete uintStorage[0x3b0a1ac531be1657049cf649eca2510ce9e3ef7df1be26d5c248fe8b298f4374]; // keccak256(abi.encodePacked("dataSizes", bytes4(keccak256("message(bytes32)"))))

    boolStorage[migrationToMediatorStorage] = true;
}
```

##### Upgrade Method for foreign

Deploy `ForeignFeeManagerAMBNativeToErc20` in ETH and initialize it with the following parameters
- `0x131c608EE32e18430eDb439296d1d40819b509E9` // Current upgradeabilityOwner of the bridge
- `2000000000000000` // Fee value
- `[0xCB576F21D18BD112D021813d4D6802d54Cacc0A0]`  // Reward address value
- `0x86aaBCc646f290b9Fc9Bd05CE17C3858d1511Da1`  // Token address

```solidity
/**
* @dev Method to migrate foreign WETC native-to-erc bridge to a mediator implementation on top of AMB
*/
function migrateToMediator() external {
    bytes32 REQUIRED_BLOCK_CONFIRMATIONS = 0x916daedf6915000ff68ced2f0b6773fe6f2582237f92c3c95bb4d79407230071; // keccak256(abi.encodePacked("requiredBlockConfirmations"))
    bytes32 GAS_PRICE = 0x55b3774520b5993024893d303890baa4e84b1244a43c60034d1ced2d3cf2b04b; // keccak256(abi.encodePacked("gasPrice"))
    bytes32 DEPLOYED_AT_BLOCK = 0xb120ceec05576ad0c710bc6e85f1768535e27554458f05dcbb5c65b8c7a749b0; // keccak256(abi.encodePacked("deployedAtBlock"))
    bytes32 HOME_FEE_STORAGE_KEY = 0xc3781f3cec62d28f56efe98358f59c2105504b194242dbcb2cc0806850c306e7; // keccak256(abi.encodePacked("homeFee"))
    bytes32 FOREIGN_FEE_STORAGE_KEY = 0x68c305f6c823f4d2fa4140f9cf28d32a1faccf9b8081ff1c2de11cf32c733efc; // keccak256(abi.encodePacked("foreignFee"))
    bytes32 VALIDATOR_CONTRACT = 0x5a74bb7e202fb8e4bf311841c7d64ec19df195fee77d7e7ae749b27921b6ddfe; // keccak256(abi.encodePacked("validatorContract"))

    bytes32 migrationToMediatorStorage = 0x131ab4848a6da904c5c205972a9dfe59f6d2afb8c9c3acd56915f89558369213; // keccak256(abi.encodePacked("migrationToMediator"))
    require(!boolStorage[migrationToMediatorStorage]);

    // Assign new AMB parameters
    _setBridgeContract(0x0000000000000000000000000000000000000000); // TODO set AMB bridge address when deployed on ETH
    _setMediatorContractOnOtherSide(0x073081832b4ecdce79d4d6753565c85ba4b3bea9);
    _setRequestGasLimit(0); // TODO define gas limit amount for home handleBridgedTokens method
    setNonce(keccak256(abi.encodePacked(address(this))));

    // Update fee manager
    addressStorage[FEE_MANAGER_CONTRACT] = 0x0000000000000000000000000000000000000000; // TODO set new fee manager address

    // Free old storage
    delete addressStorage[VALIDATOR_CONTRACT];
    delete uintStorage[GAS_PRICE];
    delete uintStorage[DEPLOYED_AT_BLOCK];
    delete uintStorage[REQUIRED_BLOCK_CONFIRMATIONS];
    delete uintStorage[HOME_FEE_STORAGE_KEY];
    delete uintStorage[FOREIGN_FEE_STORAGE_KEY];

    boolStorage[migrationToMediatorStorage] = true;
}
```

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
```diff
@@ -1,164 +1,108 @@
...
-        address firstAccount = getNextRewardAccount(F_ADDR);
-        // if list wasn't initialized
-        if (firstAccount == address(0)) {
-            firstAccount = F_ADDR;
-        }
-        setNextRewardAccount(_account, firstAccount);
-        setNextRewardAccount(F_ADDR, _account);
-        setRewardAccountsCount(rewardAccountsCount().add(1));
+        require(rewardAccounts.length.add(1) < MAX_REWARD_ACCOUNTS);
+        rewardAccounts.push(_account);
```
akolotov: Could you provide more info why did you choose to return back to the array instead of the linked list?

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
```diff
@@ -1,164 +1,108 @@
...
-
-            require(nextRewardAccount != address(0));
-        }
-
-        return list;
+    constructor(address _owner, uint256 _fee, address[] _rewardAccountList) public {
+        require(_rewardAccountList.length > 0 && _rewardAccountList.length <= MAX_REWARD_ACCOUNTS);
+        _transferOwnership(_owner);
+        _setFee(_fee);
+        rewardAccounts = _rewardAccountList;
```
akolotov: What will happen if one of the accounts is the mediator address?

akolotov:
@k1rill-fedoseev please take a look. I know that the fee manager will be required for your activity related to implementation of a pair of mediators for the STAKE token.

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
patitonar: Great question!

I performed some tests. Added the home mediator address plus one regular account as reward addresses in home and repeating the same in foreign. Fee amount 0.1%

Test 1, Home to Foreign transfer:
- Transferred 10 native tokens to home mediator.
- In foreign 9.99 tokens were minted to the user and 0.01 to the fee manager
- onTokenTransfer method of fee manager is invoked and the fee manager distributes the fee between the accounts, it transfers the fee to the mediator and calls onTokenTransfer of the mediator. This method tries to bridge the token to home side but in this test, the amount was smaller than minPerTx(), so it fails, and because of this [check](https://github.com/poanetwork/tokenbridge-contracts/blob/443ad4c3db5f2f38d5f9e8437812a314b8454243/contracts/ERC677BridgeToken.sol#L62) the token transfer of the fee is reverted and so the whole `onTokenTransfer` method invocation to the fee manager reverts. So the token fee amount remains locked in the fee manager contract.

Explorer [link](https://kovan.etherscan.io/tx/0xa17ca781613542b741bd8f48f7f4054e7a6d6c047ec5ef858bd28ee0d23d6415
) to the transaction of test 1
[Link](https://dashboard.tenderly.dev/tx/kovan/0xa17ca781613542b741bd8f48f7f4054e7a6d6c047ec5ef858bd28ee0d23d6415) to debug the transaction of test 1 in tenderly

Test 2. This time minPerTx() is set to `1 wei`. Home to Foreign transfer:
- Transferred `10` native tokens to home mediator.
- In foreign `9.99` tokens were minted to the user and 0.01 to the fee manager
- onTokenTransfer method of fee manager is invoked and the fee manager distributes the fee between the accounts, it transfers the fee to the mediator and calls onTokenTransfer of the mediator. This time the amount is within the limits so it is able to send the message to the amb bridge and burn the tokens. The message specifies as the recipient the fee manager address from foreign side and the amount of `0.005`

Explorer [link](https://kovan.etherscan.io/tx/0xe707c0a96c00860ca36d63a4982ab270d182f5d33644d189b0ea611b3a6ace7a
) to the transaction of test 2
[Link](https://dashboard.tenderly.dev/tx/kovan/0xe707c0a96c00860ca36d63a4982ab270d182f5d33644d189b0ea611b3a6ace7a) to debug the transaction of test 2 in tenderly

Test 3. Continuation of test 2
- From test 2, `0.005` tokens were transferred to home with the fee manager address from foreign as the recipient.
- In home, `0.004995` are transferred to the foreign fee manager address (which no one has access to in home)
- `0,000005` is transferred to the home fee manager.
- onTokenTransfer method of fee manager is invoked and the fee manager distributes the fee between the accounts. The fee manager tries to send `0.0000025` native tokens to the mediator but the `Send` fails (I think because of gas required by the fallback method) and so the `Sacrifice` contract was deployed and self-destructed which allowed allocating the fee amount in the mediator contract but it does not invoke any method.
- Now, that `0.0000025` is locked in the mediator contract but no ERC677 token representation will be minted for that amount.
Home Mediator locked balance: `9.9950025`
Total Supply ERC677 in foreign: `9.995`
Explorer [link](https://blockscout.com/poa/sokol/tx/0xffda1763b5c1cd4a5e574759eb3e82eabdff199193855968189e6ca5589c53c8
) to the transaction of test 3

akolotov: Well... What does it mean? Should we avoid situations when the mediator is the reward receiver? Which properties should not the reward receiver have to get the reward without issues?

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
patitonar: Since we don't have to use eternal storage in this contract I saw the opportunity to try using an array.
Using an array means only having one variable to store all the data of the list of accounts against using a mapping, a uint variable to store the count and defining a F_ADDR constant. Also, it looks to me that it is less error-prone, it is not needed to build the list manually, it is easier to iterate and overall it results in a little more readable code.

Regarding gas usage, I compared using an array vs a [linkedList](https://gist.github.com/patitonar/fdcff47b13d7936997c95c8a34e30651)

| Method        | Array           | LinkedList |
| ------------- |:-------------:| -----:|
| Deploy (initializes list with 3 accounts)     | 947903 | 1028470 |
| addRewardAccount      | 52482 | 55526 |
| removeRewardAccount | 32351 | 25628 |
| distributeFee | 105385 | 105911 |

The most used method which will be `distributeFee` has a similar gas usage.

What do you think?
I'm OK using a LinkedList which will make sense for consistency since we use it in all of the other contracts.

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
patitonar: > What does it mean?

To summarize:
If home mediator is used as reward account in home, there will be native tokens locked in the mediator that won't be able to be used.

If foreign mediator is used as reward account in foreign:
- Fee distribution can fail if the fee amount is not within limits (maybe too small or daily limit reached) and the fee will be locked in the fee manager and won't be able to be used
- If amount is within limits, it will send the tokens back to home side. The native tokens in home will be transferred to an account (address of fee manager in foreign) that can't be used. 

>Which properties should not the reward receiver have to get the reward without issues?

To get rewards without issues:
- In the case of home native tokens, it will receive the rewards anyways but ideally, to save gas, the contract should have an empty fallback method so it does not exceed the gas usage provided in the `send` method.
- In the case of foreign erc677 tokens, the only requirement applies to the mediator contract as reward and it is that the `onTokenTransfer` method should not revert. Any other contract that is not the mediator contract can safely not have the `onTokenTransfer` or can revert without affecting the token transfer.

> Should we avoid situations when the mediator is the reward receiver?

So I think the best option is to avoid using the mediator contracts as reward accounts.
Should we add special checks for this in the contract?

One option to send back to home some amount of fee tokens after a Home-to-foreign transfer, if we want to support that case, could be to accumulate the fees in a separate reward receiver contract that will have an `onTokenTransfer` method and once the amount is within limits it can call `transferAndCall(mediatorAddress, accumulatedFees, validRecipientInHome)`

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
akolotov: Thanks for the explanation. 

I agree that we can use the array here instead of linked list. I inspected the code of the `BaseMediatorFeeManager` contract and found that random access to any element of the list is not required. The only place where `isRewardAccount` is used can be optimized to reduce number of loop iterations. The operation `removeRewardAccount` is also is not going to be used often so, we can close eyes that its efficiency a bit higher.

I also imagined the bridge where the Fee Manager will be used and found that it does not make sense to spend resources now to carefully introduce and test the linked list:
  * WETC bridge -- we have only one validator. The usage of the linked list here is significant overhead.
  * POA bridge -- it is not clear if the fees will be collected there at all. 
  * STAKE bridge -- the fee will be distributed by using a consensus contract (https://www.staketoken.net/token-model/reward-mechanics), the bridge will have no information about the stakers at all.

That's why we will leave this functionality as is -- with the array of the reward addresses.

contracts/upgradeable_contracts/BaseMediatorFeeManager.sol
akolotov: Thanks for very good tests you did. They reveal few issues which we need to address:

1. Since we have one-one correspondence between the Fee Manager and the Mediator contract, it makes sense to protect the system to not allow to add the mediator contract address as the reward receiver.

2. Until we have support of multiple transfer requests (https://github.com/poanetwork/tokenbridge-contracts/issues/307) in one transaction let's try to block second, third etc requests on the mediator contract. I would imagine the case when few contracts which are registered as the reward accounts, try to pass the message through the bridge.

3. We need to develop a recovery mechanism to be able to solve the bridge imbalance appearing in the cases similar to described in the test 3. Any contract could use `selfdestruct`-approach so the mediator fallback will not be called. In this case even if the contract on the Home side does not exist any more it is still possible that the corresponding contract (created with `CREATE2` EVM-operation) on the Foreign side is still exist. I would suggest a method that can be called only by the upgradability admin, this method will compare the balance of the bridge mediator contract with the value stored in the contract (we need to introduce such value) and if they differ, allow to send request to another side of the bridge to mint a number of tokens requested in the method. Do you think it will be secure?

patitonar: >Do you think it will be secure?

It looks good to me. One other possible idea could be to extend `claimTokens` method of home side, so instead of sending a request to mint the difference on the other side, send the difference of native tokens to an account provided by parameter in the method

akolotov: > instead of sending a request to mint the difference on the other side, send the difference of native tokens to an account provided by parameter in the method

Yes, from security point of view it is identical actions.

contracts/upgradeable_contracts/MediatorMessagesGuard.sol
```diff
@@ -0,0 +1,74 @@
...
+    /**
+    * @dev Remove the lock to limit the number of messages to send to the AMB bridge
+    */
+    function unlockBridgeMessages() internal {
+        setBridgeMessageLock(false);
+    }
+
+    modifier bridgeMessageAllowed {
+        if (bridgeMessageLock()) {
+            uint256 nextMessageCount = bridgeMessageCount().add(1);
```
akolotov, line 67: could you share more thoughts about the cases when we need to allow 2 or 5 messages? My question is because I think that these cases will not be used in the WETC/POA bridges but we introduce the logic which will consume additional gas.

patitonar: You are correct, I think those cases won't be used too. Here I simplified the solution to reduce the gas consumed https://github.com/poanetwork/tokenbridge-contracts/pull/367/commits/d8b837953e62c356a57c0e44d8499e51fcb2ad36

contracts/upgradeable_contracts/MediatorMessagesGuard.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+    * @dev Remove the lock to limit the number of messages to send to the AMB bridge
+    */
+    function unlockBridgeMessages() internal {
+        setBridgeMessageLock(false);
+    }
+
+    modifier bridgeMessageAllowed {
+        if (bridgeMessageLock()) {
```
akolotov, line 65: why do we need to check both if it is locked and if the limit is reached? Will be enough to check if it locked only?

patitonar: On a transfer from Home to Foreign, on the foreign side, the lock is enabled before the fee is distributed and removed after it. So if the rewards receivers try to send the tokens to the mediator to send them back to home, only one will be able to do it and the rest won't because the limit was reached. If we only check for the lock, no reward receiver will be able to send the tokens to the mediator. Is this the expected behavior? Am I missing something?

patitonar:
Added the event `TokensBridged` to be emitted when `handleBridgedTokens` is called in this pair of mediators. https://github.com/poanetwork/tokenbridge-contracts/pull/367/commits/6e611e6832296dcac41b4512d9c8a33a71a8b0b1

I think we should also use it in the other mediators. If the definition looks good, I can apply the changes for the other mediators here too. (or probably is better in a new PR with only those changes?)

contracts/upgradeable_contracts/MediatorMessagesGuard.sol
patitonar: Thanks for the suggestion! I applied the changes https://github.com/poanetwork/tokenbridge-contracts/pull/367/commits/69a612e0e8f870b438fa5a3e8e8227f615fb8afd Please notice I use the `|` operator to correctly make it work

patitonar:
>Yes, let's have them in a separate PR.

Created https://github.com/poanetwork/tokenbridge-contracts/pull/409 to add the event to other mediators.

patitonar:
Applied a similar fix from https://github.com/poanetwork/tokenbridge-contracts/pull/408 to these mediators too https://github.com/poanetwork/tokenbridge-contracts/pull/367/commits/3c7a0fc6748cda7376c3071a46a7e74ad952f9bb

PR #372, opened 2020-02-21
Chai integration internal audit fixes

k1rill-fedoseev:


contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -252,25 +256,28 @@ contract ChaiConnector is Ownable, ERC20Bridge {
...
 
+            newInvestedAmountInDai = redeemed < invested ? invested - redeemed : 0;
             // Make sure that at least requested amount was withdrawn
             require(redeemed >= amount);
-
-            setInvestedAmointInDai(redeemed < invested ? invested - redeemed : 0);
         }
 
-        require(dsrBalance() >= investedAmountInDai());
+        setInvestedAmountInDai(newInvestedAmountInDai);
```
varasev: If the `amount` is `0` and `invested` is not zero, this call will reset the `InvestedAmountInDai`.

k1rill-fedoseev: Added condition on `amount > 0`, please take a look.

varasev: I'd propose to rewrite the `_convertChaiToDai` internal function as follows:

```solidity
function _convertChaiToDai(uint256 amount) internal {
    if (amount == 0) return;

    uint256 invested = investedAmountInDai();
    uint256 initialDaiBalance = daiBalance();

    // onExecuteMessage can call a convert operation with argument greater than the current invested amount,
    // in this case bridge should withdraw all invested funds
    uint256 withdrawal = amount >= invested ? invested : amount;

    chaiToken().draw(address(this), withdrawal);
    uint256 redeemed = daiBalance() - initialDaiBalance;

    // Make sure that at least withdrawal amount was withdrawn
    require(redeemed >= withdrawal);

    uint256 newInvested = invested > redeemed ? invested - redeemed : 0;
    setInvestedAmountInDai(newInvested);

    require(dsrBalance() >= newInvested);
}
```

Also, let's add unit tests for this function which would cover the cases for `amount = 0`, `amount >= investedAmountInDai()`, and `amount < investedAmountInDai() && amount > 0`.

PR #378, opened 2020-02-27
Gas token integration

k1rill-fedoseev:
This PR adds functionality of minting gas tokens inside `requireToPassMessage` call of foreign AMB bridge

### DIFF
--- a/contracts/mocks/ForeignAMBWithGasTokenMock.sol
+++ b/contracts/mocks/ForeignAMBWithGasTokenMock.sol
@@ -1,0 +1,18 @@
+pragma solidity 0.4.24;
+
+import "../upgradeable_contracts/arbitrary_message/ForeignAMBWithGasToken.sol";
+
+/**
+* @title ForeignAMBWithGasTokenMock
+* @dev Wrapper on ForeignAMB contract, which supports minting gas tokens while passing messages
+*/
+contract ForeignAMBWithGasTokenMock is ForeignAMBWithGasToken {
+    function gasToken() public pure returns (IGasToken) {
+        // Address generated in unit test, also hardcoded in GasTokenMock
+        return IGasToken(0xEC8bE1A5630364292E56D01129E8ee8A9578d7D8);
+    }
+
+    function collectGasTokens() external {
+        _collectGasTokens();
+    }
+}
--- a/contracts/upgradeable_contracts/GasTokenConnector.sol
+++ b/contracts/upgradeable_contracts/GasTokenConnector.sol
@@ -1,0 +1,96 @@
+pragma solidity 0.4.24;
+
+import "../interfaces/IGasToken.sol";
+import "../upgradeable_contracts/Ownable.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+/**
+* @title GasTokenConnector
+* @dev This logic allows to use GasToken (https://gastoken.io)
+*/
+contract GasTokenConnector is Ownable {
+    using SafeMath for uint256;
+
+    bytes32 internal constant GAS_TOKEN_TARGET_MINT_VALUE = 0x7188f1264c90762deb4dbddcab803d6487a8f108fbcdec3ff08c642d35826f9f; // keccak256(abi.encodePacked("gasTokenTargetMintValue"))
+    bytes32 internal constant GAS_TOKEN_RECEIVER = 0x6778a39b83358b10315b40840597142f1db77a44f312ed9c0e45ee7af3bbc552; // keccak256(abi.encodePacked("gasTokenReceiver"))
+    bytes4 internal constant ON_TOKEN_TRANSFER = 0xa4c0ed36; // onTokenTransfer(address,uint256,bytes)
+
+    /**
+    * @dev Updates gas token related parameters, target mint value and token receiver
+    * @param _targetMintValue target amount of token to mint/transfer during one call to _collectGasTokens
+    * @param _receiver receiver of minted gas token
+    */
+    function setGasTokenParameters(uint256 _targetMintValue, address _receiver) external onlyOwner {
+        uintStorage[GAS_TOKEN_TARGET_MINT_VALUE] = _targetMintValue;
+        addressStorage[GAS_TOKEN_RECEIVER] = _receiver;
+    }
+
+    /**
+    * @dev Updates gas token target mint value
+    * @param _targetMintValue target amount of token to mint/transfer during one call to _collectGasTokens
+    */
+    function setGasTokenTargetMintValue(uint256 _targetMintValue) external onlyOwner {
+        uintStorage[GAS_TOKEN_TARGET_MINT_VALUE] = _targetMintValue;
+    }
+
+    /**
+    * @dev Updates gas token receiver
+    * @param _receiver receiver of minted gas token
+    */
+    function setGasTokenReceiver(address _receiver) external onlyOwner {
+        addressStorage[GAS_TOKEN_RECEIVER] = _receiver;
+    }
+
+    /**
+    * @return Configured gas token target mint value
+    */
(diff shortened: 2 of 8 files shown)


contracts/mocks/ForeignAMBWithGasTokenMock.sol
```diff
@@ -0,0 +1,18 @@
...
+import "../upgradeable_contracts/arbitrary_message/ForeignAMBWithGasToken.sol";
+
+/**
+* @title ForeignAMBWithGasTokenMock
+* @dev Wrapper on ForeignAMB contract, which supports minting gas tokens while passing messages
+*/
+contract ForeignAMBWithGasTokenMock is ForeignAMBWithGasToken {
+    function gasToken() public pure returns (IGasToken) {
+        // Address generated in unit test, also hardcoded in GasTokenMock
+        return IGasToken(0xEC8bE1A5630364292E56D01129E8ee8A9578d7D8);
```
akolotov: Is it possible to assign `0x0000000000b3F879cb30FE243b4Dfee438691c04` to the gas token contract in the ganache configuration?

contracts/upgradeable_contracts/GasTokenConnector.sol
```diff
@@ -0,0 +1,96 @@
...
+
+        if (approved >= target) {
+            // tokens are transfered directly to receiver, avoiding transfer to the bridge contract
+            require(gst.transferFrom(msg.sender, receiver, target));
+        } else {
+            if (approved > 0) {
+                require(gst.transferFrom(msg.sender, address(this), approved));
+            }
+
+            gst.mint(target - approved);
```
akolotov: Could you provide more details for the following case:
1. Imagine that the bridge contract invoker minted gas token for the same gas price as the current transaction is being executed.
2. What will be evidently cheaper:
    - mint and approve the gas token in the invoker just before calling `requireToPassMessage`?
    - mint tokens and transfer them to the invoker in one transaction and trigger the invoker in another transaction?
    - do nothing as so the bridge contract will mint gas token by itself?

contracts/upgradeable_contracts/GasTokenConnector.sol
```diff
@@ -0,0 +1,96 @@
...
+    function _collectGasTokens() internal {
+        uint256 target = gasTokenTargetMintValue();
+
+        if (target == 0) return;
+
+        address receiver = gasTokenReceiver();
+        if (receiver == address(0)) return;
+
+        IGasToken gst = gasToken();
+        uint256 approved = gst.allowance(msg.sender, address(this));
```
akolotov: Do you think it will be secure enough to try transferring gas tokens from `tx.origin` if `msg.sender` allowed 0 tokens for transfer? I am considering the case when the transaction originator has gas tokens, whereas the invoked mediator calling `requireToPassMessage` has no functionality to mint/approve gas tokens.

akolotov:
Does it make sense to update oracle-e2e tests in the [tokenbridge monorepo](https://github.com/poanetwork/tokenbridge) in order to check that the oracle handles such transactions properly?

contracts/upgradeable_contracts/GasTokenConnector.sol
k1rill-fedoseev: So, I've created a simple invoker for tests, the results are the following, assuming bridge requires `0.5 GST2` to provers a request:
* invoker just calls `requireToPassMessage`, bridge does all minting work: `1884740` gas
* invoker mints tokens by himself, then approves them and calls `requireToPassMessage`: `1893440` gas
* someone else mints gas tokens, then transfers it to the invoker, then invoker calls `requireToPassMessage`: `1860634` + `51759` + `45790` = `1958183` gas

In general, just calling `requireToPassMessage` is the simplest/cheapest way. But for more complex scenarios, where caller has tight bounds on gas limit for `requireToPassMessage` call or already have some gas tokens, the approach `approve + requireToPassMessage` in a separate transaction should be preferred.

contracts/mocks/ForeignAMBWithGasTokenMock.sol
k1rill-fedoseev: Seems like, in a current ganache version there is no easy way to do it. There is an open issue for such functionality: https://github.com/trufflesuite/ganache-cli/issues/715.

However, I was able to find a particular address and nonce used to deploy this contract. So now I can deploy a contract mock at `0x0000000000b3F879cb30FE243b4Dfee438691c04` in ganache even if I don't know the private key for deployer address.
For our case, I think, it is enough to use a slightly modified contract mock at a different address.

contracts/upgradeable_contracts/GasTokenConnector.sol
k1rill-fedoseev: I think, we can keep this logic in mediator. It will be always possible for mediator to do something like:
```solidity
function pass(bytes memory _data, uint256 _gas) external {
  transferFrom(tx.origin, address(this), 50);
  approve(address(bridge()), 50);
  bridge().requireToPassMessage(address(this), _data, _gas);
}
```
However, if we include such logic in the bridge, it could make such interactions cheaper.
I don't see any security flaws in a case, when the bridge will try to use allowance from `tx.origin`.

contracts/mocks/ForeignAMBWithGasTokenMock.sol
akolotov: OK. It is up to you. Just thought that it makes sense to reduce number of changes in the future that would be required if the change the original contract (`ForeignAMBWithGasToken`).

contracts/upgradeable_contracts/GasTokenConnector.sol
akolotov: Yes, keep it in the mediator is a good option.

contracts/upgradeable_contracts/GasTokenConnector.sol
akolotov: Great research! Thanks! Could you create a page under `https://docs.tokenbridge.net/amb-bridge/` where describe the main idea about the gas token usage in AMB and describe these scenarios and the scenario with `tx.origin`?

k1rill-fedoseev: Yes, sure. I will start to write it today

contracts/mocks/ForeignAMBWithGasTokenMock.sol
k1rill-fedoseev: For now, I will keep it as it is.
Later, when this functionality appears in ganache, it will make sense to refactor tests. So that we can use the hardcoded addresses from the mainnet, instead of mocking it every time. 

There are also similar mocks related to other functionality (sai/dai, chai). It make sense to consider refactoring them as well.

k1rill-fedoseev:
For now, I don't think that we need explicit e2e tests for oracle. 
From the oracle point of view, there should not be any difference in contract behaviour. It won't catch any `Transfer` events from gas token contract, so its behaviour will be identical.

contracts/upgradeable_contracts/GasTokenConnector.sol
```diff
@@ -0,0 +1,96 @@
...
+import "../interfaces/IGasToken.sol";
+import "../upgradeable_contracts/Ownable.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+/**
+* @title GasTokenConnector
+* @dev This logic allows to use GasToken (https://gastoken.io)
+*/
+contract GasTokenConnector is Ownable {
+    using SafeMath for uint256;
```
patitonar: It looks like we are not using `SafeMath` methods here, I think we can remove the `import` and `using` statements

k1rill-fedoseev: Yes, you are right. I removed it

PR #382, opened 2020-03-13
Update interest receiver

k1rill-fedoseev:
This PR optimizes `InterestReceiver` for usage without `EternalStorage`, since it will be deployed without proxy contract. 
Continues #380

### DIFF
--- a/contracts/upgradeable_contracts/InterestReceiver.sol
+++ b/contracts/upgradeable_contracts/InterestReceiver.sol
@@ -2,6 +2,7 @@ pragma solidity 0.4.24;
 
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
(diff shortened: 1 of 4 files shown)


contracts/upgradeable_contracts/InterestReceiver.sol
```diff
@@ -42,8 +39,6 @@ contract InterestReceiver is ERC677Receiver, Initializable, Ownable, Claimable,
     * @param _value amount of transferred tokens
     */
     function onTokenTransfer(address, uint256 _value, bytes) external returns (bool) {
```
akolotov: Since we have some time let's introduce sending tokens to the consensus contract in the xDai chain as part of `onTokenTransfer` action. We can use `relayTokens` functionality for this. The address of the receiver in the xDai chain will be specifed in the constructor and can be changed later through `setReceiverInXDai`. The `withdraw()` operation is not needed in this case.

PR #391, opened 2020-03-23
Add support of multiple bridges in bridge token

k1rill-fedoseev:
Closes https://github.com/poanetwork/tokenbridge-contracts/issues/388

### DIFF
--- a/contracts/ERC677MultiBridgeToken.sol
+++ b/contracts/ERC677MultiBridgeToken.sol
@@ -1,0 +1,108 @@
+pragma solidity 0.4.24;
+
+import "./ERC677BridgeToken.sol";
+
+/**
+ * @title ERC677MultiBridgeToken
+ * @dev This contract extends ERC677BridgeToken to support several bridge simulteniously
+ */
+contract ERC677MultiBridgeToken is ERC677BridgeToken {
+    address public constant F_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+    uint256 internal constant MAX_BRIDGES = 50;
+    mapping(address => address) public bridgePointers;
+    uint256 public bridgeCount;
+
+    event BridgeAdded(address indexed bridge);
+    event BridgeRemoved(address indexed bridge);
+
+    constructor(string _name, string _symbol, uint8 _decimals) public ERC677BridgeToken(_name, _symbol, _decimals) {
+        bridgePointers[F_ADDR] = F_ADDR; // empty bridge contracts list
+    }
+
+    /**
+    * @dev Adds one more bridge contract into the list
+    * @param _bridge bridge contract address
+    */
+    function setBridge(address _bridge) external onlyOwner {
+        require(bridgeCount < MAX_BRIDGES);
+        require(AddressUtils.isContract(_bridge));
+        require(!isBridge(_bridge));
+
+        address firstBridge = bridgePointers[F_ADDR];
+        require (firstBridge != address(0));
+        bridgePointers[F_ADDR] = _bridge;
+        bridgePointers[_bridge] = firstBridge;
+        bridgeCount = bridgeCount.add(1);
+
+        emit BridgeAdded(_bridge);
+    }
+
+    /**
+    * @dev Removes one existing bridge contract from the list
+    * @param _bridge bridge contract address
+    */
+    function unsetBridge(address _bridge) external onlyOwner {
+        require(isBridge(_bridge));
+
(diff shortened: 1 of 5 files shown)


contracts/ERC677MultiBridgeToken.sol
```diff
@@ -0,0 +1,108 @@
+pragma solidity 0.4.24;
+
+import "./ERC677BridgeToken.sol";
+
+/**
+ * @title ERC677MultiBridgeToken
+ * @dev This contract extends ERC677BridgeToken to support several bridge simulteniously
+ */
+contract ERC677MultiBridgeToken is ERC677BridgeToken {
```
akolotov: Consider to inherit this new contract by `ERC677BridgeTokenRewardable`. Please elaborate which components could be also affected after such kind of change.

contracts/ERC677MultiBridgeToken.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+    * @dev Overrides super.callAfterTransfer, uses custom checking if an address is a known bridge contract
+    * @param _from tokens sender address
+    * @param _to tokens receiver address
+    * @param _value amount of transferred tokens
+    */
+    function callAfterTransfer(address _from, address _to, uint256 _value) internal {
+        if (AddressUtils.isContract(_to) && !contractFallback(_from, _to, _value, new bytes(0))) {
+            require(!isBridge(_to));
```
akolotov: So, the difference is only this line. Does it make sense to implement something like this:

_ERC677BridgeToken_
```solidity
function callAfterTransfer(address _from, address _to, uint256 _value) internal {
    if (AddressUtils.isContract(_to) && !contractFallback(_from, _to, _value, new bytes(0))) {
        require(!isBridge(_to));
        emit ContractFallbackCallFailed(_from, _to, _value);
    }
}

function isBridge(address _address) public view returns (bool) {
    return _address == bridgeContract;
}
```
In this case in _ERC677MultiBridgeToken_ you will need to define only `isBridge` without necessity to duplicate `callAfterTransfer`.

What do you think?

contracts/ERC677MultiBridgeToken.sol
```diff
@@ -0,0 +1,108 @@
...
+
+    constructor(string _name, string _symbol, uint8 _decimals) public ERC677BridgeToken(_name, _symbol, _decimals) {
+        bridgePointers[F_ADDR] = F_ADDR; // empty bridge contracts list
+    }
+
+    /**
+    * @dev Adds one more bridge contract into the list
+    * @param _bridge bridge contract address
+    */
+    function setBridge(address _bridge) external onlyOwner {
```
varasev: Maybe name this method as `addBridge`?

contracts/ERC677MultiBridgeToken.sol
```diff
@@ -0,0 +1,108 @@
...
+        bridgeCount = bridgeCount.add(1);
+
+        emit BridgeAdded(_bridge);
+    }
+
+    /**
+    * @dev Removes one existing bridge contract from the list
+    * @param _bridge bridge contract address
+    */
+    function unsetBridge(address _bridge) external onlyOwner {
```
varasev: As it was proposed above for renaming `setBridge` to `addBridge`, maybe rename `unsetBridge` to `removeBridge`?

varasev:
@k1rill-fedoseev Thank you for the rapid changes! I left a couple of minor comments.

varasev:
Also, should we override `setBridgeContract` in `ERC677MultiBridgeToken` as follows?

```solidity
function setBridgeContract(address) external {
    revert();
}
```

Because `ERC677BridgeToken.setBridgeContract()` function and `ERC677BridgeToken.bridgeContract()` public getter make no sense in `ERC677MultiBridgeToken`.

varasev:
And maybe set overriden `ERC677MultiBridgeToken.bridgeContract` public getter so that it would return the first bridge from the list:

```solidity
function bridgeContract() public view return (address) {
    address firstBridge = bridgePointers[F_ADDR];
    return (firstBridge == F_ADDR) ? address(0) : firstBridge;
}
```

akolotov:
> And maybe set overriden `ERC677MultiBridgeToken.bridgeContract` public getter so that it would return the first bridge from the list:

There is no obvious use case for such method. So, I would recommend to overwrite it with `revert()`.

contracts/ERC677MultiBridgeToken.sol
k1rill-fedoseev: Sure, I renamed it.

PR #394, opened 2020-03-31
STAKE token mediators

k1rill-fedoseev:
This PR adds two smart-contracts that are intended to bridge STAKE token through AMB bridge between Mainnet and xDai chains.

From the implementation perspective, the regular `AMB-ERC677-TO-ERC677` can be used, with some slight changes.

The STAKE token is supposed to have a dynamic supply in the xDai chain and a non-decreasing supply in the Mainnet chain.
Also, the configured fee will be collected for transfers in the xDai => Mainnet direction.

For that, the following bridging mechanics are being introduced:
* For the Mainnet => xDai transfers, nothing is changed from the `AMB-ERC677-TO-ERC677` logic. The tokens are locked on the mediator contract in the Mainnet and then minted in the xDai chain.
* For the xDai => Mainnet transfers:
  1) Bridged tokens are burnt in the xDai chain. 
  2) The fee is calculated and the `blockReward.addBridgeTokenRewardReceivers(fee)` is called. The collected fee will be distributed between validators later in the consensus contract.
  3) Value, after fee deduction, is bridged to the Mainnet.
  4) Tokens are unlocked in the Mainnet chain. If bridge balance is less than the bridged amount, the bridge will first mint insufficient tokens, as a consequence of this, the total supply will increase in the Mainnet.


contracts/upgradeable_contracts/stake_token_mediators/HomeStakeTokenMediator.sol
```diff
@@ -0,0 +1,78 @@
...
+            if (address(_blockRewardContract()) == address(0)) {
+                // in case if block reward contract is not configured, the fee is not collected
+                passMessage(chooseReceiver(_from, _data), _value);
+            } else {
+                // when block contract is defined, the calculated fee is substructed from the original value
+                uint256 fee = calculateFee(_value);
+                passMessage(chooseReceiver(_from, _data), _value.sub(fee));
+                if (fee > 0) {
+                    // the fee itself is distributed later in the block reward contract
+                    _blockRewardContract().call(abi.encodeWithSelector(ADD_BRIDGE_TOKEN_REWARD_RECEIVERS, fee));
```
akolotov: why do we need to invoke the method through `call` rather than `_blockRewardContract. addBridgeTokenRewardReceivers`?

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,41 @@
+pragma solidity 0.4.24;
+
+import "../amb_erc677_to_erc677/ForeignAMBErc677ToErc677.sol";
+import "../../interfaces/IBurnableMintableERC677Token.sol";
+
+contract ForeignStakeTokenMediator is ForeignAMBErc677ToErc677 {
```
akolotov: Please consider to redefine `getBridgeMode` and `getBridgeInterfacesVersion` since the logic of parent contracts is changed. The same is for the home side.

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,41 @@
...
+     * @dev Mints tokens if the bridge balance is insufficient.
+     * This can happen after STAKE tokens emission in the xDai chain
+     * @param _value amount of tokens, which are requested to be withdrawn
+     */
+    function _mintLackingTokens(uint256 _value) internal {
+        IBurnableMintableERC677Token token = IBurnableMintableERC677Token(erc677token());
+        uint256 balance = token.balanceOf(address(this));
+        if (balance < _value) {
+            // if bridge balance is less than requested value, the lacking part of tokens will be minted
+            token.mint(address(this), _value - balance);
```
akolotov: why don't mint the tokens directly to the `_recipient` account it will allow to reduce gas usage?

varasev: Just in case: please note that the [`mint` function](https://github.com/xdaichain/stake-token/blob/34d948ade2cdc3ba276c103a58dae5042fd8e9fe/contracts/Token/ERC677BridgeToken.sol#L163-L172) will revert for the recipient `address(0)`.

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
k1rill-fedoseev: Addressed by a6d008d

contracts/upgradeable_contracts/stake_token_mediators/HomeStakeTokenMediator.sol
k1rill-fedoseev: We don't need. We can use regular function call here.
Addressed by a6d008d

k1rill-fedoseev:
I have tested the deployment scripts. Everything seems to work fine.
Here is the test performed:
1. Existing Sokol-Kovan AMB bridge at `0xFe446bEF1DbF7AFE24E81e05BC8B271C1BA9a560`.
2. Deploy BlockReward mock in Home chain at `0x03E2b0876625D1Cd19194498D673cCeff8D60110`
3. Deploy `ERC677BridgeTokenRewardable` token in the Home chain. [0x3061B0f53FAdfB55075bDB1fF7A6961cb592997F](https://blockscout.com/poa/sokol/address/0x3061B0f53FAdfB55075bDB1fF7A6961cb592997F/transactions)
4. Call `setBlockRewardContract` on deployed token contract.
5. Call `setToken` on deployed block reward contract.
6. Deploy [stake token](https://github.com/xdaichain/stake-token/blob/master/contracts/Token/ERC677MultiBridgeToken.sol) in the Foreign chain. [0x3cE43Ec076324E0bEcb6d9Ea47e264309f804CAc](https://kovan.etherscan.io/address/0x3cE43Ec076324E0bEcb6d9Ea47e264309f804CAc).
7. Run deploy script using `./deploy.sh`. Mediators were deployed successfully in both chains: [0xb5d68a0010F175cbced754E5635d67A4cB343388](https://blockscout.com/poa/sokol/address/0xb5d68a0010F175cbced754E5635d67A4cB343388/transactions) and [0x2c34cc8c227d1fbcD76531528208587390352f40](https://kovan.etherscan.io/address/0x2c34cc8c227d1fbcD76531528208587390352f40).
8. Call `addBridge(0xb5d68a0010F175cbced754E5635d67A4cB343388)` on token in the Foreign chain.
9. Call `addBridge(0x2c34cc8c227d1fbcD76531528208587390352f40)` and `transferOwnership(0x2c34cc8c227d1fbcD76531528208587390352f40)` on the token in the Home chain.
10. Test that tokens can be bridged in the Foreign => Home direction, [tx](https://kovan.etherscan.io/tx/0xec50279b74249cec4e7153aeb3aa7da49c282077fcfe60359db512e016f70abf).
11. Test that tokens can be bridged in the Home => Foreign direction, [tx](https://blockscout.com/poa/sokol/tx/0x5038a9a92c5b34a1c2e779053b7977e40d17145553453ad928252a76a9b553c1/token_transfers). `0.1% ` fee was collected and minted to the blockReward contract.

varasev:
@k1rill-fedoseev Could you please measure gas consumption for deploying Foreign contract implementations to ensure the size of the compiled contracts doesn't [exceed 24 Kb](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md)? (gas cannot exceed ~6 700 000 in this case).

k1rill-fedoseev:
For the foreign mediator implementation:
* Optimized bytecode size is ~ 15 Kb
* Deploy [tx](https://kovan.etherscan.io/tx/0xa43aef519846251cd9dd63be5593534e3b381eb7062b241c2a3457f2d5ef24fc) took ~ 3 200 000 gas

For the home mediator, the numbers are similar:
* Optimized bytecode size is ~ 16 Kb

contracts/upgradeable_contracts/stake_token_mediators/HomeStakeTokenMediator.sol
```diff
@@ -0,0 +1,106 @@
...
+                if (fee > 0) {
+                    // the fee itself is distributed later in the block reward contract
+                    _blockRewardContract().addBridgeTokenRewardReceivers(fee);
+                }
+            }
+        }
+    }
+
+    /**
+     * @dev Executes action on fixed tokens
```
akolotov, line 99:
```suggestion
     * @dev Executes action on relayed request to fix the failed transfer of tokens
```

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,51 @@
...
+    function executeActionOnFixedTokens(address _recipient, uint256 _value) internal {
+        IBurnableMintableERC677Token token = IBurnableMintableERC677Token(erc677token());
+        uint256 balance = token.balanceOf(address(this));
+        if (_recipient != address(0) && balance == 0) {
+            token.mint(_recipient, _value);
+        } else if (balance < _value) {
+            token.mint(address(this), _value - balance);
+            token.transfer(_recipient, _value);
+        } else {
+            token.transfer(_recipient, _value);
```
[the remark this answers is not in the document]
k1rill-fedoseev: Seems that, first we need to include such check in the mediators for `ambErcToErc`. It makes sense to check that `_from != mediatorContractOnOtherSide()` here: https://github.com/poanetwork/tokenbridge-contracts/blob/bfea462cd6830a0a7f0bf80bf3c1f8aee93cf432/contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol#L74-L84

If we introduce such check on one side, do we need to check this second time on the other side of a bridge, when the message is received?

contracts/upgradeable_contracts/stake_token_mediators/BasicStakeTokenMediator.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity 0.4.24;
+
+import "../amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol";
```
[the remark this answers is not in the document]
k1rill-fedoseev: Logically, `BasicAMBErc677ToErc677` is needed for ERC-TO-ERC bridges only. Also, STAKE bridge is a subtype of ERC-TO-ERC bridge. 
Then, I guess, it will be better to move `stake_bridge_mediators` into `amb_erc677_to_erc677` , so it will be `stake_bridge_mediators /amb_erc677_to_erc677`.

What do you think?

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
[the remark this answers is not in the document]
k1rill-fedoseev: Ok, I found that we validate this case in `chooseReceiver`. Still, do we need a second check on the other side?

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,51 @@
...
+import "../../interfaces/IBurnableMintableERC677Token.sol";
+
+contract ForeignStakeTokenMediator is BasicStakeTokenMediator {
+    /**
+     * @dev Executes action on incoming tokens from the other network
+     * @param _recipient address of tokens receiver
+     * @param _value amount of bridged tokens
+     */
+    function executeActionOnBridgedTokens(address _recipient, uint256 _value) internal {
+        uint256 value = _value.div(10**decimalShift());
```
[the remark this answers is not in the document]
k1rill-fedoseev: Do we need to include such check in `ambErcToErc` mediators also?
Is it sufficient to check this condition only once, on one side of a bridge?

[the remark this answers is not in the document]
akolotov: The question is about the unit tests. Do we have such test? Or the answer is 'no' and moreover there is no such limitation that is why you are requesting for requirements?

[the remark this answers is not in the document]
k1rill-fedoseev: I don't see the place where we are validating value to be non-zero

contracts/upgradeable_contracts/stake_token_mediators/BasicStakeTokenMediator.sol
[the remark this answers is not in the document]
akolotov: I don't think that we will need the separate directory `stake_bridge_mediators` in this case. Consider to put everything in `amb_erc677_to_erc677` without subdirectories.

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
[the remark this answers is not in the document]
akolotov: The question was for the unit test.

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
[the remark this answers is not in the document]
akolotov: So, does it mean that there is no such check neither for the `native-to-erc20` mode nor for `relayTokens` functionality for `erc20-to-native`/`erc20-to-erc20` modes and we are able to request transfer of `0` tokens?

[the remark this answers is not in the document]
k1rill-fedoseev: If I understand it correctly, the current situation is the following:
* In `relayTokens` for `erc20-to-native`/`erc20-to-erc20`, we make such check.
* In regular ERC20 transfer for `native-to-erc20`, `erc20-to-native` and `erc20-to-erc20`, we cannot prevent token contract to execute a call with zero amount.
* In `ambErcToErc` we use ERC677 token, so it is possible to revert in `onTokenTransfer`. We don't check if amount is positive there, BUT we check that amount is within the limits. The `minPerTx` limit cannot be set to zero, that's why, I guess, we don't have an explicit statement `require(value > 0)`.

I made a manual test on deployed test STAKE bridge, it does not allow to transfer 0 tokens to the mediator. Do we need an additional explicit test for zero value, if we already have tests for values which are out of limits?

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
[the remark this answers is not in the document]
k1rill-fedoseev: No, we don't have a test for it, I will add it.

contracts/upgradeable_contracts/stake_token_mediators/ForeignStakeTokenMediator.sol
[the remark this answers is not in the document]
akolotov: > In regular ERC20 transfer for native-to-erc20, erc20-to-native and erc20-to-erc20, we cannot prevent token contract to execute a call with zero amount.

Hmm... Since for the case `native-to-erc20` we are using the ERC677 token contract, the `transfer` method will eventually executes `onTokenTransfer` that will check whether the value is within the limits.

But `erc20-to-native` and `erc20-to-erc20` do not have such check, you are right. For them, the following check https://github.com/poanetwork/tokenbridge-contracts/blob/5f57e8ff8ee887c9fad0518c7aa58aaf91cde88e/contracts/upgradeable_contracts/BasicHomeBridge.sol#L46 will make sure that `0` will not be propagated further (e.g. to the block reward contract).

> Do we need an additional explicit test for zero value

Yes, it makes sense.

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * @dev Internal function for transfer of tokens, with optional minting if current balance is insufficient
+     * @param _recipient address of tokens receiver
+     * @param _value amount of fixed tokens
+     */
+    function _transferWithOptionalMint(address _recipient, uint256 _value) internal {
+        IBurnableMintableERC677Token token = IBurnableMintableERC677Token(erc677token());
+        uint256 balance = token.balanceOf(address(this));
+        if (_recipient != address(0) && balance == 0) {
+            token.mint(_recipient, _value);
```
varasev, line 52: This line assumes that the `_value` will be minted directly to `_recipient` and emit `Transfer(address(0), _recipient, _value)` event. Is it OK that the `Transfer` event in this case contains the `from = address(0)` field? Will the event be caught by the bridge's oracle?

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,60 @@
...
+        } else if (balance < _value) {
+            token.mint(address(this), _value - balance);
+            token.transfer(_recipient, _value);
```
varasev, line 55: Please note that the `transfer` method doesn't allow the `_recipient` to be `address(0)` as well as the `mint` (see the deployed token contract on Mainnet: https://etherscan.io/address/0x0ae055097c6d159879521c384f1d2123d1f195e6#code). I guess the bridge should restrict setting the `recipient` to zero on the `Home` side (in case of using the `alternative receiver` feature). If some user wants to burn their tokens through the bridge (sending them from Home to Foreign and setting the recipient as zero), they can use another non-zero address (like [0x00...01](https://etherscan.io/address/0x0000000000000000000000000000000000000001)).

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+            token.transfer(_recipient, _value);
```
varasev, line 57: The same note regarding zero recipient (see above).

contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenMediator.sol
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * @dev Executes action on relayed request to fix the failed transfer of tokens
+     * @param _recipient address of tokens receiver
+     * @param _value amount of fixed tokens
+     */
+    function executeActionOnFixedTokens(address _recipient, uint256 _value) internal {
+        getMintHandler().mint(_recipient, _value);
+    }
```
varasev, line 159: I have a question regarding `executeActionOnFixedTokens` function: as far as I understand, it's used to return tokens back to sender if something is wrong when passing a message from bridge contract to mediator contract. I tried to analyze the following case (please correct me if I'm wrong somewhere):

For example, some sender Bob on Home side uses `alternative receiver` feature to send tokens to Alice. He sends their tokens to Home mediator contract, so `bridgeSpecificActionsOnTokenTransfer` is triggered (and the tokens are burnt on Home side) and then the message passed to Home bridge contract. Oracle catches the corresponding event, the bridge validators sign the message and request the Foreign bridge contract to pass the message to Foreign mediator contract, but `_passMessage` returns `false` for some reason.

Then someone calls the `requestFailedMessageFix` on the Foreign mediator contract. The Foreign mediator contract in turn through the Foreign bridge contract asks for `fixFailedMessage` to be called on Home side. The `fixFailedMessage` calls `executeActionOnFixedTokens` of the Home mediator contract which mints back the tokens for Alice (but not for Bob) on the Home side.

Is it intended behavior that in case of `alternative receiver` feature the failed tokens are returned back to Alice, but not to Bob on Home side?

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
k1rill-fedoseev: This restriction is already included in the `amb-erc-to-erc`
https://github.com/poanetwork/tokenbridge-contracts/blob/393ef6c10bb59d1d370b4de9df77c23c29ac7f0f/contracts/upgradeable_contracts/BaseERC677Bridge.sol#L29-L37
So, calls like `token.transferAndCall(mediator, value, 0x00..00)` will be reverted

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
varasev: Ok, got it.

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
varasev: Ok 👍

varasev: Then maybe `_recipient != address(0)` condition is redundant here? If the recipient can never be zero due to that condition in `chooseReceiver`.

k1rill-fedoseev: Yes, you are right. It is redundant here

varasev:
Should `onTokenTransfer` on the bridge contract `revert` when someone sends tokens directly to bridge contract instead of sending them to mediator contract?

contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenMediator.sol
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * @dev Updates address of contract used for handling mint operations,
+     * all subsequent mint operations will be called through this contract
+     * @param _mintHandler address of new contract
+     */
+    function setMintHandler(address _mintHandler) external onlyOwner {
+        require(AddressUtils.isContract(_mintHandler));
+        addressStorage[MINT_HANDLER] = _mintHandler;
+    }
```
varasev, line 105: Please clarify, why do we need to have a separate mint handler? In which cases will it differ from `erc677token()`?

k1rill-fedoseev: In the deployed STAKE token contract for the home chain, only owner is allowed to mint tokens.

For one bridge, we simply call `transferOwnership(mediator)` on token. Mediator uses `erc677token()` for minting.

In case of several bridges for the same token, all of them should be able to mint tokens. For that, we introduce a special contract that will be the owner of the token. This contract is also set to be a `mintHandler` in all existing bridges. This contract will be a simple proxy between bridges and a token for mint operations. A draft example of such contract can be found here: https://github.com/poanetwork/tokenbridge-contracts/blob/develop/contracts/mocks/MintHandlerMock.sol
It is used in tests for mint handler functionality.

varasev: Got it, thanks.

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignStakeTokenMediator.sol
varasev: The condition is about to be removed in https://github.com/poanetwork/tokenbridge-contracts/pull/408

akolotov:
> Should `onTokenTransfer` on the bridge contract `revert` when someone sends tokens directly to bridge contract instead of sending them to mediator contract?

If I understood your question correctly, you consider the scenario when tokens (not only STAKE tokens) are being sent to a AMB bridge contract. But the AMB token contracts have no `onTokenTransfer` functionality at all. It is assumed that if some one occasionally sent the tokens to a wrong address the only thing they could do is to ask the bridge owners to use `claimTokens` functionality: https://github.com/poanetwork/tokenbridge-contracts/blob/fdaab2777efe2f48c20a690c9f59196fa89e3631/contracts/upgradeable_contracts/BasicBridge.sol#L38

varasev:
I mean `onTokenTransfer` in `BasicAMBErc677ToErc677`: seems this contract is inherited by `HomeAMBErc677ToErc677` and `ForeignAMBErc677ToErc677` contracts. As far as I understand, these contracts are AMB bridge contracts and they can accept STAKE tokens. Is it OK, or I missed something?

My understanding is that if we use mediator contract, the mediator contract should accept tokens, but the bridge contract shouldn't (which is between mediator contract and oracle). Am I correct?

akolotov:
> My understanding is that if we use mediator contract, the mediator contract should accept tokens, but the bridge contract shouldn't (which is between mediator contract and oracle). Am I correct?

The token contract communicates with the mediator, the mediator contract communicates with the bridge contract. The oracle listens events emitted by the bridge contract.

varasev:
OK, so in AMB scheme, what will happen if I send STAKE token to the bridge contract (not to the mediator contract)?

Does AMB allow sending STAKE tokens to bridge contract? (not mediator one)

akolotov:
> OK, so in AMB scheme, what will happen if I send STAKE token to the bridge contract (not to the mediator contract)?
> 
> Does AMB allow sending STAKE tokens to bridge contract? (not mediator one)

When we are saying the AMB we mean the bridge contract. The AMB contracts have no `onTokenTransfer` functionality at all. It is assumed that if some one occasionally sent the any tokens to the bridge address the only thing they could do is to ask the bridge owners to use `claimTokens` functionality.

PR #400, opened 2020-04-08
Fix ChaiConnector behaviour when DSR is zero

### LINKED ISSUE #395: signature execution fails due to rounding error in Chai contract
After the migration to the Chai token for the xDai bridge it was found that there is [an issue in the Chai token contract](https://github.com/dapphub/chai/issues/12). This issues causes the signature execution transaction to fail: https://etherscan.io/tx/0x236ea5de2c8957c9706f78f2f0347feb9b5b2757a59c285bfcda9ff008aeb178

The revert happens on the following line:
https://github.com/poanetwork/tokenbridge-contracts/blob/5f57e8ff8ee887c9fad0518c7aa58aaf91cde88e/contracts/upgradeable_contracts/ChaiConnector.sol#L305

More details are here: 
  * https://dashboard.tenderly.dev/tx/main/0x236ea5de2c8957c9706f78f2f0347feb9b5b2757a59c285bfcda9ff008aeb178
  * http://ethtx.info/mainnet/0x236ea5de2c8957c9706f78f2f0347feb9b5b2757a59c285bfcda9ff008aeb178

Here is calculations exposing the rounding error:
image

### LINKED ISSUE #396: revert convertDaiToChai() if no need to swap to Chai
Currently it is possible the situation when someone calls `convertDaiToChai()` when the balance of DAI tokens is less than `minDaiTokenBalance()`. It will cause invocation of `approve` and `join` which will change nothing but consume the gas. 

https://github.com/poanetwork/tokenbridge-contracts/blob/5f57e8ff8ee887c9fad0518c7aa58aaf91cde88e/contracts/upgradeable_contracts/ChaiConnector.sol#L263-L267

Consider to revert if `daiBalance().sub(minDaiTokenBalance())` is zero.

k1rill-fedoseev:
Closes #395 
Closes #396

### DIFF
--- a/contracts/upgradeable_contracts/ChaiConnector.sol
+++ b/contracts/upgradeable_contracts/ChaiConnector.sol
@@ -185,7 +185,7 @@ contract ChaiConnector is Ownable, ERC20Bridge, TokenSwapper {
 
         receiver.call(abi.encodeWithSelector(ON_TOKEN_TRANSFER, address(this), interest, ""));
 
-        // Additional constant is not needed here, since we allow to withdraw only extra part of chai balance,
+        // Additional constant to tolerate the DAI balance deposited to the Chai token is not needed here, since we allow to withdraw only extra part of chai balance,
         // which is not necessary to cover 100% dai balance.
         // It is guaranteed that the withdrawal of interest won't left the bridge balance uncovered.
         require(dsrBalance() >= investedAmountInDai());
--- a/contracts/upgradeable_contracts/ChaiConnector.sol
+++ b/contracts/upgradeable_contracts/ChaiConnector.sol
@@ -185,7 +185,8 @@ contract ChaiConnector is Ownable, ERC20Bridge, TokenSwapper {
 
         receiver.call(abi.encodeWithSelector(ON_TOKEN_TRANSFER, address(this), interest, ""));
 
-        require(dsrBalance() >= investedAmountInDai());
+        // see comment in convertDaiToChai() for similar statement
+        require(dsrBalance() + 10000 >= investedAmountInDai());
 
         emit PaidInterest(receiver, interest);
     }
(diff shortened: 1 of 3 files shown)


contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -185,6 +185,9 @@ contract ChaiConnector is Ownable, ERC20Bridge, TokenSwapper {
 
         receiver.call(abi.encodeWithSelector(ON_TOKEN_TRANSFER, address(this), interest, ""));
 
+        // Additional constant is not needed here, since we allow to withdraw only extra part of chai balance,
```
akolotov, line 188:
```suggestion
        // Additional constant to tolerate the DAI balance deposited to the Chai token is not needed here, since we allow to withdraw only extra part of chai balance,
```

contracts/upgradeable_contracts/ChaiConnector.sol
```diff
@@ -185,6 +185,9 @@ contract ChaiConnector is Ownable, ERC20Bridge, TokenSwapper {
...
+        // which is not necessary to cover 100% dai balance.
+        // It is guaranteed that the withdrawal of interest won't left the bridge balance uncovered.
```
akolotov, line 190: Do you have a unit test to cover possible scenarios to confirm this?

k1rill-fedoseev: Yes, it is covered by regular scenarios for `payInterest`. 

We are paying interest in Chai directly, without any conversion. So, we cannot lose anything there. When we are evaluating the amount of payed interest in Chai here: https://github.com/poanetwork/tokenbridge-contracts/blob/f45c0cf7aef7bb0cd2fada9be6ff9ecde71d9492/contracts/upgradeable_contracts/ChaiConnector.sol#L181 we leave the sufficient amount of chai on the bridge balance, which is enough to cover the whole xDai supply.

akolotov: > We are paying interest in Chai directly, without any conversion.
Yeah, OK. It makes sense.

Issue #373, opened 2020-02-21
Use uintStorage as a storage for bytes32


k1rill-fedoseev:
In several places, the `bytesStorage` is used for storing `bytes32` variables, since there is no separate storage for `bytes32` types in `EternalStorage` contract.

Consider to use `uintStorage` as a storage for `bytes32`, in order to save some gas.

In particular, there are two functions in `MessageProcessor`, which are relying on this, `setFailedMessageDataHash` and `setTransactionHash`.
There is a third usage of this in `setNonce` of AMB Mediator, it was addressed in 4a615de

k1rill-fedoseev:
There is nothing complicated with `setTransactionHash`, since `transactionHash` is a temporary variable, which do not persist between calls to bridge contract. So, this code part can be replaced with 
```solidity
function transactionHash() external view returns (bytes32) {
    return bytes32(uintStorage[TRANSACTION_HASH]);
}

function setTransactionHash(bytes32 _txHash) internal {
    uintStorage[TRANSACTION_HASH] = uint256(_txHash);
}
```
For `setFailedMessageDataHash` the situation is a little bit different, saved data is then being used again. I see two ways of dealing with it:
* Leave it as it is
* Support both ways of storing, something like:
```solidity
function failedMessageDataHash(bytes32 _txHash) external view returns (bytes32) {
    bytes32 key = keccak256(abi.encodePacked("failedMessageDataHash", _txHash));
    uint256 dataHash =  uintStorage[key];
    return dataHash > 0 ? bytes32(dataHash) : Bytes.bytesToBytes32(bytesStorage[key]);
}

function setFailedMessageDataHash(bytes32 _txHash, bytes data) internal {
    uintStorage[keccak256(abi.encodePacked("failedMessageDataHash", _txHash))] = uint256(keccak256(data));
}
```
For this case, I got the following numbers for execution costs:
* Old version with `bytesStorage`:
  * `setFailedMessageDataHash` is `42049` gas
  * `failedMessageDataHash` is `2521` gas
* New version with `uintStorage`:
  * `setFailedMessageDataHash` is `20740 ` gas
  * `failedMessageDataHash` is `1358` gas for reading new dataHash from `uintStorage`, `3418` gas for reading.

akolotov:
Addressed in [`5.0.0`](https://github.com/poanetwork/tokenbridge-contracts/releases/tag/5.0.0)

PR #403, opened 2020-04-14
Multi AMB requests in one transaction

### LINKED ISSUE #307: Several request to AMB in one transaction
At this particular moment bridge contracts does not allow to call `requireToPassMessage` several times in one transaction.

Here is explanation provided by @patitonar:
```
calling requireToPassMessage N times in one tx I think will cause an undesired
behaviour:
1. It will generate N events
2. Oracle will get N events (which has the same txHash) and process correctly
    all of them
3. On the contracts invocations by the oracles:
  - If data in the events are the same:
    Only one event will be correctly processed. Other will be rejected by
    submitSignature/executeAffirmation signature validation.
  - If data is different in the events:
    All events will be processed and execute the contracts calls, but stored
    information regarding callStatus, failedMessageDataHash, failedMessageReceiver,
    failedMessageSender will be overwritten as events are executed because they
    share the same txHash.
```
After internal discussion the following proposal was developed (originally posted [here](https://hackmd.io/@1oqceZGlR6WSWAv2TTWJeA/SyU-8SROS)):

_(it breaks current bridge logic)_

`requireToPassMessage` will generate a `messageId` and return it. `messageId` should be part of the event, probably as a separate field so it is easier for user to find it when inspecting the transaction/logs. If that is the case, the validators should include the messageId into the message replacing the transactionHash since it is no longer needed. We should also consider to include `messageId` as part of `eventData` directly. 
```solidity
    function requireToPassMessage(address _contract, bytes _data, uint256 _gas) public returns (bytes32) {
        require(_gas >= getMinimumGasUsage(_data) && _gas <= maxGasPerTx());
        bytes32 messageId = nonce();
        bytes eventData = abi.encodePacked(msg.sender, _contract, messageId, _gas, uint8(0x00), _data);
        setNonce(keccak256(abi.encodePacked(messageId, eventData)));
        emitEventOnMessageRequest(messageId, eventData);
        return messageId;
    }
```

Using messageId instead of transactionHash, every message would be unique, so message processor could correctly process and store related information for every message. It could look like this:
```solidity
    function processMessage(
        address sender,
        address executor,
        bytes32 messageId,
        uint256 gasLimit,
        bytes1, /* dataType */
        uint256, /* gasPrice */
        bytes memory data
    ) internal {
        bool status = _passMessage(sender, executor, data, gasLimit, messageId);

        setMessageCallStatus(messageId, status);
        if (!status) {
            setFailedMessageReceiver(messageId, executor);
            setFailedMessageSender(messageId, sender);
        }
        emitEventOnMessageProcessed(sender, executor, messageId, status);
    }

    function _passMessage(address _sender, address _contract, bytes _data, uint256 _gas, bytes32 _messageId)
        internal
        returns (bool)
    {
        setMessageSender(_sender);
        setMessageId(_messageId);
        bool status = _contract.call.gas(_gas)(_data);
        setMessageSender(address(0));
        setMessageId(bytes32(0));
        return status;
    }

```

#### Example of usage in AMBERC677ToERC677
```solidity
    function passMessage(address _from, uint256 _value) internal {
        bytes4 methodSelector = this.handleBridgedTokens.selector;
        bytes memory data = abi.encodeWithSelector(methodSelector, _from, _value);

        bytes32 messageId = bridgeContract().requireToPassMessage(mediatorContractOnOtherSide(), data, requestGasLimit());

        setMessageHashValue(messageId, _value);
        setMessageHashRecipient(messageId, _from);
    }
```
Nonce was removed because it is no longer needed, it uses messageId to store information related to the transfer in case the transaction fails in the other network.

To request to fix a failed transaction, instead of using transactionHash, now we use messageId:
```solidity
    function requestFailedMessageFix(bytes32 _messageId) external {
        require(!bridgeContract().messageCallStatus(_messageId));
        require(bridgeContract().failedMessageReceiver(_messageId) == address(this));
        require(bridgeContract().failedMessageSender(_messageId) == mediatorContractOnOtherSide());

        bytes4 methodSelector = this.fixFailedMessage.selector;
        bytes memory data = abi.encodeWithSelector(methodSelector, _messageId);
        bridgeContract().requireToPassMessage(mediatorContractOnOtherSide(), data, requestGasLimit());
    }
```

Method to fix the request:
```solidity
    function fixFailedMessage(bytes32 _messageId) external {
        require(msg.sender == address(bridgeContract()));
        require(messageSender() == mediatorContractOnOtherSide());
        require(!messageFixed(_messageId));

        address recipient = messageRecipient(_messageId);
        uint256 value = messageValue(_messageId);
        setMessageFixed(_messageId);
        executeActionOnFixedTokens(recipient, value);
        emit FailedMessageFixed(_messageId, recipient, value);
    }
```

k1rill-fedoseev:
Resolves #307


k1rill-fedoseev:
### Regarding 2.
How we should distinguish these two functions? We cannot use both:
* `function requireToPassMessage(address, bytes, uint256) external returns (bytes32)`
* `function requireToPassMessage(address, bytes, uint256) external`

, they have the same signature. 

The old version of mediators can successfully call new version of `requireToPassMessage` by simply ignoring its output. However, there are problems with two other functions: `transactionHash()` and `failedMessageDataHash(bytes32 _txHash)`. If we remove them, old mediators logic will definitely break. In case our plan is to support old versions of mediators too, I propose to do the following:
* `transactionHash()` can return the same value, as `messageId()` does. It won't be a real transaction hash anymore, but the mediator contract won't notice any difference. In case the mediator emits some events that include a returned value of `transactionHash()`, off-chain logic for handling such events should be updated.
* `failedMessageDataHash(bytes32)` can be kept, as it is implemented now. And the AMB contract should continue to save `dataHash` for each failed message, even it is not needed anymore for the new mediators.

If we do so, then the old version of mediator still will be able to do something like:
```solidity
function requestFailedMessageFix(bytes32 _txHash) external {
  require(!bridgeContract().messageCallStatus(_txHash));
  require(bridgeContract().failedMessageReceiver(_txHash) == address(this));
  require(bridgeContract().failedMessageSender(_txHash) == mediatorContractOnOtherSide());
  bytes32 dataHash = bridgeContract().failedMessageDataHash(_txHash);
  bytes4 methodSelector = this.fixFailedMessage.selector;
  bytes memory data = abi.encodeWithSelector(methodSelector, dataHash);
  bridgeContract().requireToPassMessage(mediatorContractOnOtherSide(), data, requestGasLimit());
}
```
And the new version mediator will be able to do something like:
```solidity
function requestFailedMessageFix(bytes32 messageId) external {
  require(!bridgeContract().messageCallStatus(messageId));
  require(bridgeContract().failedMessageReceiver(messageId) == address(this));
  require(bridgeContract().failedMessageSender(messageId) == mediatorContractOnOtherSide());
  bytes4 methodSelector = this.fixFailedMessage.selector;
  bytes memory data = abi.encodeWithSelector(methodSelector, messageId);
  bridgeContract().requireToPassMessage(mediatorContractOnOtherSide(), data, requestGasLimit());
}
```
### Regarding 3.
Replay attack won't be possible, since the message structure and processing logic didn't change. 
For foreign contract, `messageId` has the same format and position in the encoded message, as `txHash`, so if someone will try to relay old already handled transaction, contract will treat `txHash` from this message as a `messageId`, but the `require(!relayedMessages(messageId));` will revert the execution. 
For home contract the situation is similar, if some validator will try to `executeAffirmation` on old message, the execution will be reverted, since the provided `msgHash` was already marked as processed.

akolotov:
@k1rill-fedoseev 

I agree on the item 2. Thanks.

But for the item 3 we need to think more. Since the tx hash now is not the part of the message could it compromise security? In other words, if the message id is unique enough? E.g. it could be the same in different chains if the bridge contract was deployed from the same account with the same number of transaction (nonce). Is a similar property applicable for the tx hashes and they also could be non-unique?

contracts/interfaces/IAMB.sol
```diff
@@ -3,10 +3,9 @@ pragma solidity 0.4.24;
 interface IAMB {
     function messageSender() external view returns (address);
     function maxGasPerTx() external view returns (uint256);
-    function transactionHash() external view returns (bytes32);
+    function messageId() external view returns (bytes32);
     function messageCallStatus(bytes32 _txHash) external view returns (bool);
```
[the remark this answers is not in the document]
k1rill-fedoseev: Yes, it will be better. Updated argument names.

k1rill-fedoseev:
This PR will eventually close #373 also. The existing bridge does not have any failed messages, so the support of previously used approach is not needed.

contracts/libraries/Bytes.sol
```diff
@@ -5,18 +5,6 @@ pragma solidity 0.4.24;
  * @dev Helper methods to transform bytes to other solidity types.
  */
 library Bytes {
-    /**
```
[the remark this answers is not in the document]
k1rill-fedoseev: Kept this function. Added a comment about its argument length.

akolotov:
Let's prepare the changes for the oracle as well. We can trail the upgrade then on the bridge between Kovan and Sokol.

akolotov:
@patitonar please join to us for review

contracts/upgradeable_contracts/amb_native_to_erc20/ForeignAMBNativeToErc20.sol
```diff
@@ -196,15 +196,15 @@ contract ForeignAMBNativeToErc20 is BasicAMBNativeToErc20, ReentrancyGuard, Base
...
-    * @param _txHash the transaction hash associated that generated fee distribution
+    * @param _messageId id of the message that generated fee distribution
     */
-    function distributeFee(IMediatorFeeManager _feeManager, uint256 _fee, bytes32 _txHash) internal {
+    function distributeFee(IMediatorFeeManager _feeManager, uint256 _fee, bytes32 _messageId) internal {
         // Right now, AMB bridge supports only one message per transaction.
         // The receivers of the fee could try to send back the fees through the mediator,
         // so here we add a lock to limit the number of messages that the mediator can send to the bridge,
         // allowing a maximum of 1 message
         enableMessagesRestriction();
```
patitonar, line 206: Now that AMB supports multiple messages per transaction, I think this restriction `enableMessagesRestriction` is no longer needed, right @akolotov ?

contracts/libraries/ArbitraryMessage.sol
```diff
@@ -4,56 +4,63 @@ import "../interfaces/IBridgeValidators.sol";
...
-    // https://github.com/paritytech/parity-bridge/issues/61
+    /**
+    * @dev Unpacks data fields from AMB message
+    * layout of message :: bytes:
+    * offset  0: 32 bytes :: uint256 - message length
+    * offset 32: 32 bytes :: bytes32 messageId
+    * offset 64: 32 bytes :: uint256 chainId
+    * offset 96: 20 bytes :: address - sender address
+    * offset 116: 20 bytes :: address - executor contract
+    * offset 136: 4 bytes :: uint256 - gasLimit
```
patitonar, line 15:
```suggestion
    * offset 136: 4 bytes :: uint32 - gasLimit
```

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
```diff
@@ -47,46 +47,50 @@ contract MessageProcessor is EternalStorage {
         addressStorage[MESSAGE_SENDER] = _sender;
     }
 
+    function messageId() public view returns (bytes32) {
+        return bytes32(uintStorage[MESSAGE_ID]);
+    }
+
     function transactionHash() external view returns (bytes32) {
```
patitonar, line 54: Should we add a comment to mention that this method is kept for backward compatibility? The same question applies to `failedMessageDataHash` and `setFailedMessageDataHash`

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -2,22 +2,48 @@ pragma solidity 0.4.24;
...
+            messageId,
+            chainId,
+            msg.sender,
+            _contract,
+            uint32(_gas),
+            uint8(0x00),
+            _data
+        );
+
+        emitEventOnMessageRequest(messageId, eventData);
```
patitonar, line 36: It looks like the `messageId` parameter is present in the two parameters of the event, in the first dedicated field and as part of the encodedData. Is this the expected behavior? I guess the tradeoff of this is the extra gas expended on the event vs the extra work of the validators to include the `messageId` into the message when processing the events

contracts/upgradeable_contracts/arbitrary_message/MessageProcessor.sol
k1rill-fedoseev: Yes, it makes sense. I added a doc comments in 2b04c78

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
k1rill-fedoseev: In all events where `messageId` is present, it is an indexed field. So it can be used for searching of all events related for one particular message.

I don't know if it is useful at this moment, but I think it can be helpful in the future.

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -2,22 +2,48 @@ pragma solidity 0.4.24;
...
+    * @param _contract executor address on the other side
+    * @param _data calldata passed to the executor on the other side
+    * @param _gas gas limit used on the other network for executing a message
+    */
+    function requireToPassMessage(address _contract, bytes _data, uint256 _gas) public returns (bytes32) {
         require(_gas >= getMinimumGasUsage(_data) && _gas <= maxGasPerTx());
-        emitEventOnMessageRequest(abi.encodePacked(msg.sender, _contract, _gas, uint8(0x00), _data));
+
+        uint256 chainId = _chainId();
+        bytes20 bridgeId = bytes20(keccak256(abi.encodePacked(chainId, address(this))));
```
akolotov, line 21:
```suggestion
        bytes32 bridgeId = keccak256(abi.encodePacked(chainId, address(this))) & 0x00000000ffffffffffffffffffffffffffffffffffffffff0000000000000000;
        uint64 nonce = _nonce();
        _setNonce(nonce + 1);
        
        bytes32 messageId = bytes32(ENCODED_BRIDGE_VERSION) << 224 | bridgeId | bytes32(nonce);
```

akolotov: Seems that it makes equivalent things but consumes less gas

contracts/upgradeable_contracts/arbitrary_message/VersionableAMB.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity 0.4.24;
+
+import "../VersionableBridge.sol";
+
+contract VersionableAMB is VersionableBridge {
+    // bridge version as a single 4-bytes number, included into every outgoing relay request
+    uint32 internal constant ENCODED_BRIDGE_VERSION = 0x00040000;
```
akolotov, line 7: Seems that it is more logical to move this constant to `ArbitraryMessage.sol` since it reflects the version of how the message is packed. That's why consider to rename it to `MESSAGE_PACKING_VERSION` or something like this.

contracts/libraries/ArbitraryMessage.sol
```diff
@@ -4,56 +4,63 @@ import "../interfaces/IBridgeValidators.sol";
...
-        uint256 srcdataptr = 32 + 20 + 20 + 32 + 1;
+        // 32 (message id) + 32 (chain id) + 20 (sender)  + 20 (executor) + 4 (gasLimit) + 1 (dataType)
+        uint256 srcdataptr = 32 + 32 + 20 + 20 + 4 + 1;
         assembly {
-            txHash := mload(add(_data, 32))
-            sender := mload(add(_data, 52))
-            executor := mload(add(_data, 72))
-            gasLimit := mload(add(_data, 104))
-            dataType := and(mload(add(_data, 136)), 0xFF00000000000000000000000000000000000000000000000000000000000000)
+            messageId := mload(add(_data, 32))
```
akolotov, line 52: Does it make sense to check that first 4 bytes of the message id match with `ENCODED_BRIDGE_VERSION`? If not, it seems that it is not the message we need to handle.

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
```diff
@@ -38,16 +38,24 @@ contract BasicHomeAMB is BasicAMB, MessageDelivery, MessageProcessor {
...
         address executor;
-        bytes32 txHash;
-        uint256 gasLimit;
+        uint32 gasLimit;
         bytes1 dataType;
         uint256 gasPrice;
         bytes memory data;
-        (sender, executor, txHash, gasLimit, dataType, gasPrice, data) = ArbitraryMessage.unpackData(_message, false);
-        processMessage(sender, executor, txHash, gasLimit, dataType, gasPrice, data);
+        (messageId, chainId, sender, executor, gasLimit, dataType, gasPrice, data) = ArbitraryMessage.unpackData(
```
akolotov, line 54: It seems that it is a good place to check that the chain Id of the contract matches with the chain Id in the message.

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
```diff
@@ -16,17 +16,21 @@ contract BasicForeignAMB is BasicAMB, MessageRelay, MessageDelivery, MessageProc
...
+        uint32 gasLimit;
         bytes1 dataType;
         uint256 gasPrice;
         bytes memory data;
 
-        (sender, executor, txHash, gasLimit, dataType, gasPrice, data) = ArbitraryMessage.unpackData(_data, true);
-        require(!relayedMessages(txHash));
-        setRelayedMessages(txHash, true);
-        processMessage(sender, executor, txHash, gasLimit, dataType, gasPrice, data);
+        (messageId, chainId, sender, executor, gasLimit, dataType, gasPrice, data) = ArbitraryMessage.unpackData(
```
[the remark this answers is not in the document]
k1rill-fedoseev: Seems that the bridge only knows the chain id of its own chain. So we cannot check that the chain id of the incoming message is correct.

From security point of view, remembering other side's chain id does not bring any improvement, that's why I don't think that we need such check.

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
k1rill-fedoseev: Nice improvement. Added in a805643. I have also changed the constant type to bytes32, so that we don't need to perform bit shift every time.

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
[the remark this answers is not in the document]
akolotov: ok

contracts/libraries/ArbitraryMessage.sol
k1rill-fedoseev: Added such checks in `BasicHomeAMB.sol` and `ForeignHomeAMB.sol` in e35136b

akolotov: Do you think it makes sense to introduce a separate method for this to simplify the code?

contracts/upgradeable_contracts/arbitrary_message/VersionableAMB.sol
k1rill-fedoseev: I think it makes sense to keep all versions-related stuff in one place, it will also help us to not forget to change this constant when bridge version is changed.

Another point is related to libraries in solidity, seems that it is not possible to access constants from library. I think that introducing a public getter is a big overhead in our case, so I propose to leave it in `VersionableAMB.sol`.

akolotov: Got you. Please add in the library that if the format of the message is changed, it is necessary to update the message packing version as well.

k1rill-fedoseev: Ok, added a note

contracts/upgradeable_contracts/arbitrary_message/BasicForeignAMB.sol
```diff
@@ -29,9 +29,7 @@ contract BasicForeignAMB is BasicAMB, MessageRelay, MessageDelivery, MessageProc
             _data,
             true
         );
-        require(
-            messageId & 0xffffffff00000000000000000000000000000000000000000000000000000000 == MESSAGE_PACKING_VERSION
-        );
+        _validateMessageId(messageId);
```
akolotov, line 32:
```suggestion
        require(_isMessageVersionValid(messageId));
```

akolotov: it is not good to hide `require` within the method.

k1rill-fedoseev: Got it

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -2,22 +2,56 @@ pragma solidity 0.4.24;
...
-        emitEventOnMessageRequest(abi.encodePacked(msg.sender, _contract, _gas, uint8(0x00), _data));
+
+        bytes memory sourceChainId = _sourceChainId();
+        bytes memory destinationChainId = _destinationChainId();
+
+        // 4 bytes - message version
+        // 20 bytes - bridge id
+        // 8 bytes - message nonce
+        bytes32 messageId = MESSAGE_PACKING_VERSION |
+            (keccak256(abi.encodePacked(sourceChainId, address(this))) &
```
akolotov, line 27: Bearing other comments in mind the following refactoring could be suggested. 

Consider to introduce a new method `packHeader(address, uint256)`.
Here is quite optimized version of the method:
```solidity
function _packHeader(address _contract, uint256 _gas) view internal returns(bytes header) {
    uint256 sourceChainId = _sourceChainId();
    uint256 sourceChainIdLength = _sourceChainIdLenght();
    uint256 destinationChainId = _destinationChainId();
    uint256 destinationChainIdLength = _destinationChainIdLength();
    
    uint256 mVer = MESSAGE_PACKING_VERSION; //note: it must be 0x00400000, not shifted
    uint256 nonce = _nonce();
    
    bytes32 bridgeId = keccak256(abi.encodePacked(sourceChainId, address(this)));
    // 79 = 4 + 20 + 8 + 20 + 20 + 4 + 1 + 1 + 1
    header = new bytes(79 + sourceChainIdLength + destinationChainIdLength);

    assembly {
        let header_ptr := add(header, 32)
        mstore(header_ptr, mVer)
        mstore(add(header_ptr, 4), shl(32, bridgeId))
        mstore(add(header_ptr, 20), shl(192, nonce))
        mstore(add(header_ptr, 32), shl(96, caller))
        mstore(add(header_ptr, 52), shl(96, _contract))
        mstore(add(header_ptr, 72), shl(224, _gas))
        mstore8(add(header_ptr, 76), sourceChainIdLength)
        mstore8(add(header_ptr, 77), destinationChainIdLength)
        mstore8(add(header_ptr, 78), 0x0)
        mstore(add(header_ptr, 79), shl(sub(256, shl(3, sourceChainIdLength)), 
                                        sourceChainId)) // shl(3, sourceChainIdLength) == sourceChainIdLength * 8
        mstore(add(add(header_ptr, 79), sourceChainIdLength), shl(sub(256, shl(3, destinationChainIdLength)), 
                                                                  destinationChainId)) // shl(3, destinationChainIdLength) == destinationChainIdLength * 8
    }
}
```
The returned value of this method will be encoded with `_data` in order to get full message.

It will allow to keep chain ids and their lengths as integers (rather than set of bytes) that should simplify an incoming message checking and setting the values through the initializer and the setter.

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -17,6 +33,7 @@ contract BasicAMB is BasicBridge {
         require(_gasPrice > 0);
         require(_requiredBlockConfirmations > 0);
 
+        _setChainIds(_sourceChainId, _destinationChainId);
```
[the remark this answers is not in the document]
k1rill-fedoseev: I thought it is reasonable to store chain ids in `bytes` rather than in a pair of `uint256`, since for short bytes blobs only one storage slot will be occupied.
I compared gas costs for both approaches, seems that using `uintStorage` is still a bit better solution.

I updated the contracts with the proposed approach.

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
```diff
@@ -42,4 +59,88 @@ contract BasicAMB is BasicBridge {
...
+    function _isMessageVersionValid(bytes32 _messageId) internal returns (bool) {
+        return
+            _messageId & 0xffffffff00000000000000000000000000000000000000000000000000000000 == MESSAGE_PACKING_VERSION;
+    }
+
+    /**
+     * Internal function for validating destination chain id of the received message
+     * @param _chainId destination chain id of the received message
+     */
+    function _isDestinationChainIdValid(uint256 _chainId) internal returns (bool res) {
```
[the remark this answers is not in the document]
k1rill-fedoseev: Should be ok for now, since we store chain ids in `uintStorage` instead of `bytesStorage`.

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
k1rill-fedoseev: As far as I can see, we can also switch the order of header packing (from last field to first). This will allow as to get rid of unnecessary `shl` operations.

contracts/upgradeable_contracts/arbitrary_message/BasicAMB.sol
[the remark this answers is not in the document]
akolotov: Yeah, I was considering that a short blob could take one slot and that's why we will have just one `SLOAD` operation but was also optimistic that operations with bytes set packing/unpacking could take more resources :)

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -17,41 +17,67 @@ contract MessageDelivery is BasicAMB {
...
+            0x00000000ffffffffffffffffffffffffffffffffffffffff0000000000000000;
+        // 79 = 4 + 20 + 8 + 20 + 20 + 4 + 1 + 1 + 1
+        header = new bytes(79 + sourceChainIdLength + destinationChainIdLength);
+
+        assembly {
+            let ptr := add(header, mload(header)) // points to the last word of header
+            mstore(ptr, destinationChainId)
+            mstore(sub(ptr, destinationChainIdLength), sourceChainId)
+
+            // can be ommitted for dataType 0x00, since header is already initialized with zeros
```
akolotov, line 69: Are you sure? What we we use a memory released by a previous code within the same execution environment?

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -17,41 +17,67 @@ contract MessageDelivery is BasicAMB {
...
+        bytes32 mVer = MESSAGE_PACKING_VERSION;
+        uint256 nonce = _nonce();
+
+        bytes32 bridgeId = keccak256(abi.encodePacked(sourceChainId, address(this))) &
+            0x00000000ffffffffffffffffffffffffffffffffffffffff0000000000000000;
+        // 79 = 4 + 20 + 8 + 20 + 20 + 4 + 1 + 1 + 1
+        header = new bytes(79 + sourceChainIdLength + destinationChainIdLength);
+
+        assembly {
+            let ptr := add(header, mload(header)) // points to the last word of header
```
akolotov, line 65: Well, I think you need to add a comment before the assembly part saying that "in order to safe gas the header is packed from the end"

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -17,41 +17,67 @@ contract MessageDelivery is BasicAMB {
...
+    function _packHeader(address _contract, uint256 _gas) internal view returns (bytes memory header) {
+        uint256 sourceChainId = _sourceChainId();
+        uint256 sourceChainIdLength = _sourceChainIdLength();
+        uint256 destinationChainId = _destinationChainId();
+        uint256 destinationChainIdLength = _destinationChainIdLength();
+
+        bytes32 mVer = MESSAGE_PACKING_VERSION;
+        uint256 nonce = _nonce();
+
+        bytes32 bridgeId = keccak256(abi.encodePacked(sourceChainId, address(this))) &
```
akolotov, line 59: I would also add a comment here "it does not make sense to keep the bridge Id in a storage since the operation to read it will take 800 gas whereas the operations to get sha3 plus to combine the integer and the address into the byte blob will consume less"

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
akolotov: cool! Very clever!

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
akolotov, line 69:
```suggestion
            // can be ommitted for dataType 0x00, since header is already initialized with zeros after the previous mstore operation
```

k1rill-fedoseev: In case `chainId` is 32 bytes, I think that this zero will come from `header = new bytes(79 + sourceChainIdLength + destinationChainIdLength);` line. This bytes blob should be allocated from the free memory pointer at `0x40`, so it will be pre-filled with zeros.

Am I right?

akolotov: I am not sure that it is pre-filled with zeros. Did you look at the decompiled solidity code to confirm this? Because, in general, the solidity considerations will work even for the next code to allocate a new set of bytes in the memory:
```solidity
assembly {
    new_ptr := mload(0x40)
    mstore(0x40, add(ptr, and(add(add(sizeOfNewDataBlob, 0x20), 0x1f), not(0x1f))))
}
``` 
As you can see it does not assume that the free memory is zeroed.

k1rill-fedoseev: Ok, then I think it makes sense to uncomment this `mstore` operation, it should not consume a lot of gas.

PR #410, opened 2020-04-29
Add permit feature to token contract

varasev:
This PR adds `permit` feature to the token contract.

### DIFF
--- a/contracts/ERC677BridgeTokenRewardable.sol
+++ b/contracts/ERC677BridgeTokenRewardable.sol
@@ -6,9 +6,9 @@ contract ERC677BridgeTokenRewardable is ERC677MultiBridgeToken {
     address public blockRewardContract;
     address public stakingContract;
 
-    constructor(string _name, string _symbol, uint8 _decimals)
+    constructor(string _name, string _symbol, uint8 _decimals, uint256 _chainId)
         public
-        ERC677MultiBridgeToken(_name, _symbol, _decimals)
+        ERC677MultiBridgeToken(_name, _symbol, _decimals, _chainId)
     {
         // solhint-disable-previous-line no-empty-blocks
     }
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ flats
 .idea
 coverage
 *.sublime-*
+.0x-artifacts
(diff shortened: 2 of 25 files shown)


varasev:
Travis shows some issue with `ethUtil` package. I will take a look.

contracts/ERC677BridgeTokenRewardable.sol
```diff
@@ -6,9 +6,9 @@ contract ERC677BridgeTokenRewardable is ERC677MultiBridgeToken {
     address public blockRewardContract;
     address public stakingContract;
 
-    constructor(string _name, string _symbol, uint8 _decimals)
+    constructor(string _name, string _symbol, uint8 _decimals, uint256 _chainId)
```
akolotov, line 9: In order to reduce the delta for introduced changes do you think it makes sense to have only `ERC677BridgeTokenRewardable` token contract to be permittable? So, the hierarchy of the token contracts could be:
-- ERC677BridgeToken
----PermittableToken
------ERC677MultiBridgeToken
--------ERC677BridgeTokenRewardable

Otherwise the changes you suggest will require the corresponding modifications in the tokenbridge monorepo, unit and e2e tests there assume that the deployment of the ERC677BridgeToken token does not require to specify the chain id.

varasev: Sure, I will change the inheritance.

varasev:
@k1rill-fedoseev Could you please take a look at one of the unit tests related to DSR. Travis caught the following mismatch:

```
657 passing (28m)
  1 failing
  1) Contract: ForeignBridge_ERC20_to_Native
       chai token
         Zero DSR
           should allow to executeSignatures after pay interest:
      AssertionError: expected '7400000000000000004' to be less than '7400000000000000000'
      + expected - actual
      -7400000000000000004
      +7400000000000000000
      
      at Context.it (test/erc_to_native/foreign_bridge.test.js:2207:66)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

contracts/ERC677BridgeTokenRewardable.sol
varasev: This is done

PR #416, opened 2020-05-05
Hiding of fee distribution implementation

akolotov:
Logically the mediator should not be aware about the process how fees are distributed, this should lay on the fee manager shoulders. That's why the code to send fee to the block reward contract was moved to the fee manager.

### DIFF
--- a/contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenFeeManager.sol
+++ b/contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenFeeManager.sol
@@ -64,6 +64,14 @@ contract HomeStakeTokenFeeManager is BlockRewardBridge, Ownable {
     }
 
     /**
+     * @dev Returns the state of the fee manager configuration: whether
+     * it is ready to collect and distribute fee or not.
+     */
+    function isFeeCollectingActivated() public view returns (bool) {
+        return ((address(_blockRewardContract()) != address(0)) && (getFee() > 0));
+    }
+
+    /**
      * @dev Distributes fee as per the logic of the fee manager.
      * In this particular case, the amount of fee is passed the block 
      * reward contract which will mint new tokens and distribute them
--- a/contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenFeeManager.sol
+++ b/contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenFeeManager.sol
@@ -45,7 +45,8 @@ contract HomeStakeTokenFeeManager is BlockRewardBridge, Ownable {
     }
 
     /**
-     * @dev Sets the fee percentage amount for the mediator operations. Only the owner can call this method.
+     * @dev Sets the fee percentage amount for the mediator operations.
+     * Only the owner can call this method.
      * @param _fee the fee percentage
      */
     function setFee(uint256 _fee) external onlyOwner {
@@ -60,5 +61,18 @@ contract HomeStakeTokenFeeManager is BlockRewardBridge, Ownable {
         require(_fee < MAX_FEE);
         uintStorage[FEE] = _fee;
         emit FeeUpdated(_fee);
+    }
+
+    /**
+     * @dev Distributes fee as per the logic of the fee manager.
+     * In this particular case, the amount of fee is passed the block 
+     * reward contract which will mint new tokens and distribute them
+     * among the stakers.
+     * @param _fee amount of tokens to be distributed
+     */
+    function _distributeFee(uint256 _fee) internal {
+        if (address(_blockRewardContract()) != address(0)) {
+            _blockRewardContract().addBridgeTokenRewardReceivers(fee);
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenFeeManager.sol
```diff
@@ -61,4 +62,25 @@ contract HomeStakeTokenFeeManager is BlockRewardBridge, Ownable {
...
+
+    /**
+     * @dev Distributes fee as per the logic of the fee manager.
+     * In this particular case, the amount of fee is passed the block 
+     * reward contract which will mint new tokens and distribute them
+     * among the stakers.
+     * @param _fee amount of tokens to be distributed
+     */
+    function _distributeFee(uint256 _fee) internal {
+        if (address(_blockRewardContract()) != address(0)) {
```
k1rill-fedoseev, line 82: Seems that we are already sure, that BlockReward contract is defined, from `isFeeCollectingActivated()` condition: https://github.com/poanetwork/tokenbridge-contracts/blob/0bc587436e6ad8ec196dfedbac7681b6da6d2aac/contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenMediator.sol#L143-L151

Does it make sense to double check it here?

patitonar, line 82: In `bridgeSpecificActionsOnTokenTransfer`, before calling `_distributeFee`, we call `isFeeCollectingActivated` that makes sure that block reward contract address is set. Should we also keep this other block reward check in case this method is called from different method?

patitonar: I just saw the same question was asked in another comment. I'm resolving this to continue the discussion there

akolotov: @patitonar @k1rill-fedoseev 
Thanks for the question. From my perspective we should leave check that `_blockRewardContract ` is not zero before calling a method from this contract in `_distributeFee`. Just in case if someone will call `_distributeFee` from another place.

But I see that we call `_blockRewardContract` twice and this is not optimal.

What I can do here is to add a comment in the comment before `_distributeFee` that the caller must be sure that the block reward contract is initialized and to remove if-statement with the safety check from this method. What do you think?

k1rill-fedoseev: Yeah, I think it will be fine. Let's add a comment

contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenMediator.sol
```diff
@@ -141,17 +140,16 @@ contract HomeStakeTokenMediator is BasicStakeTokenMediator, HomeStakeTokenFeeMan
...
-                // when block reward contract is defined, the calculated fee is subtracted from the original value
+            if (isFeeCollectingActivated()) {
                 uint256 fee = calculateFee(_value);
+                // the calculated fee is subtracted from the original value
                 passMessage(_from, chooseReceiver(_from, _data), _value.sub(fee));
                 if (fee > 0) {
-                    // the fee itself is distributed later in the block reward contract
-                    _blockRewardContract().addBridgeTokenRewardReceivers(fee);
+                    // the fee manager will take care about fee distribution
+                    _distributeFee(fee);
```
patitonar, line 149: Do you think it makes sense to add events when fees are distributed from the mediators? In other bridges contracts we have:
https://github.com/poanetwork/tokenbridge-contracts/blob/165d828437826a43166ca787f8d39fb81f00c398/contracts/upgradeable_contracts/RewardableBridge.sol#L8-L9

They are useful to calculate the amount of fee distributed to display that information on the statistic page of the Bridge UI.
Although, for the specific case of STAKE token the event `BridgeTokenRewardAdded` from the block reward can be used for that purpose
 https://github.com/poanetwork/posdao-contracts/blob/c14fddb6bc9863f5d2c2075df3d755eb7637c712/contracts/base/BlockRewardAuRaTokens.sol#L74

akolotov: Yeah, I saw this event that's why decided to not include another one here. Will it complicate the logic of TokenBridge UI?

patitonar: No, it's OK to use that event in the Bridge UI. I just wanted to check that we considered the option and that using `BridgeTokenRewardAdded` is the correct way to go

contracts/upgradeable_contracts/amb_erc677_to_erc677/HomeStakeTokenFeeManager.sol
akolotov: Addressed in 20b3059

Issue #424, opened 2020-05-25
Remove restriction for multiple calls within fee distribution process for native-to-erc677 extension


akolotov:
Before the AMB bridge started supporting multiple messages within one transaction the following restriction was introduces in the native-to-ERC677 AMB extension to get rid of the bridge imbalance:

https://github.com/poanetwork/tokenbridge-contracts/blob/129f4b4a4e94908d411845928ff30e640a8e3e66/contracts/upgradeable_contracts/amb_native_to_erc20/ForeignAMBNativeToErc20.sol#L206-L209

It seems that after #403 this restriction is not actual any more so consider to remove it in order to safe gas.

PR #425, opened 2020-06-03
Fix infinite AMB message loop

k1rill-fedoseev:
It was discovered, that it is not safe to allow passing messages while other message is currently processed. Specially formatted message can trigger an infinite loop, and eventually, all validators will burn their balance on gas expenses.

In order to fix this, the following restriction is introduced:
https://github.com/poanetwork/tokenbridge-contracts/blob/69c8f0bc94ff013d9317808076734a7a4feaf69a/contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol#L19-L20

This check will also make it impossible to bridge received fee in the same transaction, although it was possible to do so in the `AMB_NATIVE_TO_ERC` mode before.

### DIFF
--- a/test/amb_native_to_erc20/foreign_mediator.test.js
+++ b/test/amb_native_to_erc20/foreign_mediator.test.js
@@ -1541,6 +1541,7 @@ contract('ForeignAMBNativeToErc20', async accounts => {
 
       const initialBalanceRewardAddress1 = await token.balanceOf(rewardAccountList[0])
       const initialBalanceRewardAddress2 = await token.balanceOf(rewardAccountList[1])
+      const initialBalanceRewardAddress3 = await token.balanceOf(rewardAccountList[2])
 
       await ambBridgeContract.executeMessageCall(
         contract.address,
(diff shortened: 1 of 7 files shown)


test/amb_native_to_erc20/foreign_mediator.test.js
```diff
@@ -1561,17 +1562,15 @@ contract('ForeignAMBNativeToErc20', async accounts => {
       const erc20Token = await ERC20Mock.at(token.address)
       const transferEvents = await getEvents(erc20Token, { event: 'Transfer' })
       // 5 transfer events: 1 Mint to user, 1 Mint to fee manager, fee manager 1 transfer to each reward account,
-      // 1 feeReceiver contract transfer to mediator, 1 transfer to 0 to burn the tokens
-      expect(transferEvents.length).to.be.equal(7)
+      // bridge should not allow to send messages in backward direction, so fee manager should not be able to send tokens through the bridge
```
akolotov, line 1565: Could you clarify how this test will be changed after https://github.com/poanetwork/tokenbridge-contracts/issues/424 addressing?

k1rill-fedoseev: I think it should be left as it is now. This restriction for multiple calls allowed at most 1 message per transaction. As for now, passing new messages while processing other message is not allowed at all. So, removing that check in #424 should not change the contract behaviour.

akolotov: ok. thanks

PR #440, opened 2020-06-19
Add erc-to-native mediators

k1rill-fedoseev:


contracts/upgradeable_contracts/amb_erc20_to_native/BlockReward.sol
```diff
@@ -0,0 +1,178 @@
...
+        return false;
+    }
+
+    function _setExtraReceiverAmount(uint256 _amount, address _receiver) private {
+        uintStorage[keccak256(abi.encodePacked(EXTRA_RECEIVER_AMOUNT, _receiver))] = _amount;
+    }
+
+    function _setMinted(uint256 _amount, address _account) private {
+        bytes32 hash = keccak256(abi.encodePacked(MINTED_FOR_ACCOUNT_IN_BLOCK, _account, block.number));
+        uintStorage[keccak256(abi.encodePacked(MINTED_FOR_ACCOUNT_IN_BLOCK, _account, block.number))] = _amount;
```
akolotov, line 167: can we use `hash` calculated above?

k1rill-fedoseev: Yes, of course. That was simply a copy-paste mistake.
Fixed it.

contracts/upgradeable_contracts/amb_erc20_to_native/HomeFeeManagerAMBErc20ToNative.sol
```diff
@@ -0,0 +1,52 @@
...
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    /**
+    * @dev Transfer the fee as native tokens to the reward account.
+    * @param _rewardAddress address that will receive the native tokens.
+    * @param _fee amount of native tokens to be distribute.
+    */
+    function onFeeDistribution(address _rewardAddress, uint256 _fee) internal {
+        if (isOppositeDirection) {
```
akolotov, line 45: What was the reason to pass this parameter through the storage?

contracts/upgradeable_contracts/amb_erc20_to_native/HomeFeeManagerAMBErc20ToNative.sol
```diff
@@ -0,0 +1,52 @@
...
+            IBlockReward(blockReward).addExtraReceiver(_fee, _rewardAddress);
```
akolotov, line 46: Well, it is not logically correct to consider the fee manager as the the bridge contract within the block reward contract. Moreover, it caused the introduction of the total supply within the mediator contract. Consider the scheme we used in the vanilla bridge when the methods of the fee manager were invoked by `delegatecall`. It allows to not introduce new entity in the block reward contract.

contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
```diff
@@ -126,6 +150,17 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge {
         uint256 valueToMint = _value.mul(10**decimalShift());
         bytes32 _messageId = messageId();
 
+        IMediatorFeeManagerBothDirections feeManager = IMediatorFeeManagerBothDirections(feeManagerContract());
+        _setTotalSupply(totalSupply().add(valueToMint));
+        if (feeManager != address(0)) {
+            uint256 fee = feeManager.calculateOppositeFee(valueToMint);
+            if (fee != 0) {
+                feeManager.call(abi.encodeWithSelector(feeManager.distributeOppositeFee.selector, fee));
```
akolotov, line 157: Why we cannot call feeManager.distributeOppositeFee directly?

contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
```diff
@@ -1,15 +1,17 @@
...
 
 /**
 * @title HomeAMBErc20ToNative
 * @dev Home mediator implementation for erc20-to-native bridge intended to work on top of AMB bridge.
 * It is design to be used as implementation contract of EternalStorageProxy contract.
 */
-contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge {
-    bytes32 internal constant TOTAL_BURNT_COINS = 0x17f187b2e5d1f8770602b32c1159b85c9600859277fae1eaa9982e9bcf63384c; // keccak256(abi.encodePacked("totalBurntCoins"))
+contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, RewardableMediator {
+    bytes32 internal constant TOTAL_SUPPLY_COINS = 0xb32740a8fb91550d50618ece1336a41a472ba5977838b1cd512bba896c895fc6; // keccak256(abi.encodePacked("totalSupplyCoins"))
```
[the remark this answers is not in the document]
k1rill-fedoseev: Ok, I got it. I updated the logic, so that now fee manager is inherited by the primary mediator contract, so we can use `TOTAL_BURNT_COINS` as it was used before.

contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
```diff
@@ -34,9 +33,8 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, Rewar
         uint256 _requestGasLimit,
         uint256 _decimalShift,
         address _owner,
-        address _feeManager,
         address _blockReward
-    ) external onlyRelevantSender returns (bool) {
+    ) public onlyRelevantSender returns (bool) {
```
akolotov, line 37: Should it be public or internal?

k1rill-fedoseev: I think it should be public, so that it is possible to initialize a mediator by both: `initialize` and `rewardableInitialize`

contracts/upgradeable_contracts/amb_erc20_to_native/ForeignAMBErc20ToNative.sol
```diff
@@ -0,0 +1,174 @@
...
+    /**
+    * @dev This method is called when transferAndCall is used from ERC677 to transfer the tokens to this contract.
+    * It will initiate the bridge operation that will lock the amount of tokens transferred and mint the native tokens on
+    * the other network.
+    * @param _from address that transferred the tokens.
+    * @param _value amount of tokens transferred.
+    * @param _data this parameter could contain the address of an alternative receiver of the tokens on the other network,
+    * otherwise it will be empty.
+    */
+    function onTokenTransfer(address _from, uint256 _value, bytes _data) external returns (bool) {
```
akolotov, line 127: Thought that probably it is OK that our mediators support ERC677 but then noticed that public availability of `erc677token` method could confuse users. Consider to introduce a new method `erc20token` that will return the value of `erc677token` (if it is possible to hide `erc677token` in this case at all?).

image

k1rill-fedoseev: Agree. I hided the `erc677token()` method in erc-to-native and native-to-erc mediators, and introduced another one: `erc20token()`. Please, take a look.

akolotov: It is not correct to hide it for the `native-to-erc` mediators since they definitely work with a ERC677 token.

PR #457, opened 2020-07-28
Remove half duplex erc20 token support

### LINKED ISSUE #405: Remove code related to half duplex erc20 token support
After the changes introduced in https://github.com/poanetwork/tokenbridge-contracts/pull/404 the code to support a half duplex erc20 token is no longer needed.

patitonar:
Closes #405


contracts/upgradeable_contracts/erc20_to_native/ForeignBridgeErcToNative.sol
```diff
@@ -100,75 +82,17 @@ contract ForeignBridgeErcToNative is BasicForeignBridge, ERC20Bridge, OtherSideB
...
-    ) public pure returns (bool) {
-        return false;
-    }
-
-    function halfDuplexErc20token() public pure returns (ERC20) {
-        return ERC20(0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359);
-    }
-
-    function setMinHDTokenBalance(uint256 _minBalance) external onlyOwner {
-        uintStorage[MIN_HDTOKEN_BALANCE] = _minBalance;
```
akolotov, line 129: Can you also suggest a method to clean up all storage data used by the previous version of contracts as so if we upgrade the bridge tokens contract in the future we can invoke this method to free storage slots?

patitonar: All related contract addresses (Sai, SaiTop, MigrationContract) were hardcoded in the code, so there is only a few storage data to clean. The following method could be called when upgrading the contract version:
```solidity  
function cleanHalfDuplexStorage(address _receiver) external {
    require(msg.sender == address(this));

    delete uintStorage[0x48649cf195feb695632309f41e61252b09f537943654bde13eb7bb1bca06964e]; // keccak256(abi.encodePacked("minHDTokenBalance"))
    delete boolStorage[0xbeb8b2ece34b32b36c9cc00744143b61b2c23f93adcc3ce78d38937229423051]; // keccak256(abi.encodePacked("lockedSaiFixed"))
}
```
Given that in this case the method only frees up storage I didn't include a flag to allow only one execution by the contract in order to make the returned gas worth. But if you think we should include it anyway here is another version:
```solidity  
function cleanHalfDuplexStorage(address _receiver) external {
    require(msg.sender == address(this));
    bytes32 HALF_DUPLEX_CLEANED = 0xae6828552b008f109b269417f4452957f57fdccaa12685e65c4dd6519defe0a5; // keccak256(abi.encodePacked("halfDuplexCleaned"))
    require(!boolStorage[HALF_DUPLEX_CLEANED]);
    boolStorage[HALF_DUPLEX_CLEANED] = true;

    delete uintStorage[0x48649cf195feb695632309f41e61252b09f537943654bde13eb7bb1bca06964e]; // keccak256(abi.encodePacked("minHDTokenBalance"))
    delete boolStorage[0xbeb8b2ece34b32b36c9cc00744143b61b2c23f93adcc3ce78d38937229423051]; // keccak256(abi.encodePacked("lockedSaiFixed"))
}
```

akolotov: Thanks a lot! It is enough so far.

PR #462, opened 2020-08-03
Move fee manager into home mediator

k1rill-fedoseev:
After an additional discussion, it was decided to move fee manager functionality into the `HomeMultiAMBErc20ToErc677` mediator. Updated contracts bytecode size:
* **Home**: `22.7 kb`
* **Foreign**: `16.3 kb`


contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/BasicMultiTokenBridge.sol
```diff
@@ -269,11 +269,11 @@ contract BasicMultiTokenBridge is EternalStorage, Ownable {
             if (_minPerTx == 0) {
```
akolotov, line 269: Consider to add a comment when it could happen

PR #463, opened 2020-08-03
Deinit token in fixFailedMessage

k1rill-fedoseev:
Added the ability for reinitialising tokens, when `fixFailedMessage` is called for the first send message. Later, these tokens can be initialized once again.


k1rill-fedoseev:
I have updated the `callflows.md` and bumped the version, please take a look.

PR #468, opened 2020-08-11
AMB ENS mirroring

k1rill-fedoseev:
This PR implements a way of mirroring existing ENS records to the other chain.
The following logic is suggested:
* Whenever someone wants to mirror some ENS record, he/she calls `bridgeENSNode(namehash)` on the foreign mediator.
* AMB bridge transfers the information about the ENS node owner, ttl and associated address obtained from the resolver.
* On the home side, ENS registry is implemented as a part of the mediator contract. There is also a public resolver available. There are no any registrars. The only way of registering a new node, is to mirror it using the foreign mediator contract.
* When node is first seen on the home side, some basic setup is done (owner, ttl, associated address are assigned to the node). Default public resolver is also assigned to this node.
* For subsequent mirror operations, only owner address is updated, ttl and associated address are left unchanged.
* Current owner on the ENS node on the home side can update resolver and ttl settings. However, he/she is not allowed to change the owner address or create/update any subnodes.
* Public resolver is available as a separate non-upgradeable contract. It is intended to be used in the same way, as it is used in the Ethereum Mainnet. Implementation for the resolver was taken from https://github.com/ensdomains/resolvers/blob/8b97b0383607897dc4fe1f514007b0a3a72e9dbd/contracts/PublicResolver.sol


k1rill-fedoseev:
Closed in favour of a separate repository: https://github.com/poanetwork/amb-mediators

PR #479, opened 2020-08-27
Extra check for balance increase

k1rill-fedoseev:
The proposed fix is to address the issue with tokens that does not revert for `transferFrom` and `transfer` calls and return `false` instead.
It is not possible to apply the approach suggested in https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0b489f4d79544ab9b870abe121077043feb971b8/contracts/token/ERC20/SafeERC20.sol#L64 due to significant rework required to switch to newer version of Solidity. Later it will be investigated a possibility to implement it in form of assembly.

### DIFF
--- a/contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
+++ b/contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
@@ -62,6 +62,12 @@ contract BasicAMBErc677ToErc677 is
         return mediatorContractOnOtherSide();
     }
 
+    /**
+    * @dev Initiates the bridge operation that will lock the amount of tokens transferred and mint the tokens on
+    * the other network. The user should first call Approve method of the ERC677 token.
+    * @param _receiver address that will receive the minted tokens on the other network.
+    * @param _value amount of tokens to be transferred to the other network.
+    */
     function relayTokens(address _receiver, uint256 _value) external {
         // This lock is to prevent calling passMessage twice if a ERC677 token is used.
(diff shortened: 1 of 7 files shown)


fleupold (commented):
Looks good to me.
One thing to consider is that the double SLOAD might be a bit gas-inefficient. Here is how uniswap deals with it (which I believe is pretty much following the ERC20 standard definition): https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Pair.sol#L45-L46

akolotov (approved):

contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
```diff
@@ -62,6 +62,12 @@ contract BasicAMBErc677ToErc677 is
         return mediatorContractOnOtherSide();
     }
 
+    /**
```
akolotov, line 65: If it becomes specific for the home side only, does it make sense to move it to the corresponding contract then?

k1rill-fedoseev: It is also used in the stake token mediators pair. If we move the method to the `HomeAMBErc677ToErc677` contract, then we will need to place its duplicate in the `BasicStakeTokenMediator` contract.

akolotov: Got you. Thanks!

akolotov (approved):

Issue #183, opened 2019-05-21
migration to Solidity 5.x


akolotov:
As part of migration to the new version of Solidity compiler it is necessary to make sure that we extended list of supported EVM version introduced under #180.

To be more precise it is necessary to:
- [ ] change the comments in `deploy/.env.example` by adding more versions (e.g. `spuriousDragon`, `byzantium`, `constantinople` and `petersburg` - exact number of versions can be found [here](https://solidity.readthedocs.io/en/v0.5.5/using-the-compiler.html#target-options))
- [ ] change the comments in `deploy/README.md` by stating that the default value is `petersburg`
- [ ] change `deploy/src/loadContracts.js` to use different directories depending on different versions provided in the configuration file
- [ ] change `deploy/src/loadEnv.js` to make possible to configure compilations for `constantinople` and `petersburg` EVM versions
- [ ] change `truffle-config.js`  to use different directories and different value of `evmVersion` depending on different versions provided in the configuration file
- [ ] change `package.json` (`script: compile`) to reflect necessity to build contracts for more versions.

PR #485, opened 2020-09-08
Migrate from npm to yarn

k1rill-fedoseev:
I noticed that `yarn` package manager works much faster than the default `npm`. Also, `yarn` is used in the monorepo, so I propose to use it in this repo as well.

While changes delta in this PR is insane, there are only a few things were done:
* Move `./contracts/*` to `./contracts/contracts/*`, update import paths in deploy scripts.
* Add `contracts`, `deploy`, `upgrade` workspaces to the root `package.json`
* Remove npm's `package-lock.json` files, generate `yarn.lock` by running `yarn install`. As a side effect, this caused a lot of dependency updates.
* Enable linting for `./deploy` and `./upgrade`, it was already configured previously, but was not run.
* Run `yarn lint:js:fix` and `yarn lint:sol:fix` to fix appeared lint errors.
* Replace usages of `npm` with `yarn` commands in docs, scripts, Dockerfiles, etc.


akolotov (commented):
Well... I would not recommend merge these changes till some issue that requires similar re-work is being addressed. Such issue could be #183.

akolotov:
Could you share ideas why previously the lint does not require such changes for the contracts?

akolotov:
why is it require to move contracts to `./contracts/contracts/` instead of `./contracts`?

k1rill-fedoseev:
> Could you share ideas why previously the lint does not require such changes for the contracts?

I think some linting rules were changed or added to the new version of solhint.

> why is it require to move contracts to `./contracts/contracts/` instead of `./contracts`?

That is the recommended way to use yarn workspaces, to keep each workspace code and dependencies in its own directory.

PR #503, opened 2020-09-21
Add missing zero address checks

k1rill-fedoseev:
Some of the checks like `require(_owner != address(0))` were missing in a couple of contracts. This PR adds them where needed.


akolotov (approved):
Approving with the following explanation:

Even if the check `owner != address(0)` is more logical to perform at the beginning of a method since eventually it will require less gas to consume during the execution, it is suitable for these contracts to have it later because it will safe some gas for deployment and because the owner initialization/changing functionality are not methods that going to be executed often.

PR #509, opened 2020-10-08
Fix overdraw management

### LINKED ISSUE #508: Double spend possible for transfers above the limit
The intent of the overdraw management feature is to allow the bridge admins to fix the variables indicating that someone attempted to send assets exceeding the daily limit or the max per tx limit. If it was a valid transfer made by user rather than some malicious action from the bridge oracle, this feature allows to return funds to the user by informing another side of the bridge to unlock assets (if we consider ERC20-to-native or ERC20-to-ERC20 bridge modes). At the same time, the bridge admins could increase the limit and the oracles could re-send the confirmation and it will lead to tokens minting as per the new limit. It means that the funds can be unlocked and minted by sequence of actions that will lead to doubling the tokens.

Consider to change the behavior of smart contracts as so it will forbid for some specific bridged request to mint the tokens if the funds were returned with referring to this request or forbid to return assets if the tokens were minted as part of successful execution of this request.

k1rill-fedoseev:
Closes #508

### DIFF
--- a/contracts/upgradeable_contracts/HomeOverdrawManagement.sol
+++ b/contracts/upgradeable_contracts/HomeOverdrawManagement.sol
@@ -3,31 +3,25 @@ pragma solidity 0.4.24;
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "./Upgradeable.sol";
 import "./RewardableBridge.sol";
-import "./BasicTokenBridge.sol";
+import "./BasicHomeBridge.sol";
 import "./BaseOverdrawManagement.sol";
 
-contract OverdrawManagement is BaseOverdrawManagement, RewardableBridge, Upgradeable, BasicTokenBridge {
+contract HomeOverdrawManagement is BaseOverdrawManagement, RewardableBridge, Upgradeable, BasicHomeBridge {
     using SafeMath for uint256;
 
-    event UserRequestForSignature(address recipient, uint256 value);
-
-    function fixAssetsAboveLimits(bytes32 txHash, bool unlockOnForeign, uint256 valueToUnlock)
+    function fixAssetsAboveLimits(bytes32 hashMsg, bool unlockOnForeign, uint256 valueToUnlock)
         external
         onlyIfUpgradeabilityOwner
     {
-        require(!fixedAssets(txHash));
+        uint256 signed = numAffirmationsSigned(hashMsg);
+        require(!isAlreadyProcessed(signed));
         require(valueToUnlock <= maxPerTx());
-        address recipient;
-        uint256 value;
-        (recipient, value) = txAboveLimits(txHash);
+        (address recipient, uint256 value) = txAboveLimits(hashMsg);
         require(recipient != address(0) && value > 0 && value >= valueToUnlock);
         setOutOfLimitAmount(outOfLimitAmount().sub(valueToUnlock));
         uint256 pendingValue = value.sub(valueToUnlock);
-        setTxAboveLimitsValue(pendingValue, txHash);
-        emit AssetAboveLimitsFixed(txHash, valueToUnlock, pendingValue);
-        if (pendingValue == 0) {
-            setFixedAssets(txHash);
-        }
+        setTxAboveLimitsValue(pendingValue, hashMsg);
+        emit AssetAboveLimitsFixed(hashMsg, valueToUnlock, pendingValue);
         if (unlockOnForeign) {
             address feeManager = feeManagerContract();
             uint256 eventValue = valueToUnlock;
--- a/contracts/upgradeable_contracts/HomeOverdrawManagement.sol
+++ b/contracts/upgradeable_contracts/HomeOverdrawManagement.sol
@@ -6,9 +6,19 @@ import "./RewardableBridge.sol";
 import "./BasicHomeBridge.sol";
 import "./BaseOverdrawManagement.sol";
 
+/**
+ * @title HomeOverdrawManagement
+ * @dev This contract implements functionality for recovering from out-of-limits executions in Home side vanilla bridges.
+ */
 contract HomeOverdrawManagement is BaseOverdrawManagement, RewardableBridge, Upgradeable, BasicHomeBridge {
     using SafeMath for uint256;
 
+    /**
+    * @dev Fixes locked tokens, that were out of execution limits during the call to executeAffirmation.
+    * @param hashMsg reference for bridge operation that was out of execution limits.
+    * @param unlockOnForeign true if fixed tokens should be unlocked to the other side of the bridge.
+    * @param valueToUnlock unlocked amount of tokens, should be less than maxPerTx() and saved txAboveLimitsValue.
+    */
     function fixAssetsAboveLimits(bytes32 hashMsg, bool unlockOnForeign, uint256 valueToUnlock)
         external
         onlyIfUpgradeabilityOwner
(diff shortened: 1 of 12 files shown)


contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
```diff
@@ -129,7 +129,6 @@ contract BasicAMBErc677ToErc677 is
         external
         onlyIfUpgradeabilityOwner
     {
-        require(!fixedAssets(messageId));
         require(valueToUnlock <= maxPerTx());
```
akolotov, line 132: It looks like the check that `valueToUnlock` is within the limit is required only if `unlockOnForeign` is true.

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
```diff
@@ -129,13 +134,32 @@ contract HomeBridgeErcToErc is
...
+        internal
+        returns (bool)
+    {
+        (address aboveLimitsRecipient, uint256 aboveLimitsValue) = txAboveLimits(_hashMsg);
+        if (aboveLimitsRecipient != address(0)) {
+            // revert if a given transaction hash was already processed by the call to fixAssetsAboveLimits
+            require(aboveLimitsValue == _value);
+            setTxAboveLimits(address(0), 0, _hashMsg);
+            setOutOfLimitAmount(outOfLimitAmount().sub(_value));
+        }
```
akolotov, line 156: This code is the same in this file and in `HomeBridgeErcToNative.sol`. Does it make sense to have a separate method for this in `HomeOverdrawManagement`?

contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
```diff
@@ -121,26 +126,43 @@ contract HomeBridgeNativeToErc is EternalStorage, BasicHomeBridge, RewardableHom
...
     }
 
-    function onExecuteAffirmation(address _recipient, uint256 _value, bytes32 txHash) internal returns (bool) {
+    /**
+     * @dev Internal callback to be called on successfull message execution.
+     * Should be called only after enough affirmations from the validators are already collected.
+     * @param _recipient address of the receiver where the new coins should be unlocked.
+     * @param _value amount of coins to unlock.
+     * @param _txHash reference transaction hash on the Foreign side of the bridge which cause this operation.
+     * @param _hashMsg unique identifier of the particular bridge operation.
```
akolotov, line 135: Please add that it is not used within this method but is required for unification with other bridge modes.

contracts/upgradeable_contracts/HomeOverdrawManagement.sol
```diff
@@ -0,0 +1,45 @@
...
+    * @param unlockOnForeign true if fixed tokens should be unlocked to the other side of the bridge.
+    * @param valueToUnlock unlocked amount of tokens, should be less than maxPerTx() and saved txAboveLimitsValue.
+    */
+    function fixAssetsAboveLimits(bytes32 hashMsg, bool unlockOnForeign, uint256 valueToUnlock)
+        external
+        onlyIfUpgradeabilityOwner
+    {
+        uint256 signed = numAffirmationsSigned(hashMsg);
+        require(!isAlreadyProcessed(signed));
+        require(valueToUnlock <= maxPerTx());
```
akolotov, line 28: seems that it makes sense only if `unlockOnForeign` is true

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
```diff
@@ -164,13 +188,19 @@ contract HomeBridgeErcToErc is
...
+     * @param _hashMsg unique identifier of the particular bridge operation.
+     */
+    function onFailedAffirmation(address _recipient, uint256 _value, bytes32 _txHash, bytes32 _hashMsg) internal {
+        (address recipient, uint256 value) = txAboveLimits(_hashMsg);
         require(recipient == address(0) && value == 0);
         setOutOfLimitAmount(outOfLimitAmount().add(_value));
-        setTxAboveLimits(_recipient, _value, _txHash);
-        emit AmountLimitExceeded(_recipient, _value, _txHash);
+        setTxAboveLimits(_recipient, _value, _hashMsg);
+        emit AmountLimitExceeded(_recipient, _value, _txHash, _hashMsg);
```
akolotov, line 204: Do we need the tx hash here at all? In which case it could be necessary? Will it possible to extract it from the transaction input data in such case?

contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
```diff
@@ -129,7 +129,6 @@ contract BasicAMBErc677ToErc677 is
         external
```
akolotov, line 129: Could you imagine a scenario when it is required to use this functionality for this type of the bridge? Do we have a similar functionality for the opposite direction?

akolotov (changes requested):

contracts/upgradeable_contracts/HomeOverdrawManagement.sol
k1rill-fedoseev: Thanks, makes sense

contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
k1rill-fedoseev: Since it is a `BasicAMBErc677ToErc677.sol` contract, the functionality will be available on both sides of the bridge. 
I guess it makes sense to change a misleading parameter name `unlockOnForeign` to `unlockOnOtherSide`.

contracts/upgradeable_contracts/amb_erc677_to_erc677/BasicAMBErc677ToErc677.sol
k1rill-fedoseev: Yes, makes sense

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
k1rill-fedoseev: Thanks. Added a `_clearAboveLimitsMarker` function.

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
k1rill-fedoseev: I thought it can be useful in a situation when we want to find out if a given transfer was marked as out of limits.
In current implementation, we can search logs by using `txHash` directly.
But if we remove the `txHash` out from this event, then we need to construct a message manually, then take its hash, and search through the logs with this particular message hash.

It is not a big problem, so it is possible to remove it and keep the only version of `AmountLimitExceeded` event.
What do you think?

contracts/upgradeable_contracts/native_to_erc20/HomeBridgeNativeToErc.sol
k1rill-fedoseev: Done.

contracts/upgradeable_contracts/erc20_to_erc20/HomeBridgeErcToErc.sol
akolotov: Ok, it makes sense to leave both the tx hash and the msg hash both in the event to allow easily to look up if an affirmation request was failed or not. Then consider to have messageId (for the event with 3 arguments) and txHash (for the event with 4 arguments) to be indexed in the event description in order to simplify the process.

Probably, it makes sense to provide different name for these events like AmountLimitExceededInAMBExt and AmountLimitExceeded, doesn't it?

Consider

akolotov (changes requested):

akolotov (approved):

PR #510, opened 2020-10-12
Add missing fix mediator balance

### LINKED ISSUE #504: fixMediatorBalance missed in erc20-to-erc677 AMB extensions
It was discovered that the method `fixMediatorBalance` is missed on the foreign side of the erc20-to-erc677 mediator for AMB. Since this extension is used for the Moons bridge, it is not possible to recover funds that were sent accidentally by the `transfer` method to the mediator directly.

k1rill-fedoseev:
Closes #504

### DIFF
--- a/contracts/upgradeable_contracts/BasicBridge.sol
+++ b/contracts/upgradeable_contracts/BasicBridge.sol
@@ -50,7 +50,12 @@ contract BasicBridge is
         return uintStorage[REQUIRED_BLOCK_CONFIRMATIONS];
     }
 
-    function claimTokens(address _token, address _to) public onlyIfUpgradeabilityOwner validAddress(_to) {
+    /**
+     * @dev Withdraws the erc20 tokens or native coins from this contract.
+     * @param _token address of the claimed token or address(0) for native coins.
+     * @param _to address of the tokens/coins receiver.
+     */
+    function claimTokens(address _token, address _to) public onlyIfUpgradeabilityOwner {
         claimValues(_token, _to);
     }
 
--- a/contracts/upgradeable_contracts/BasicBridge.sol
+++ b/contracts/upgradeable_contracts/BasicBridge.sol
@@ -51,15 +51,6 @@ contract BasicBridge is
     }
 
     /**
-     * @dev Withdraws the erc20 tokens or native coins from this contract.
-     * @param _token address of the claimed token or address(0) for native coins.
-     * @param _to address of the tokens/coins receiver.
-     */
-    function claimTokens(address _token, address _to) public onlyIfUpgradeabilityOwner {
-        claimValues(_token, _to);
-    }
-
-    /**
     * @dev Internal function for updating fallback gas price value.
     * @param _gasPrice new value for the gas price, zero gas price is allowed.
     */
--- a/contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
+++ b/contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
@@ -207,9 +207,9 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, HomeF
     * @param _to address that will receive the locked tokens on this contract.
     */
     function claimTokens(address _token, address _to) external onlyIfUpgradeabilityOwner {
-        // Since native coins are being minted by the blockReward contract and burned by sending them to the address(0),
-        // they are not locked at the contract during the normal operation.
-        // So, any erc20 token and native coins can be safely claimed.
+        // In case native coins were forced into this contract by using a selfdestruct opcode,
+        // they should be handled by a call to fixMediatorBalance, instead of using a claimTokens function.
+        require(_token != address(0));
         claimValues(_token, _to);
     }
 
--- a/contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
+++ b/contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
@@ -202,13 +202,14 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, HomeF
     }
 
     /**
-    * @dev Allows to transfer any locked token on this contract that is not part of the bridge operations.
-    * Native tokens are not allowed to be claimed.
-    * @param _token address of the token.
+    * @dev Allows to transfer any locked tokens or native coins on this contract.
+    * @param _token address of the token, address(0) for native coins.
     * @param _to address that will receive the locked tokens on this contract.
     */
     function claimTokens(address _token, address _to) external onlyIfUpgradeabilityOwner {
-        require(_token != address(0));
+        // Since native coins are being minted by the blockReward contract and burned by sending them to the address(0),
+        // they are not locked at the contract during the normal operation.
+        // So, any erc20 token and native coins can be safely claimed.
         claimValues(_token, _to);
     }
 
--- a/contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
+++ b/contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
@@ -207,7 +207,7 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, HomeF
     * @param _token address of the token.
     * @param _to address that will receive the locked tokens on this contract.
     */
-    function claimTokens(address _token, address _to) external onlyIfUpgradeabilityOwner validAddress(_to) {
+    function claimTokens(address _token, address _to) external onlyIfUpgradeabilityOwner {
         require(_token != address(0));
         claimValues(_token, _to);
     }
@@ -217,7 +217,7 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, HomeF
     * without the invocation of the required methods.
     * @param _receiver the address that will receive the tokens on the other network
     */
-    function fixMediatorBalance(address _receiver) external onlyIfUpgradeabilityOwner {
+    function fixMediatorBalance(address _receiver) external onlyIfUpgradeabilityOwner validAddress(_receiver) {
         uint256 balance = address(this).balance;
         uint256 available = maxAvailablePerTx();
         if (balance > available) {
--- a/contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignAMBErc677ToErc677.sol
+++ b/contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignAMBErc677ToErc677.sol
@@ -95,4 +95,14 @@ contract ForeignAMBErc677ToErc677 is BasicAMBErc677ToErc677, MediatorBalanceStor
         _setMediatorBalance(mediatorBalance().sub(_value));
         erc677token().safeTransfer(_recipient, _value);
     }
+
+    /**
+    * @dev Allows to transfer any locked token on this contract that is not part of the bridge operations.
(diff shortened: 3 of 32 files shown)


k1rill-fedoseev:
After recent commits, this PR will also close #513

contracts/upgradeable_contracts/erc20_to_erc20/BasicForeignBridgeErcToErc.sol
```diff
@@ -34,7 +34,13 @@ contract BasicForeignBridgeErcToErc is BasicForeignBridge {
         return 0xba4690f5; // bytes4(keccak256(abi.encodePacked("erc-to-erc-core")))
     }
 
+    /**
+     * @dev Withdraws the erc20 tokens or native coins from this contract. Bridged token cannot be withdrawn by this function.
+     * @param _token address of the claimed token or address(0) for native coins.
+     * @param _to address of the tokens/coins receiver.
+     */
     function claimTokens(address _token, address _to) public {
```
akolotov, line 42: Shouldn't it be external?

contracts/upgradeable_contracts/BasicBridge.sol
```diff
@@ -50,7 +50,12 @@ contract BasicBridge is
         return uintStorage[REQUIRED_BLOCK_CONFIRMATIONS];
     }
 
-    function claimTokens(address _token, address _to) public onlyIfUpgradeabilityOwner validAddress(_to) {
+    /**
+     * @dev Withdraws the erc20 tokens or native coins from this contract.
+     * @param _token address of the claimed token or address(0) for native coins.
+     * @param _to address of the tokens/coins receiver.
+     */
+    function claimTokens(address _token, address _to) public onlyIfUpgradeabilityOwner {
```
akolotov, line 58: can we check whether it is invoked only from the derived contracts? If so, should it be `internal`?

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/ForeignMultiAMBErc20ToErc677.sol
```diff
@@ -201,7 +201,11 @@ contract ForeignMultiAMBErc20ToErc677 is BasicMultiAMBErc20ToErc677 {
...
     */
-    function fixMediatorBalance(address _token, address _receiver) public onlyIfUpgradeabilityOwner {
+    function fixMediatorBalance(address _token, address _receiver)
+        external
+        onlyIfUpgradeabilityOwner
+        validAddress(_receiver)
+    {
         require(isTokenRegistered(_token));
         uint256 balance = ERC677(_token).balanceOf(address(this));
         uint256 expectedBalance = mediatorBalance(_token);
```
akolotov, line 211: Should `mediatorBalance` be used from MediatorBalanceStorage.sol? It will require updating of `multi_amb_erc20_to_erc677/callflows.md`

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignAMBErc677ToErc677.sol
```diff
@@ -57,11 +81,28 @@ contract ForeignAMBErc677ToErc677 is BasicAMBErc677ToErc677 {
         bytes _data
     ) internal {
         if (!lock()) {
+            _setMediatorBalance(mediatorBalance().add(_value));
```
akolotov, line 84: should it be reflected in `amb_erc677_to_erc677/callflows.md`?

contracts/upgradeable_contracts/amb_native_to_erc20/HomeAMBNativeToErc20.sol
```diff
@@ -73,7 +72,7 @@ contract HomeAMBNativeToErc20 is BasicAMBNativeToErc20 {
         require(msg.value > 0);
         require(withinLimit(msg.value));
         addTotalSpentPerDay(getCurrentDay(), msg.value);
-        setMediatorBalance(mediatorBalance().add(msg.value));
+        _setMediatorBalance(mediatorBalance().add(msg.value));
```
akolotov, line 75: it should be reflected in `callflows.md` similar to other contracts supporting this feature.

akolotov (changes requested):

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/ForeignMultiAMBErc20ToErc677.sol
k1rill-fedoseev: No, we cannot use `mediatorBalance` from the `MediatorBalanceStorage`, since its signature is `mediatorBalance()`, not `mediatorBalance(address)`. So, this contract defines and uses its own versions of `mediatorBalance`/`_setMediatorBalance` functions.

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignAMBErc677ToErc677.sol
k1rill-fedoseev: In `amb_erc677_to_erc677/callflows.md` we have a behaviour described for Stake pair of mediators only. In Stake mediators, we do not need fixMediatorBalance functionality, since it operates on our own erc677 token.

contracts/upgradeable_contracts/amb_native_to_erc20/HomeAMBNativeToErc20.sol
k1rill-fedoseev: Added these calls to `amb_native_to_erc20/callflows.md` in 3f8366f

contracts/upgradeable_contracts/BasicBridge.sol
k1rill-fedoseev: Since this function contains only one line of code, I would prefer then to simply copy it there it is needed. In that case, we won't need internal functions at all. Also, it will be possible to mark those functions as `external` after such change.

contracts/upgradeable_contracts/erc20_to_erc20/BasicForeignBridgeErcToErc.sol
k1rill-fedoseev: Since this overrides and calls `super.claimTokens`, they both need to be public, this is forced by the solidity compiler.
But since we can get rid of `claimTokens` in `BasicBridge.sol`, it is possible to make it external.

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/ForeignMultiAMBErc20ToErc677.sol
akolotov: Will it increase the runtime byte code size if `mediatorBalance(address)` and `_setMediatorBalance(address)` are part of the MediatorBalanceStorage.sol?

contracts/upgradeable_contracts/amb_erc677_to_erc677/ForeignAMBErc677ToErc677.sol
akolotov: Thanks

contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
```diff
@@ -202,13 +202,14 @@ contract HomeAMBErc20ToNative is BasicAMBErc20ToNative, BlockRewardBridge, HomeF
...
-    * @param _token address of the token.
+    * @dev Allows to transfer any locked tokens or native coins on this contract.
+    * @param _token address of the token, address(0) for native coins.
     * @param _to address that will receive the locked tokens on this contract.
     */
     function claimTokens(address _token, address _to) external onlyIfUpgradeabilityOwner {
-        require(_token != address(0));
+        // Since native coins are being minted by the blockReward contract and burned by sending them to the address(0),
+        // they are not locked at the contract during the normal operation.
+        // So, any erc20 token and native coins can be safely claimed.
```
akolotov, line 212: it is not clear. what will happen if native tokens are sent by the `selfdestruct` call? They will not be burnt. But this imbalance must be addressed by usage the `fixMediatorBalance` functionality rather than by tokens claiming. So, I would leave the code restricting claiming the native tokens (token address == 0x0)

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
```diff
@@ -119,6 +119,18 @@ contract HomeBridgeErcToNative is
...
+    /**
+     * @dev Withdraws the erc20 tokens or native coins from this contract.
+     * @param _token address of the claimed token or address(0) for native coins.
+     * @param _to address of the tokens/coins receiver.
+     */
+    function claimTokens(address _token, address _to) external onlyIfUpgradeabilityOwner {
+        // Since native coins are being minted by the blockReward contract and burned by sending them to the address(0),
+        // they are not locked at the contract during the normal operation.
+        // So, any erc20 token and native coins can be safely claimed.
+        claimValues(_token, _to);
```
akolotov, line 131: What will happen if native tokens are sent by the `selfdestruct` call? They will not be burnt. Does it mean that we need to provide an ability to claim the native tokens anyway?

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/ForeignMultiAMBErc20ToErc677.sol
k1rill-fedoseev: Yes, it will increase the runtime byte code size for sure for all of the mediators contracts. I am not sure if a compiler can remove unused internal functions, but for sure it can't decide to remove public or external functions.

I have created a `MediatorBalanceStorage.sol` to extract the duplicated methods from regular mediators to a single file, but for multi token extension we don't need to do so, since we have only one such contract for now.

If it will be necessary in the future, we can create a `MediatorMultiBalanceStorage.sol` in order to keep  `mediatorBalance(address)` and `_setMediatorBalance(address)` there.

akolotov: > for sure it can't decide to remove public or external functions.

it makes sense

contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
k1rill-fedoseev: I thought it makes sense to have these both options available in such non-normal operation case: either to send tokens to the foreign side, or to claim them on the home side. In both cases, only the upgradeability owner can call these methods, and only he/she decides the receiver address, so I don't think there is a lack of some security here.

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
k1rill-fedoseev: Yes, correct, and since we do not have a mechanism like `fixMediatorBalance` in vanilla bridges, the only option is to use the `claimTokens` functionality.

akolotov: Ok. Let's add a comment here describing that we are allowing to claim native tokens in order to handle this particular case when tokens are sent as part of minting tokens for coinbase account (this configuration is weird but it could be still) and as part of the `selfdesctruct` call -- in both cases our fallback function will not be called.

contracts/upgradeable_contracts/amb_erc20_to_native/HomeAMBErc20ToNative.sol
akolotov: Well, consider to apply Occam's razor principle here and leave only one way to handle such situation in order to draw an unambiguous direction to solve the issue with native tokens assigned without invocation of the fallback method.

akolotov (approved):

PR #530, opened 2020-10-14
Update ERC677 behaviour

### LINKED ISSUE #516: ERC667 Compliance
As per security audit, the following was found:

The implemented ERC677 Tokens are extending `transferAndcall` logic to `transfer` and `transferFrom`, which is not compliant with the ERC667 standard. Using `callAfterTransfer` inside `transfer` or `transferFrom` function is problematic since it is not an intended behavior by the ERC667 standard. Even if the function does not throw, if the recipient implements `onTokenTransfer`, the transaction may behave differently since the user might not be expecting a call to the implemented `onTokenTransfer`.
Please note that even if this implementation enhances the UX for cross-chain token transfers, the implemented tokens will still be widely used to interact with their respective blockchain ecosystems.

**Recommendation:** Only call `onTokenTransfer` through `transferAndCall` or if the recipient address in the `transfer`/`transferFrom` function is the associated bridge contract address.

k1rill-fedoseev:
Closes #516

### DIFF
--- a/contracts/ERC677BridgeToken.sol
+++ b/contracts/ERC677BridgeToken.sol
@@ -16,8 +16,6 @@ contract ERC677BridgeToken is IBurnableMintableERC677Token, DetailedERC20, Burna
 
     address internal bridgeContractAddr;
 
-    event ContractFallbackCallFailed(address from, address to, uint256 value);
-
     constructor(string _name, string _symbol, uint8 _decimals) public DetailedERC20(_name, _symbol, _decimals) {
         // solhint-disable-previous-line no-empty-blocks
     }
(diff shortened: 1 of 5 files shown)


contracts/ERC677BridgeToken.sol
```diff
@@ -68,9 +66,8 @@ contract ERC677BridgeToken is IBurnableMintableERC677Token, DetailedERC20, Burna
     }
 
     function callAfterTransfer(address _from, address _to, uint256 _value) internal {
```
akolotov, line 68: Consider to add comment describing that we don't follow the erc677 standard only for the bridge contract.

k1rill-fedoseev: Sure, did it.

akolotov (approved):

PR #534, opened 2020-10-15
Refactor BlockReward.reward() function

### LINKED ISSUE #520: Reward function is underspecified
As per security audit, several points were found in the `BlockReward.reward` function that can be improved:

* Document why the lengths of its input parameters supposed to be zero.
* Document how the receivers different from the benefactors.
* Merge two consecutive loops with the same iterator range `0..extraLength` into 1 loop.
* There seems to be an invariant that is not explicitly checked, namely: the sum of `extraAmounts` for all `extraReceivers` is equal to the sum of `bridgeAmounts` for all allowed bridge addresses. Add the corresponding assertion statement.

k1rill-fedoseev:
Closes #520


contracts/upgradeable_contracts/amb_erc20_to_native/BlockReward.sol
```diff
@@ -75,32 +75,42 @@ contract BlockReward is EternalStorage {
...
 
+        // Lists with extra receivers and their rewards. Extra receivers are generated by the bridge/mediator contracts,
+        // and they are not passed in the benefactors array.
         address[] memory receivers = new address[](extraLength);
         uint256[] memory rewards = new uint256[](extraLength);
 
         uint256 i;
 
+        uint256 sumOfRewards;
+        uint256 sumOfBridgeAmounts;
```
akolotov, line 104: should we initialize them with 0s explicitly?

akolotov (approved):

contracts/upgradeable_contracts/amb_erc20_to_native/BlockReward.sol
k1rill-fedoseev: I am quite sure that Solidity compiler enforces zero initial values by default. But using explicit zeros seems nice for me as well.

PR #538, opened 2020-10-18
Disable fee collection if sender is a reward address

### LINKED ISSUE #505: Allow to withdraw a bridged token completely on Omnibridge
The current implementation of the fee manager on the OmniBridge (multi-token AMB extension) do not allow to completely withdraw the balance of the bridged token from the home chain to the foreign chain. Some small amount of tokens will be always distributed again among the accounts specified as the fee receivers.

Consider to check if the sender of the tokens is one of the fee receivers and to send this fraction of fee through the bridge instead of transferring it within the home chain.

k1rill-fedoseev:
Closes #505

### DIFF
--- a/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
+++ b/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
@@ -249,10 +249,12 @@ contract HomeMultiAMBErc20ToErc677 is BasicMultiAMBErc20ToErc677, HomeFeeManager
         if (!lock()) {
             bytes32 _messageId = messageId();
             uint256 valueToBridge = _value;
-            uint256 fee = _distributeFee(HOME_TO_FOREIGN_FEE, _token, valueToBridge);
-            if (fee > 0) {
-                emit FeeDistributed(fee, _token, _messageId);
-                valueToBridge = valueToBridge.sub(fee);
+            if (!isRewardAddress(_from)) {
+                uint256 fee = _distributeFee(HOME_TO_FOREIGN_FEE, _token, valueToBridge);
+                if (fee > 0) {
+                    valueToBridge = valueToBridge.sub(fee);
+                    emit FeeDistributed(fee, _token, _messageId);
+                }
             }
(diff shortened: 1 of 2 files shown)


contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
```diff
@@ -249,10 +249,12 @@ contract HomeMultiAMBErc20ToErc677 is BasicMultiAMBErc20ToErc677, HomeFeeManager
         if (!lock()) {
             bytes32 _messageId = messageId();
             uint256 valueToBridge = _value;
-            uint256 fee = _distributeFee(HOME_TO_FOREIGN_FEE, _token, valueToBridge);
-            if (fee > 0) {
-                emit FeeDistributed(fee, _token, _messageId);
-                valueToBridge = valueToBridge.sub(fee);
+            if (!isRewardAddress(_from)) {
```
akolotov, line 252: Imagine that we have a dozen of the fee recipients so I think that others receive their fraction of fee from this withdrawal operation. So, I would say that `_distributeFee` must accept a parameter with the the current sender and if the sender is the fee recipient don't transfer tokens to it. The resulting fee will be `calculateFee(_feeType, _token, _value)` minus `feePerAccount` (`diff` also should be taken into account)

k1rill-fedoseev: I think this won't solve the general problem. Imagine a situation where we have 2 reward addresses, 10% fee, first reward account owns 100 tokens. 
If first reward account withdraws 100 tokens, second account will receive 10% / 2 = 5% fee, that is 5 tokens.
If now second account will try to withdraw his 5 tokens, first account will receive 0.25 tokens. So there won't be a finite way to withdraw all tokens from the home side.

akolotov: Understood. So, let's add the comment to this line of the code in order to not forget this clarification.

akolotov:
Do you think this issue is applicable for other modes of the bridge? If so, please create an issue. We will address it later.

akolotov (approved):

PR #539, opened 2020-10-18
Fix behaviour for the unset fee manager contract

k1rill-fedoseev:
This PR does the following:
 * Refactors some of the fee manager functions
 * Fixes behaviour for the unset/invalid fee manager contract

### DIFF
--- a/contracts/upgradeable_contracts/RewardableBridge.sol
+++ b/contracts/upgradeable_contracts/RewardableBridge.sol
@@ -9,16 +9,39 @@ contract RewardableBridge is Ownable, FeeTypes {
     event FeeDistributedFromSignatures(uint256 feeAmount, bytes32 indexed transactionHash);
 
     bytes32 internal constant FEE_MANAGER_CONTRACT = 0x779a349c5bee7817f04c960f525ee3e2f2516078c38c68a3149787976ee837e5; // keccak256(abi.encodePacked("feeManagerContract"))
+    bytes4 internal constant GET_HOME_FEE = 0x94da17cd; // getHomeFee()
+    bytes4 internal constant GET_FOREIGN_FEE = 0xffd66196; // getForeignFee()
+    bytes4 internal constant GET_FEE_MANAGER_MODE = 0xf2ba9561; // getFeeManagerMode()
     bytes4 internal constant SET_HOME_FEE = 0x34a9e148; // setHomeFee(uint256)
     bytes4 internal constant SET_FOREIGN_FEE = 0x286c4066; // setForeignFee(uint256)
     bytes4 internal constant CALCULATE_FEE = 0x9862f26f; // calculateFee(uint256,bool,bytes32)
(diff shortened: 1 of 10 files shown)


akolotov (commented):
Is it possible to leave only the fix for "behaviour for the unset/invalid fee manager contract" without refactoring the rest of the code? I doubt that increasing the delta without improving functionality is good for further security audit: more delta -> more efforts to review -> higher cost of the audit.

contracts/upgradeable_contracts/RewardableBridge.sol
```diff
@@ -18,38 +18,30 @@ contract RewardableBridge is Ownable, FeeTypes {
...
         bytes4 method = _feeType == HOME_FEE ? GET_HOME_FEE : GET_FOREIGN_FEE;
         bytes memory callData = abi.encodeWithSelector(method);
 
         assembly {
             let result := callcode(gas, feeManager, 0x0, add(callData, 0x20), mload(callData), 0, 32)
-            fee := mload(0)
 
-            switch result
-                case 0 {
-                    revert(0, 0)
```
akolotov, line 33: Before the changes, if the result was unsuccessful, the call reverted. It means that it could revert entire transaction. Are there cases when this code is invoked during a transaction execution rather than a method call? If so, is this code safe to be changed in this manner?

k1rill-fedoseev: So, here is the quick analysis. We have 5 places with such `callcode` patterns:
* `getFeeManagerMode()`, method is not called on-chain, nothing unsafe here.
* `_getFee(bytes32)` is called inside `getHomeFee()` and `getForeignFee()` of `Rewardable*Bridge*.sol` set of contracts. These methods are not called onchain inside our contracts.
* `calculateFee(uint256,bool,address,bytes32)` is called in several vanilla modes contracts.
  - In all cases, a call is located within a `if (feeManager != address(0))` condition. So, the behaviour for an unset fee manager will be exactly the same as before. 
  - For invalid fee manager contracts, zero will be returned instead of a transaction revert, meaning that setting a fee manager to an invalid contract will result in disabled fees rather than disabled bridge operations, I don't see a problem in such behaviour.
  - For valid contracts, the underlying call might only fail due to `SafeMath` exceptions. In theory, if someone want to bridge `2**200` tokens, then fee manager will revert, but the outside transaction will continue its execution assuming that fee is zero. I think it is better to revert the whole transaction in such situation, so let's ad an extra check.
* `blockRewardContract()` from `HomeBridgeErcToErcPOSDAO` is not called inside our contracts, so it is safe to return zero address.
* `getAmountToBurn(uint256)` works in exactly the same way as `calculateFee` does. We can also add an extra check there.

k1rill-fedoseev: Added revert statements for extra safety where needed, please take a look 10d7c0f

akolotov: thanks for analysis!

akolotov (approved):

PR #545, opened 2020-10-29
Calculate fee amount from collected signatures only once

k1rill-fedoseev:
This PR fixes a possible problem when fee values are updated between `UserRequestForSignature` and `CollectedSignatures` events happened for a particular transfer. This can lead to incorrect balances on different sides of the bridge.

### DIFF
--- a/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
+++ b/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
@@ -192,6 +192,11 @@ contract HomeBridgeErcToNative is
         return true;
     }
 
+    /**
+     * @dev Internal function to be called when enough signatures are collected.
+     * Distributed the fee for collecting signatures.
+     * @param _message encoded message signed by the validators.
+     */
     function onSignaturesCollected(bytes _message) internal {
         (address recipient, uint256 amount, bytes32 txHash, ) = Message.parseMessage(_message);
         (address feeManager, uint256 fee) = _restoreCalculatedFee(recipient, amount);
--- a/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
+++ b/contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
@@ -39,10 +39,9 @@ contract HomeBridgeErcToNative is
         address feeManager = feeManagerContract();
         uint256 valueToBurn = msg.value;
         if (feeManager != address(0)) {
-            uint256 fee = calculateFee(valueToTransfer, feeManager, HOME_FEE);
+            uint256 fee = calculateFee(valueToTransfer, false, feeManager, HOME_FEE);
             valueToTransfer = valueToTransfer.sub(fee);
             valueToBurn = getAmountToBurn(valueToBurn);
-            _saveCalculatedFee(_receiver, valueToTransfer, feeManager, fee);
         }
         setTotalBurntCoins(totalBurnt.add(valueToBurn));
         address(0).transfer(valueToBurn);
@@ -184,7 +183,7 @@ contract HomeBridgeErcToNative is
         uint256 valueToMint = _shiftValue(_value);
         address feeManager = feeManagerContract();
         if (feeManager != address(0)) {
-            uint256 fee = calculateFee(valueToMint, feeManager, FOREIGN_FEE);
+            uint256 fee = calculateFee(valueToMint, false, feeManager, FOREIGN_FEE);
             distributeFeeFromAffirmation(fee, feeManager, _txHash);
             valueToMint = valueToMint.sub(fee);
         }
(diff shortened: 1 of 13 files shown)


contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
```diff
@@ -183,25 +184,23 @@ contract HomeBridgeErcToNative is
...
-            uint256 amount;
-            bytes32 txHash;
-            address contractAddress;
-            (recipient, amount, txHash, contractAddress) = Message.parseMessage(_message);
-            uint256 fee = calculateFee(amount, true, feeManager, HOME_FEE);
-            distributeFeeFromSignatures(fee, feeManager, txHash);
-        }
+        (address recipient, uint256 amount, bytes32 txHash, ) = Message.parseMessage(_message);
+        (address feeManager, uint256 fee) = _restoreCalculatedFee(recipient, amount);
+        distributeFeeFromSignatures(fee, feeManager, txHash);
```
akolotov, line 203: It looks like this approach will not fix the original issue.
Imagine the situation when the finalization threshold on the bridge is 20, and there is a service which sends some reward through the bridge every 10 blocks.

Pre-req: the fee is `f1` for the block `X-1`
Step 1: the block `X`, the service sends the reward `5` to Alice, the `f1*5` tokens are not burnt.
Step 2: the block `X+5`, the bridge governance changes the fee to `f2`
Step 3: the block `X+10`, the service sends the reward `5` to Alice and stops its operations, the `f2*5` tokens are not burnt.
Step 4: the block `X+20`, the oracles send their signatures and `f2*5` tokens are distributed among the reward recipients
Step 5: the block `X+30`, the oracles send their signatures and if `f2` < `f1`, `f2*5` tokens are distributed among the reward recipients but `(f1-f2)*5` tokens will remain on the bridge contract balance. If `f2` > `f1`, the reward distribution will fails since the balance of the bridge contract is less than required.

akolotov (commented):
I have left a comment where presented that the issue is still persist with this approach. Seems that the original issue cannot be addressed with the current bridge message format. Since the case of changing the fee manager or fee value is very rare, consider to clearly state in the methods that change the fee manager and the fee value that they must be performed only if the bridge operations are disabled (e.g. with setting the max per tx limit to 0) otherwise the bridge imbalance is possible. I believe it should be enough to address the original issue. What do you think?

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
k1rill-fedoseev: Actually, not exactly, in the current implementation, the second transfer cannot be submitted and processed until the signatures for the first transfer are collected. 

At https://github.com/poanetwork/tokenbridge-contracts/blob/62cab6e866e382b9f87f8945c93d9e1be6cadbf4/contracts/upgradeable_contracts/RewardableBridge.sol#L144-L151 I ensure that there are no any pending fees for the pair `(user, amount)`. Until this pending fee will be popped out from storage by https://github.com/poanetwork/tokenbridge-contracts/blob/62cab6e866e382b9f87f8945c93d9e1be6cadbf4/contracts/upgradeable_contracts/RewardableBridge.sol#L159-L166, user won't be able to submit exactly the same transfer.

I think it is possible to extend this functionality further, so that we can avoid such blocking behaviour, but for that we will need to maintain the stack/queue of pending transfer fees for a particular pair of `(user, amount)`. It will increase transaction gas usage, but since such logic is located on the home side of the bridge, it should not be a big problem.

What do you think of a such approach?

contracts/upgradeable_contracts/native_to_erc20/ForeignBridgeNativeToErc.sol
```diff
@@ -123,11 +123,9 @@ contract ForeignBridgeNativeToErc is
...
         address feeManager = feeManagerContract();
         if (feeManager != address(0)) {
-            uint256 fee = calculateFee(valueToMint, false, feeManager, HOME_FEE);
-            if (fee != 0) {
-                distributeFeeFromSignatures(fee, feeManager, _txHash);
-                valueToMint = valueToMint.sub(fee);
-            }
+            uint256 fee = calculateFee(valueToMint, feeManager, HOME_FEE);
+            distributeFeeFromSignatures(fee, feeManager, _txHash);
+            valueToMint = valueToMint.sub(fee);
```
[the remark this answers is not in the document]
k1rill-fedoseev: Ok, we can save a small amount of gas by overriding a `distributeFeeFromSignatures` function for this particular case.

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
akolotov: > the second transfer cannot be submitted and processed until the signatures for the first transfer are collected.

OK. I missed this part. I believe it will be hard for a service developer to explain such code behavior. Especially it is risky for the cases when the oracles are experiencing the issue with the RPC connectivity and that is why they do not provide signatures for few minutes. It could break the logic of that 3rd party service.

akolotov: > I think it is possible to extend this functionality further, so that we can avoid such blocking behaviour, but for that we will need to maintain the stack/queue of pending transfer fees for a particular pair of (user, amount). It will increase transaction gas usage, but since such logic is located on the home side of the bridge, it should not be a big problem.

Based on the previous experience it seems that it is cheaper to assume that we can manage the stop of the bridge operations for such kind of maintenance actions. The situation to change the fee value or the fee manager are quite rare so don't think it makes sense to increase the cost of transactions for the end user just because we want to make sure that the imbalance will not happen during the maintenance period that could happen once or twice during the bridge lifetime.

k1rill-fedoseev: Ok, I added a comment with a recommendation of stopping the bridge operations before updating fee manager contracts / home fee percentage.
Does it also make sense to add a `require` statements to ensure such behaviour? So that we do not allow to change home -> foreign fee configuration until the bridge is stopped?

akolotov (approved):

contracts/upgradeable_contracts/erc20_to_native/HomeBridgeErcToNative.sol
akolotov: > Does it also make sense to add a require statements to ensure such behaviour?

No, the comment is enough. Will not increase the code delta without a real necessity.

PR #546, opened 2020-10-30
Support manual lane in AMB contracts

### LINKED ISSUE #542: Manual lane introduction for AMB messages relaying to Foreign side
Since in the subsidized mode the oracles perform their signatures relay to the Foreign chain, some AMB extensions can be attacked to generate lots of transfers to the Foreign side through the bridge and it will dry out the oracles accounts balances. 

It makes sense to allow AMB extensions to have an ability to indicate to the AMB oracles that some messages should not be relayed in the subsidized mode. The oracles will provide their signatures as confirmations only so either the extension owners or the account originating the message will be responsible for transferring the signatures to the Foreign chain. These messages will be called "forwarded to the manual lane". The messages which confirmations are still relayed by oracles on the Foreign side will be called "forwarded to the oracle driven lane".

In order to achieve this the suggestion is to reuse [the `dataType` field](https://github.com/poanetwork/tokenbridge-contracts/blob/b22383f351b75f74d1926c9a6a10493af8209ce4/contracts/libraries/ArbitraryMessage.sol#L14) in an AMB message.

If a mediator requires to indicate that the message should not be transferred in the subsidized mode it will call a new method `requireToConfirmMessage(address _contract, bytes _data, uint256 _gas)` instead of `requireToPassMessage`. When this method is called the AMB contracts prepare the AMB message as so the 8th (the most significant) bit of the `dataType` field is set. This bit will be inspected later by the AMB oracles in order to handle the message properly.

k1rill-fedoseev:
Closes #542

### DIFF
--- a/contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
+++ b/contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
@@ -15,6 +15,8 @@ contract BasicHomeAMB is BasicAMB, MessageDelivery {
         uint256 NumberOfCollectedSignatures
     );
 
+    uint256 internal constant SEND_TO_MANUAL_LANE = 0xf0;
+
     function executeAffirmation(bytes message) external onlyValidator {
         bytes32 hashMsg = keccak256(abi.encodePacked(message));
         bytes32 hashSender = keccak256(abi.encodePacked(msg.sender, hashMsg));
@@ -44,7 +46,7 @@ contract BasicHomeAMB is BasicAMB, MessageDelivery {
     * @param _gas gas limit used on the other network for executing a message.
     */
     function requireToConfirmMessage(address _contract, bytes _data, uint256 _gas) external returns (bytes32) {
-        return _sendMessage(_contract, _data, _gas, 0xf0);
+        return _sendMessage(_contract, _data, _gas, SEND_TO_MANUAL_LANE);
     }
 
     /**
--- a/contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
+++ b/contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
@@ -44,7 +44,7 @@ contract BasicHomeAMB is BasicAMB, MessageDelivery {
     * @param _gas gas limit used on the other network for executing a message.
     */
     function requireToConfirmMessage(address _contract, bytes _data, uint256 _gas) external returns (bytes32) {
-        return requireToPassMessage(_contract, _data, _gas);
+        return _sendMessage(_contract, _data, _gas, 0xf0);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -66,14 +66,16 @@ contract MessageDelivery is BasicAMB, MessageProcessor {
         // 79 = 4 + 20 + 8 + 20 + 20 + 4 + 1 + 1 + 1
         header = new bytes(79 + srcChainIdLength + dstChainIdLength);
 
+        uint256 datatype = _messageDatatype();
```
akolotov, line 69: Should we increase the version of AMB contracts now in order to reflect that their functionality was changed?

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
```diff
@@ -37,6 +37,16 @@ contract BasicHomeAMB is BasicAMB, MessageDelivery {
...
     }
 
+    /**
+    * @dev Requests message relay to the opposite network, message is sent to the manual lane.
+    * @param _contract executor address on the other side.
+    * @param _data calldata passed to the executor on the other side.
+    * @param _gas gas limit used on the other network for executing a message.
+    */
+    function requireToConfirmMessage(address _contract, bytes _data, uint256 _gas) external returns (bytes32) {
+        return requireToPassMessage(_contract, _data, _gas);
```
akolotov, line 47: does it make sense to add a comment here describing that the behavior of requireToPassMessage will be changed since internally it checks the selector of invoked method. But I would strongly recommend to change the code as so we manage the dataType byte more explicitly -- the current method is very smart but it could be difficult to keep such trick in the mind every time the code is read.

akolotov (changes requested):

contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
k1rill-fedoseev: Bumped AMB contracts version to `5.4.0`

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
k1rill-fedoseev: Ok, I have changed the way it works, now it should be better.

contracts/upgradeable_contracts/arbitrary_message/BasicHomeAMB.sol
```diff
@@ -37,6 +37,16 @@ contract BasicHomeAMB is BasicAMB, MessageDelivery {
...
     }
 
+    /**
+    * @dev Requests message relay to the opposite network, message is sent to the manual lane.
+    * @param _contract executor address on the other side.
+    * @param _data calldata passed to the executor on the other side.
+    * @param _gas gas limit used on the other network for executing a message.
+    */
+    function requireToConfirmMessage(address _contract, bytes _data, uint256 _gas) external returns (bytes32) {
+        return _sendMessage(_contract, _data, _gas, 0xf0);
```
akolotov, line 47: does it make sense to introduce constants for `0xf0` and `0` used for the `_sendMessage` call?

akolotov (approved):

PR #547, opened 2020-11-02
Omnibridge usage of manual/oracle-driven lanes

### LINKED ISSUE #543: Manual lane introduction on the OmniBridge contracts for transfers to Foreign side
As soon as #542 is implemented, it makes sense to change [the OmniBridge contracts](https://github.com/poanetwork/tokenbridge-contracts/tree/master/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677) functionality to support a new way to disable handling of AMB messages in the subsidized mode ("manual lane forwarding")

The OmniBridge contracts on the Home side must allow to
  * [ ] configure an address of a token as so any relay operation to the Foreign chain for this token will generate a message to be forwarded to the manual lane; there possible to have several tokens configured for the manual lane
  * [ ] configure an address of a recipient as so any relay operation of any token to this recipient on the Foreign chain will generate a message to be forwarded to the manual lane (_inclusion rule_)
  * [ ] configure an address of a sender as so any relay operation from this sender of any token will generate a message to be forwarded to the manual lane (_inclusion rule_)
  * [ ] configure an address of a recipient as so any relay operation of the token already configured for the manual lane to this recipient on the Foreign chain will generate a message to be forwarded to the oracle driven lane (_exception rule_)
  * [ ] configure an address of a sender as so any relay operation from this sender of the token already configured for the manual lane will generate a message to be forwarded to the oracle driven lane (_exception rule_)

### LINKED ISSUE #544: Manual lane support on the OmniBridge AMB extension for transfer to Foreign chain
As soon as #542 is implemented, it makes sense to change [the OmniBridge contracts](https://github.com/poanetwork/tokenbridge-contracts/tree/master/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677) functionality to support a new way to disable handling of AMB messages in the subsidized mode ("manual lane forwarding")

The OmniBridge contracts on the Home side must allow to
  * [ ] configure an address of a token as so any relay operation to the Foreign chain for this token will generate a message to be forwarded to the manual lane; there possible to have several tokens configured for the manual lane
  * [ ] configure an address of a recipient as so any relay operation of any token to this recipient on the Foreign chain will generate a message to be forwarded to the manual lane (_inclusion rule_)
  * [ ] configure an address of a sender as so any relay operation from this sender of any token will generate a message to be forwarded to the manual lane (_inclusion rule_)
  * [ ] configure an address of a recipient as so any relay operation of the token already configured for the manual lane to this recipient on the Foreign chain will generate a message to be forwarded to the oracle driven lane (_exception rule_)
  * [ ] configure an address of a sender as so any relay operation from this sender of the token already configured for the manual lane will generate a message to be forwarded to the oracle driven lane (_exception rule_)

k1rill-fedoseev:
Closes #544


contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenLanePermissions.sol
```diff
@@ -0,0 +1,116 @@
...
+    uint256 internal constant RECEIVER_DESTINATION_LANE_TAG = 2;
+
+    /**
+     * @dev Checks the oracle driven lane permissions for a particular bridge operation.
+     * @param _token address of the token contract on the foreign side of the bridge.
+     * @param _sender address of the tokens sender on the home side of the bridge.
+     * @param _receiver address of the tokens receiver on the foreign side of the bridge.
+     * @return true, if oracle driven lane should be used for the operation.
+     */
+    function oracleDrivenLaneAllowed(address _token, address _sender, address _receiver) public view returns (bool) {
```
akolotov, line 21: It seems for me that the functionality is not implemented in the form as the initial requirements (#544) assume:

Let me re-phrase the code of `oracleDrivenLaneAllowed`:
  - if the sender was configured with a positive integer it will be always forwarded to the oracle driven lane regardless to the token address
  - if the receiver was configured with a positive integer it will be always forwarded to the oracle driven lane regardless to the token address
  - if the sender was configured with a negative integer it will be always forwarded to the manual lane regardless to the token address
  - if the token was configured with a negative integer it will be always forwarded to the manual lane regardless to the token address
  - if the receiver was configured with a negative integer it will be always forwarded to the manual lane regardless to the token address

It differ from the initial requirement in those moments which relate to exclusion rules. The idea there is to allow some accounts to send/receive a particular token automatically even if the relay operations for this token are forwarded by the manual lane.

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenLanePermissions.sol
```diff
@@ -0,0 +1,116 @@
+pragma solidity 0.4.24;
+
+import "../../upgradeable_contracts/Ownable.sol";
+
+/**
+ * @title MultiTokenLanePermissions
+ * @dev Multi token mediator functionality for managing destination AMB lanes permissions.
+ */
+contract MultiTokenLanePermissions is Ownable {
+    uint256 internal constant TOKEN_DESTINATION_LANE_TAG = 0;
```
akolotov, line 10: could you confirm that these constant are not held on the storage so we don't need to assume a special procedure to initialize them as part of the contracts upgrade procedure.

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
```diff
@@ -273,20 +278,25 @@ contract HomeMultiAMBErc20ToErc677 is BasicMultiAMBErc20ToErc677, HomeFeeManager
...
-        bytes32 _messageId = bridgeContract().requireToPassMessage(
-            mediatorContractOnOtherSide(),
-            data,
-            requestGasLimit()
-        );
+        address executor = mediatorContractOnOtherSide();
+        uint256 gasLimit = requestGasLimit();
+        IAMB bridge = bridgeContract();
+
+        bytes32 _messageId = oracleDrivenLaneAllowed(foreignToken, _from, _receiver)
```
akolotov, line 292: Consider also to increase version of the multi-token extension contracts.

akolotov (changes requested):

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenLanePermissions.sol
k1rill-fedoseev: Yes, I can confirm that, such internal constants go directly to the bytecode, nothing special is needed to use them under a proxy contract. In other contracts, we have similar constants as well: `MAX_FEE`, `MAX_VALIDATORS`, etc..

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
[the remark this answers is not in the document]
k1rill-fedoseev: Yes, setting a rule for unbridged tokens looks like a benefit for me. I think adding an explicit foolproof `require` statement makes sense, if the default destination lane (if destination lanes are not set for sender/receiver/token, use the manual lane). However, now, the default behaviour is to use an oracle-driven lane, so there won't be a way to block the token until it first bridged. Don't you think that this can be a downside of a such check?

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenLanePermissions.sol
k1rill-fedoseev: Do you mean that the sender/receiver specific settings should be configured for a particular token, not for all of them? So that the following situation is possible:
* tokens A and B are set to use the manual lane by default
* address C is allowed to use oracle-driven lane for token A, but not for token B

akolotov: yes, it was an intention -- to allow a particular account to transfer particular token automatically even if the token is intended to go through the manual lane.

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenLanePermissions.sol
akolotov: thanks

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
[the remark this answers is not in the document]
akolotov: The functionality to forward messages operating with the particular pair of tokens to the manual lane is needed to stop attacks to dry out the validators' account balance. So, the scenario could be the following: someone create a fake erc20 token in the mainnet and transfer a billion of tokens to the xDai chain, then they relay small amounts of the token back to the Ethereum Mainnet. We don't know in advance which token will be transferred from the Mainnet that's why I think that the scenario to block a token which is not bridged yet is rare. But we definitely would have a motivation to handle a particular bridged token as so the oracles will not send signatures to the Ethereum Mainnet to finish the relay operation.

Anyway, it is not a big deal if you think that the lane for the tokens relaying operation must be identified by the token address on the foreign side. Just document this properly as so the code reviewers (e.g. security auditors) in the future could understand our reasoning. So, add the comment here and in the method `setTokenDestinationLane`.

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenLanePermissions.sol
akolotov: The main reason from my side is that most probably there will be less cases when lots of tokens should b transferred by/to the particular account where as it could be a case when a business logic of an application is build as so some application-specific token is transferred to the mainnet by an autonomous agent (that could have no ether, only xdai) to an alternative account.

akolotov: Seems that the following requirements are not addressed now:
  - configure an address of a recipient as so any relay operation of any token to this recipient on the Foreign chain will generate a message to be forwarded to the manual lane (inclusion rule)
  - configure an address of a sender as so any relay operation from this sender of any token will generate a message to be forwarded to the manual lane (inclusion rule)

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenForwardingRules.sol
```diff
@@ -0,0 +1,76 @@
...
+     * Makes a top-down checking of forwarding rules, from specific ones to more generic ones.
+     * @param _token address of the token contract on the foreign side of the bridge.
+     * @param _sender address of the tokens sender on the home side of the bridge.
+     * @param _receiver address of the tokens receiver on the foreign side of the bridge.
+     * @return destination lane identifier, where the message should be forwarded to.
+     *  1 - oracle-driven-lane should be used.
+     *  0 - default behaviour should be applied.
+     * -1 - manual lane should be used.
+     */
+    function destinationLane(address _token, address _sender, address _receiver) public view returns (int256) {
```
akolotov, line 25: This approach is very straightforward, thanks! But the main drawback here is excessive gas usage: for the worst case it will require 8 RSTORE and 8 keccak operations.

Can we optimize it a bit?

1. `forwardingRule(ANY_ADDRESS, ANY_ADDRESS, ANY_ADDRESS)` --  will be always 0 for the multi-token extension between Ethereum and xDai: we either subsidize its operations or not -- it will be configured through the AL on the oracle side

2. `forwardingRule(_token, _sender, _receiver)` and `forwardingRule(ANY_ADDRESS, _sender, _receiver)` are too specific and I cannot see any useful case when we could enable it.

3. As per the original requirements a positive returned value of `forwardingRule(_token, _sender, ANY_ADDRESS)` or `forwardingRule(_token, ANY_ADDRESS, _receiver)` makes sense only if `forwardingRule(_token, ANY_ADDRESS, ANY_ADDRESS)` returns a negative value. The case when `forwardingRule(_token, ANY_ADDRESS, ANY_ADDRESS)` returns a positive value is not going to be used.

akolotov (changes requested):

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/MultiTokenForwardingRules.sol
k1rill-fedoseev: Agree on that, I optimized the number of used of storage read operations according to your comments.
Please, take a look.

akolotov: Do you think that this snippet will be closer to the initial requirements?

```
lane = forwardingRule(_token, ANY_ADDRESS, ANY_ADDRESS);
if (lane < 0) {
    int256 lane_n = forwardingRule(_token, _sender, ANY_ADDRESS);
    if (lane_n > 0) return 1;
    lane_n = forwardingRule(_token, ANY_ADDRESS, _receiver);
    if (lane_n > 0) return 1;
}
if (lane != 0) return lane;
```

In this case it seems logical to introduce `setTokenFwdRule`, `setSenderExceptionForTokenFwdRule`, `setReceiverExceptionForTokenFwdRule`, `setSenderFwdRule` and `setReceiverFwdRule`. They will call `setForwardingRule` (it should be internal in this case) and force: 
  * `setForwardingRule` to be called with `-1` in `setTokenFwdRule`
  * `setForwardingRule` to be called with `1` in `setSenderExceptionForTokenFwdRule`
  * `setForwardingRule` to be called with `1` in `setReceiverExceptionForTokenFwdRule`
  * `setForwardingRule` to be called with `-1` in ``setSenderFwdRule`
  * `setForwardingRule` to be called with `-1` in ``setReceiverFwdRule`

What do you think?

k1rill-fedoseev: How does this snippet handles the case with inclusion rules?

akolotov: Sorry, I missed the commit "Reduce amount of SLOAD operations required" where you already implemented this logic. It looks very similar to what I am suggesting:
```
lane = forwardingRule(_token, ANY_ADDRESS, ANY_ADDRESS);
if (lane < 0) {
    int256 lane_n = forwardingRule(_token, _sender, ANY_ADDRESS);
    if (lane_n > 0) return 1;
    lane_n = forwardingRule(_token, ANY_ADDRESS, _receiver);
    if (lane_n > 0) return 1;
}
if (lane != 0) return lane;

lane = forwardingRule(ANY_ADDRESS, _sender, ANY_ADDRESS); 
if (lane != 0) return lane;
return forwardingRule(ANY_ADDRESS, ANY_ADDRESS, _receiver);
```

Together with the logic I suggested for the methods `setTokenFwdRule`, `setSenderExceptionForTokenFwdRule`, `setReceiverExceptionForTokenFwdRule`, `setSenderFwdRule` and `setReceiverFwdRule` it could correspond to the requirements introduced initially.

akolotov (approved):

PR #552, opened 2020-11-11
Modular omnibridge

k1rill-fedoseev:
This PR splits the Multi-token extension (OmniBridge) on separate modules, this is required due to reached contract size limit of 24 KB on the Home side of the bridge.

* Old version of `HomeMultiAMBErc20ToErc677` (was 24 KB) was split on the following pieces:
  * `HomeMultiAMBErc20ToErc677` - upgradeable core contract (13.5 KB)
  * `MultiTokenFeeManager` - non-upgradeable module contract (4 KB)
  * `MultiTokenForwardingRulesManager` - non-upgradeable module contract (1.5 KB)
  * `MultiTokenBridgeLimitsManager` - non-upgradeable module contract (5 KB)
  * `TokenFactory` - non-upgradeable module contract (2 KB)
* Old version of `FomeMultiAMBErc20ToErc677` (was 16.5 KB) was split on the following pieces:
  * `FomeMultiAMBErc20ToErc677 ` - upgradeable core contract (11.5 KB)
  * `MultiTokenBridgeLimitsManager` - non-upgradeable module contract (5 KB)

Such change to the modular system brings other useful advantages as well:
* Cheaper core contract implementation upgrade due to the smaller bytecode size.
* Possible functional changes without upgrading core contract at all. This can be achieved by changing a corresponding module address to a newer version.

In order to allow a smooth migration from the already in-use old Omnibridge contracts, some modules use lazy initialisation, (e. g. bridge limits/per-token fees are initialized at the very first operation involving them, not during token creation). A proxy upgrade procedure will probably require some additional one-time method for initialisation of modules addresses at the upgrade time, so that an upgrade can be performed without stopping the bridge operation.


akolotov:
This was implemented in a dedicated repo: https://github.com/poanetwork/omnibridge.

PR #553, opened 2020-11-12
Optimize Omnibridge gas usage

k1rill-fedoseev:
This PR slightly reduces the gas usage for `bridgeSpecificActionsOnTokenTransfer` function in the MultiToken extension.

The following actions were performed:
* Remove unnecessary `abi.encodePacked(...)` call when dealing with `relayTokens` functionality.
* ~~Remove redundant `require(!lock());` check in `_relayTokens`, since the temporary lock variable cannot be set to `true` in the beginning of the `relayTokens` call.~~
* Remove redundant `lock()` check in `bridgeSpecificActionsOnTokenTransfer`, since the check was either already done previously in the code (in `onTokenTransfer`), or the lock was explicitly disabled by the call to `setLock(false)` (in `_relayTokens`).

### DIFF
--- a/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
+++ b/contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
@@ -148,7 +148,7 @@ contract HomeMultiAMBErc20ToErc677 is
             // require(isTokenRegistered(token));
             require(withinLimit(token, _value));
             addTotalSpentPerDay(token, getCurrentDay(), _value);
-            bridgeSpecificActionsOnTokenTransfer(token, _from, _value, _data);
+            bridgeSpecificActionsOnTokenTransfer(token, _from, chooseReceiver(_from, _data), _value);
         }
         return true;
     }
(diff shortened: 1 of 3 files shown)


contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
```diff
@@ -165,7 +165,6 @@ contract HomeMultiAMBErc20ToErc677 is
         // This lock is to prevent calling passMessage twice if a ERC677 token is used.
         // When transferFrom is called, after the transfer, the ERC677 token will call onTokenTransfer from this contract
         // which will call passMessage.
-        require(!lock());
```
akolotov, line 168: I agree that the check whether the operation is locked in `bridgeSpecificActionsOnTokenTransfer` is redundant. But it seems that it safe to double check that `relayTokens` is not called by some token contract one more time as part of the `transferFrom` invocation. What do you think?

k1rill-fedoseev: Ok, even if it seems strange to me that the token contract itself will try to relay different tokens as part of the `transferFrom` operation, this check still helps to verify that the lock was not set somewhere earlier in the transaction.
So let's leave it as it was, I am ok with that.

akolotov (approved):

PR #555, opened 2020-11-12
Optimize use of temporary variables passed through storage

k1rill-fedoseev:
It was found that the Solidity compiler does not optimize the calculation of mapping storage slots in the cases where the slot remains constant. This PR refactors the code for temporary variables that are saved and removed from the storage within a single transaction. There are two benefits of such improvement:
* Transactions become cheaper, we are saving `~50` gas at each access operation to such variable by skipping redundant . For instance, usual Omnibridge withdraw operation contains 8 such accesses.
* Bytecode size decreases: AMB contracts bytecode decreased by `~300` bytes, mediators bytecode decreased by `~100` bytes each.

Also, worth noting that such optimizations can be also applied to other storage usages as well. For example, we can precalculate the exact storage slot of `addressStorage[BRIDGE_CONTRACT]`, this will have the same effect as for temporary variables. However, such refactoring should be done with extra caution in order to maintain the exactly same storage slots as it was before.

### DIFF
--- a/contracts/upgradeable_contracts/ReentrancyGuard.sol
+++ b/contracts/upgradeable_contracts/ReentrancyGuard.sol
@@ -3,13 +3,17 @@ pragma solidity 0.4.24;
 import "../upgradeability/EternalStorage.sol";
 
 contract ReentrancyGuard is EternalStorage {
-    bytes32 internal constant LOCK = 0x6168652c307c1e813ca11cfb3a601f1cf3b22452021a5052d8b05f1f1f8a3e92; // keccak256(abi.encodePacked("lock"))
-
-    function lock() internal returns (bool) {
-        return boolStorage[LOCK];
+    function lock() internal returns (bool res) {
+        assembly {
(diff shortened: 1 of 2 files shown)


akolotov (commented):
Consider to add comment before every inline assembly part that even if the operation like `boolStorage[LOCK]` introduces another level of addressing by using the order number of the boolStorage mapping, this is safe to sload/store value by using just the slot identifier  since the slot is not persistent -- it is cleared as part of the successful execution of the contract.

akolotov (commented):
> Also, worth noting that such optimizations can be also applied to other storage usages as well. For example, we can precalculate the exact storage slot of `addressStorage[BRIDGE_CONTRACT]`, this will have the same effect as for temporary variables. However, such refactoring should be done with extra caution in order to maintain the exactly same storage slots as it was before.

In general I would say that it not needed. Since it will increase the code complexity for reading significantly and could lead to mysterious issues in the future while introduce lesser benefits.

akolotov (approved):

PR #556, opened 2020-11-13
Add TokensBridgingInitiated events to mediators

### LINKED ISSUE #499: Add events to log deposit/withdrawal requests on AMB extensions
Currently AMB extensions (omnibridge, erc20-to-erc677, native-to-erc677, erc20-to-native) do not use events in order to notify applications about deposit/withdrawal requests that's why the applications must use combination of `Transfer` and `UserRequestForAffirmation`/`UserRequestForSignature` events to understand that the transaction was used to send tokens through the bridge. It makes sense to enhance the mediators contracts to simplify the code of applications.

k1rill-fedoseev:
Closes #499


akolotov (changes requested):
It could look confusing if the process is the token withdrawals but the event is `TokensDeposited`. Consider to rename it to something more common like `TokensBridgingInitiated`.

akolotov (approved):

PR #560, opened 2020-11-19
Reverse token bridging support for multi-token extension

k1rill-fedoseev:
Closes #558

Issue #558, opened 2020-11-16
Reversed omnibridge


akolotov:
Since there exists tokens that are native w.r.t. the xDai chain and the demand to transfer them to the Ethereum Mainnet is being increased it is necessary to extend functionality of the multi-token AMB extension to allow transfer not only bridged tokens from the Home chain.

If the mutli-token mediator recognizes that a token contract was not previously bridged from the Ethereum Mainnet, the part of tokens is locked on the mediator contract, another is distributed among the reward receivers. The request to deploy a new token contract and mint the corresponding amount of tokens is sent to the another side of AMB bridge.
As soon as the requests from the Home mediator appears in the Ethereum Mainnet, the new token contract is deployed, the corresponding amount of tokens is minted.
The opposite operation for the token bridged from the xDai chain must burn the tokens.
If the mediator on Home side identifies that a token was already reflected on the Ethereum Mainnet, the transfer operation just mint the token without the token contract deployment.

akolotov:
Addressed in a separate repo for the OmniBridge contracts: https://github.com/poanetwork/omnibridge

PR #561, opened 2020-11-24
Reversed Omnibridge by an adjacent pair of mediators

k1rill-fedoseev:
This PR suggest another possible way to solve #558 
As opposite to #560, this PR implements a reversed mode by using a separate independant pair of contracts.


akolotov:
Closed since the OmniBridge contracts were moved to a separate repo: https://github.com/poanetwork/omnibridge

PR #566, opened 2020-12-08
Support custom minter contract for bridged stake token

k1rill-fedoseev:


varasev (approved):

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
```diff
@@ -303,4 +303,31 @@ contract HomeMultiAMBErc20ToErc677 is
...
+     * @dev Internal function for getting minter proxy address.
+     * Returns the token address itself, expect for the case with bridged STAKE token.
+     * For bridged STAKE token, returns the hardcoded TokenMinter contract address.
+     * @param _token address of the token to mint.
+     * @return address of the minter contract that should be used for calling mint(address,uint256)
+     */
+    function _getMinterFor(address _token) internal view returns (IBurnableMintableERC677Token) {
+        if (_token == address(0xb7D311E2Eb55F2f68a9440da38e7989210b9A05e)) {
+            // hardcoded address of the TokenMinter address
+            return IBurnableMintableERC677Token(0x1111111111111111111111111111111111111111);
```
akolotov, line 317: I would avoid to merge this particular code into the `develop` or `master` branches, I would recommend to replace `0x1111111111111111111111111111111111111111` by `0xb7D311E2Eb55F2f68a9440da38e7989210b9A05e` in this line since we can guarantee that nothing will be broken if by some reason the code will be deployed as is. Other adopters of the TokenBridge contract would decide to do so for their instances of TokenBridge.

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/BasicMultiAMBErc20ToErc677.sol
```diff
@@ -58,7 +58,7 @@ contract BasicMultiAMBErc20ToErc677 is
     * @return patch value of the version
     */
     function getBridgeInterfacesVersion() external pure returns (uint64 major, uint64 minor, uint64 patch) {
-        return (1, 3, 0);
+        return (1, 3, 1);
```
akolotov, line 61: Since the functionality of the contract is going to be changed and the changes are not a fix, I believe we can update the version to `1.4.0` here.

akolotov (changes requested):

akolotov (approved):

PR #572, opened 2021-01-18
Allow to make new AMB requests while processing other messages

k1rill-fedoseev:


contracts/upgradeable_contracts/arbitrary_message/MessageDelivery.sol
```diff
@@ -29,9 +29,6 @@ contract MessageDelivery is BasicAMB, MessageProcessor {
     * @param _dataType AMB message dataType to be included as a part of the header
     */
     function _sendMessage(address _contract, bytes _data, uint256 _gas, uint256 _dataType) public returns (bytes32) {
-        // it is not allowed to pass messages while other messages are processed
-        require(messageId() == bytes32(0));
```
akolotov, line 33: Can we consider to have this option configurable? By default, this check must be turned on and if it is configured explicitly generating a message for AMB during AMB message execution must be allowed. The reason is the following: for some cheap pairs of networks (e.g. POA Chain and xDai chain), the request-and-claim approach will not be used to simplify the user experience. But perhaps we would like still limit AMB-loops and AMB messages multiplications.

k1rill-fedoseev: Yes, it is possible. Please have a look.

akolotov (approved):

PR #576, opened 2021-02-20
Helpers to collect signatures for erc20-to-native and AMB

akolotov:
@k1rill-fedoseev has prepared two helpers which were deployed to xDai chain to help users manually collect signatures provided by the bridge oracles. The collected signatures are being used on another side of the bridge to execute the transfer.


akolotov:
@k1rill-fedoseev can you suggest changes to fix the issue with the helper contracts compilation in the `lint` and `test` checks of the GitHub actions

k1rill-fedoseev:
Done

PR #580, opened 2021-03-09
Fix relayTokens for tokens supporting fees

k1rill-fedoseev:


akolotov (approved):

contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeMultiAMBErc20ToErc677.sol
```diff
@@ -183,17 +212,19 @@ contract HomeMultiAMBErc20ToErc677 is
...
         require(withinLimit(token, _value));
         addTotalSpentPerDay(token, getCurrentDay(), _value);
 
+        uint256 balanceBefore = token.balanceOf(address(this));
         setLock(true);
-        token.transferFrom(msg.sender, to, _value);
+        token.transferFrom(msg.sender, address(this), _value);
         setLock(false);
-        bridgeSpecificActionsOnTokenTransfer(token, msg.sender, _receiver, _value);
+        uint256 balanceDiff = token.balanceOf(address(this)).sub(balanceBefore);
```
akolotov, line 225: do we need these actions for the the Home contracts? We are working with our version of ERC677 token and we know for sure that the balance will be changed by the same value which is specified in the `transferFrom` method call, don't we?

k1rill-fedoseev: Agree, seems like for half-duplex OB there is not need in such changes on Home side.

akolotov (approved):

PR #581, opened 2021-03-10
Disable per-token fees

k1rill-fedoseev:


contracts/upgradeable_contracts/multi_amb_erc20_to_erc677/HomeFeeManagerMultiAMBErc20ToErc677.sol
```diff
@@ -76,7 +77,7 @@ contract HomeFeeManagerMultiAMBErc20ToErc677 is BaseRewardAddressList, Ownable,
     * @return fee value associated with the requested fee type.
     */
     function getFee(bytes32 _feeType, address _token) public view validFeeType(_feeType) returns (uint256) {
-        return uintStorage[keccak256(abi.encodePacked(_feeType, _token))];
+        return uintStorage[keccak256(abi.encodePacked(_feeType, address(0)))];
```
akolotov, line 80: Would it be more suitable?
```solidity
function getFee(bytes32 _feeType, address _token) public view validFeeType(_feeType) returns (uint256) {
    if (uintStorage[keccak256(abi.encodePacked(_feeType, _token))] > 0) {
        return uintStorage[keccak256(abi.encodePacked(_feeType, address(0)))];
    } else return 0;
}
```
The logic is following: if fees for some tokens are disabled, the new code will not enable them again.

k1rill-fedoseev: Ok, that makes sense. I didn't thought about disabled fees for some tokens.

akolotov (approved):

PR #623, opened 2021-08-15
Stricter preconditions for payInterest

k1rill-fedoseev:
The `InterestReceiverStakeBuyback` and `InterestReceiverSwapToETH` contracts both utilize Uniswap in order to swap tokens. There is some slippage protection, but no protection against price manipulation is built in.

Because the `payInterest` function in the `InterestConnector` is external and permissionless, anyone can call it at any time. Therefore, once enough interest accrues, one could take out a flash loan to manipulate the relevant Uniswap pools and call `payInterest` to make a profit.

Additionally, even if `payInterest` weren't permissionless, it is possible for the Uniswap call in `onInterestReceived` to fail, leaving the tokens sitting in the `InterestReceiver` contract. Here, the `onInterestReceived` function is external and permissionless, meaning it's once again prone to being attacked via price manipulation.

Furthermore, assuming the interest payment was not callable by untrusted parties at all, it would still be possible to manipulate the price by sandwiching that transaction between two others, either by chance or by collusion with miners.

The fix is to disable ability of storing tokens on the interest receiver contract and introduces extra EOA-only check to payInterest method.

This reduces the risk of possible price manipulations, since it is no longer possible to use flash-loans to perform an attack. 

Front-running of the transaction is considered less dangerous, as it requires a much higher attack cost (at least 2 Uniswap transactions gas fees + 2 slippage and swap fees).

### DIFF
--- a/contracts/upgradeable_contracts/InterestReceiverStakeBuyback.sol
+++ b/contracts/upgradeable_contracts/InterestReceiverStakeBuyback.sol
@@ -26,14 +26,6 @@ contract InterestReceiverStakeBuyback is InterestReceiverBase {
         // (min received %) * (amount / 1 DAI) * (STAKE per 1 DAI)
         uint256 minAmount = (minReceivedFraction * amount * uniswapRouterV2.getAmountsOut(1 ether, path)[2]) / 10**36;
 
-        bytes memory data = abi.encodeWithSelector(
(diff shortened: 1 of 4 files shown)


akolotov (approved):

azboyzranch (commented):
>.

PR #628, opened 2021-08-19
Update GSN interface

LyzeOfKiel:
## 1

The `postRelayedCall` function of the `TokenPaymaster` contract is public and allows anyone to call it. In this function the uniswap router is called and leftover funds are sent back to the user. Thus, this function modifies the state. Without an access control check that only allows the `RelayHub` to call it, an attacker can craft call data that will drain the ERC20 balance of the contract. Note that normally the contract is not supposed to have an ERC20 balance, however, this can change in situations as described in Anyone can steal GSN fees.

The `preRelayedCall` and `postRelayedCall` functions both have following comment in the official GSN `IPaymaster` class:

https://github.com/poanetwork/tokenbridge-contracts/blob/b3511bf0987bbfef661e28dd1a6fbe1735f90ac0/contracts/gsn/interfaces/IPaymaster.sol#L57

The `preRelayedCall` function does not modify state but it could still make sense to add an access control check.

## 2

The contracts use an outdated GSN interface in the definition of `IPaymaster`. In particular the following differences exist:
  * The function `getGasLimits` should now be called `getGasAndDataLimits()` and should return four instead of three values.
  * A `trustedForwarder` function should be implemented.

Furthermore, the `RelayData` struct has changed. As less context information is available the security guarantees are weaker.

----

The updates in this PR consist of:

1) Add `relayHubOnly` modifier for `postRelayedCall`
2) Update GSN Interface:
    1) Add `GasAndDataLimits` struct and `trustedForwarder` function in `IPaymaster` interface
    2) Update `RelayData` struct
3) Add `claimTokens` function in `TokenPaymaster`

### DIFF
--- a/contracts/gsn/token_paymaster/TokenPaymaster.sol
+++ b/contracts/gsn/token_paymaster/TokenPaymaster.sol
@@ -6,21 +6,23 @@ import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
 import "../BasePaymaster.sol";
 import "./IUniswapV2Router02.sol";
 import "../../upgradeable_contracts/GSNForeignERC20Bridge.sol";
+import "../../upgradeable_contracts/Claimable.sol";
 
-contract TokenPaymaster is BasePaymaster {
-    address private token;
+contract TokenPaymaster is BasePaymaster, Claimable {
+    ERC20 private token;
     IUniswapV2Router02 private router;
     address private bridge;
 
     address[] private tokenWethPair = new address[](2);
     uint256 public postGasUsage = 300000;
+    uint256 public calldataSizeLimit = 1348;
 
     constructor(address _relayHub, address _forwarder, address _token, address _router, address _bridge) public {
         _setOwner(msg.sender);
-        relayHub = IRelayHub(_relayHub);
-        trustedForwarder = IForwarder(_forwarder);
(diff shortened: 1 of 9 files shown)


contracts/gsn/token_paymaster/TokenPaymaster.sol
```diff
@@ -6,21 +6,23 @@ import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
...
-contract TokenPaymaster is BasePaymaster {
-    address private token;
+contract TokenPaymaster is BasePaymaster, Claimable {
+    ERC20 private token;
     IUniswapV2Router02 private router;
     address private bridge;
 
     address[] private tokenWethPair = new address[](2);
     uint256 public postGasUsage = 300000;
+    uint256 public calldataSizeLimit = 1348;
```
k1rill-fedoseev, line 18: Why exactly this constant? If there is a way to calculate it, we should put a comment on that.

LyzeOfKiel: Changed to a default GSN value - 10500. It should fit enough signatures for target bridge function. But in case it is not enough there is a setter for this value

k1rill-fedoseev (approved):

akolotov (approved):

PR #684, opened 2023-09-01
Update AMBBridgeHelper.sol

0xScratch:
Changed `uint256 i = 0` to `uint256 i`, there's no need to initialize some 'uint' variable to 0 cuz the default value already sets itself to 0. Saves some gas too.

Moreover, there's one more change we can make in this 'for' loop, that is changing `i++` to `++i`, if this change isn't messing up with the code logic, This is always beneficial to do so. Although, I haven't made this change, just a suggestion if you are down with it!


0xScratch:
Closing due to inactivity!

